PR #2, opened 2023-02-05
Solver APIs

PierreGuilmin:


torchq/mesolve.py
```diff
@@ -0,0 +1,65 @@
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+import torch
+import torchdiffeq
+
+
+def mesolve(
```
gautierronan, line 8: Pour les fonctions principales comme celle-ci, je trouve plus lisible de mettre le typing dans le docstring. Surtout si on commence à avoir des `Union[QTensor, list of QTensor, nn.Module]` ou truc comme ça pour `H`.
Pour toutes les autres fonctions par contre, le typing dans les arguments c'est très bien!

PierreGuilmin: Oui c'est vrai. Une autre option, c'est de créer des types agrégés :
```python
from typing import Union
TimeQTensor = Union[QTensor, List[QTensor], ...]

def mesolve(
    H: TimeQTensor,
# ....
```

PierreGuilmin: Ho and we should speak English, in case an external collaborator joins or for our future curious users digging the old PRs 😋

gautierronan: As you wish, I don't have a super strong opinion on the subject. But I know that when I was reading the source code of torchdyn I was having a hard time understanding the typing because I was not familiar with it. While torchdiffeq is not using any typing, and I found it easier to understand (if variable names are well chosen). I'm guessing most python users are not familiar with typing either, so only keeping it in the dosctring would be probably make the function more accessible.

gautierronan (approved):

gautierronan (approved):
Let's merge this PR and the other one for now, and once QTensor is written, we can probably start iterating on mesolve/sesolve to get to something more concrete. I will write the odeint and odeint_adjoint at some point in the coming weeks too.

torchq/mesolve.py
```diff
@@ -0,0 +1,65 @@
...
+) -> torch.Tensor:
+    if options is None:
+        options = {}
+
+    if solver == 'rk4':
+        if not 'step_size' in options:
+            options['step_size'] = 0.001
+        return _mesolve_torchdiffeq(H, Ls, rho0, tlist, options)
+
+    raise ValueError('invalid solver (supported: \'rk4\')')
```
abocquet, line 24: Why do directly sending the `solver` parameter to `torchdiffeq.odeint` `method` parameter?

PierreGuilmin: I just didn't try the other solvers, but I agree we could do as you suggest 👍

PR #4, opened 2023-02-09
Add QTensor class

abocquet:
No `to_numpy` function is required since it's built-in torch `Tensor`


torchq/qtensor.py
```diff
@@ -0,0 +1,61 @@
...
+class QTensor(torch.Tensor):
+    """
+    A torch tensor that represents a quantum object (operator, density matrix, state vector...)
+
+    Subclasses of Tensor can slow torch down significantly
+    TODO: check performance impact and implement mitigations
+    Ref : https://benjaminwarner.dev/2022/06/14/debugging-pytorch-performance-decrease
+    """
+
+    def __init__(self, *args, **kwargs):
```
gautierronan, line 17: Should we provide an optional `type` argument as in qutip? This would be `{'bra', 'ket', 'oper', 'operator-ket', 'operator-bra', 'super'}`.

abocquet: I've never used them. How useful is it?
If we put them, shouldn't we need a `density-matrix` type?

gautierronan: Not sure what it's for to be honest! I guess if we ever need it we can add it then.

PR #5, opened 2023-02-10
ME solve with Rouchon scheme

abocquet:


gautierronan:
In my mind, the function `mesolve` should not directly implement any solver. It should instead be a wrapper for calling the appropriate ODE solver (see my other PR).

So, `mesolve` should:

- check the input arguments
- create an appropriate`nn.Module` depending on the solver. For RK solvers, this Module should yield `L(t) @ rho(t)`. For Rouchon solvers, it should yield `rho(t+dt) = sum M rho(t) M^dag` instead.
- Feed this `nn.Module` to the `odeint` integrator.
- Unpack the results and return them in the appropriate format.

This allows freedom to define various Rouchon-like solvers (order 1, order 2, with or without the inv_sqrtm correction, time-dependent or not, etc.).

abocquet:
I completely agree, but I wanted to have a first version of `mesolve` that works. Then later when we add another solver, we can add more abstraction

PierreGuilmin:
> I completely agree, but I wanted to have a first version of mesolve that works. Then later when we add another solver, we can add more abstraction

But we _had_ a version of `mesolve` that works 😋 I'd like to compare `torchdiffeq` solvers, the Rouchon method, our ode solver, etc… Plus I guess people who don't know Mr. Rouchon will trust more easily a Runge-Kutta (which is actually more appropriate in some settings), and we should let them choose. So could we keep the `torchdiffeq` solver for now? As @gautierronan suggests we can maybe already let `mesolve` be a wrapper with our two current methods: `torchdiffeq` Runge-Kutta and the Rouchon solver you implemented.

I'm doing a proper review now.

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
     H: torch.Tensor,
-    Ls: List[torch.Tensor],
     rho0: torch.Tensor,
-    tlist: np.ndarray,
-    solver: str = 'rk4',
-    options: Optional[Dict[str, Any]] = None,
-) -> torch.Tensor:
-    if options is None:
-        options = {}
+    times: np.ndarray,
```
PierreGuilmin, line 13: I like `tsave` as suggested by @gautierronan, to make it clear that this argument is not fixing the solver timesteps.

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
+    c_ops: List[torch.Tensor] = None,
+    e_ops: List[torch.Tensor] = None,
+    n_substeps: int = 1,
```
PierreGuilmin, line 16: This should be in `options` and specific to the Rouchon solver.

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
-    Ls: List[torch.Tensor],
     rho0: torch.Tensor,
-    tlist: np.ndarray,
-    solver: str = 'rk4',
-    options: Optional[Dict[str, Any]] = None,
-) -> torch.Tensor:
-    if options is None:
-        options = {}
+    times: np.ndarray,
+    c_ops: List[torch.Tensor] = None,
```
PierreGuilmin, line 14: Could we pick `jump_ops` rather? I know it's different from QuTiP, but naming-wise QuTiP has some very bad standard so we should be careful.

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
+    e_ops: List[torch.Tensor] = None,
+    n_substeps: int = 1,
+    save_states=True,
+    progress_bar=False,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    assert n_substeps > 0
```
PierreGuilmin, line 20: Could we add a reference to the original paper after displacing this method to its own function (https://journals.aps.org/pra/abstract/10.1103/PhysRevA.91.012118 if I'm not mistaken)?

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
+    R = (
+        torch.eye(*H.shape)
+        + (-1j * H + 0.5 * c_dag_c) @ (1j * H + 0.5 * c_dag_c) * dt**2
+    )
+    inv_sqrt_R = _inv_sqrtm(R)
+
+    M_r = torch.eye(*H.shape, dtype=H.dtype) - (1j * H + 0.5 * c_dag_c) * dt
+    M_r_tilde = M_r @ inv_sqrt_R
+
+    c_ops_tilde = [op @ inv_sqrt_R for op in c_ops]
```
PierreGuilmin, line 35: I think this part deserves a little more comments.

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
+
+def _inv_sqrtm(matrix: torch.Tensor) -> torch.Tensor:
+    r"""
+    Power of a matrix using Eigen Decomposition.
+    Args:
+        matrix: matrix
+        p: power
+    Returns:
+        Power of a matrix
+    """
```
PierreGuilmin, line 80: I know Mr. Rouchon does this using Schur decomposition, I'm guessing for big system it's hopeless to diagonalize the matrix. We can keep this bit for a later PR if you want.

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
+    for _ in times:
+        for _ in range(n_substeps):
+            next_rho = M_r_tilde @ rho
+            next_rho = next_rho @ hermitian(M_r_tilde)
+            next_rho += sum(
+                [
+                    c_op_tilde @ rho @ hermitian(c_op_tilde) * dt
+                    for c_op_tilde in c_ops_tilde
+                ]
+            )
```
PierreGuilmin, line 53: There's a specific speedup for time-independent Hamiltonian that we could use, where one pre-computes the action of $e^{-iH\Delta t/2}$ and applies it both on the right and on the left. We can keep this for a later PR if you want.

PierreGuilmin (changes requested):
Very nice thanks 👏  I'm not 100% done yet with the review, but here's a first batch of comments.

torchqdynamics/mesolve.py
abocquet: `c_ops` and `e_ops` are common to all me solver I think. They are in qutip

torchqdynamics/mesolve.py
```diff
@@ -1,27 +1,90 @@
...
+
+    dt = (times[1] - times[0]) / n_substeps
+    c_dag_c = torch.stack([hermitian(op) @ op for op in c_ops]).sum(dim=0)
+    R = (
+        torch.eye(*H.shape)
+        + (-1j * H + 0.5 * c_dag_c) @ (1j * H + 0.5 * c_dag_c) * dt**2
+    )
+    inv_sqrt_R = _inv_sqrtm(R)
+
+    M_r = torch.eye(*H.shape, dtype=H.dtype) - (1j * H + 0.5 * c_dag_c) * dt
```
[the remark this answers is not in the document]
PierreGuilmin: It's actually clearer in Pierre's last paper (2022: [A tutorial introduction to quantum stochastic master equations based on the qubit/photon system](https://www.sciencedirect.com/science/article/abs/pii/S1367578822001006)) than the one you used (2016: [Anatomy of fluorescence: quantum trajectory statistics from continuously measuring spontaneous emission](https://link.springer.com/article/10.1007/s40509-016-0075-9)). The one I talked about in another comment is the first version of the solver, without the renormalisation (2015: [Efficient quantum filtering for quantum feedback control](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.91.012118)). I think we can cite all three with the relevant sections (I can add it in another PR if you want 😊).

Here's the bit ($S=R$ of your code, I think it's better to compute the normalisation factor than using a first-order formula as in the 2016 paper):
Screenshot 2023-02-15 at 15 29 30

And taken from Pierre's Matlab code (here `L` is a relaxation operator and `V2` some two-photon exchange purely dissipative operator, but we don't really care):
```matlab
M0=Id-(L'*L+V2'*V2)*dt/2;
S=M0'*M0+dt*(L'*L+V2'*V2);
[V,T]=schur(S);
Ds=real(diag(T));
Ds=Ds.^(-0.5);
S=V*diag(Ds)*V';
M0s=M0*S;
Ls=L*S;
V2s=V2*S;
```

torchqdynamics/mesolve.py
PierreGuilmin: See also Pierre's last paper section 3.4 for this (2022: [A tutorial introduction to quantum stochastic master equations based on the qubit/photon system](https://www.sciencedirect.com/science/article/abs/pii/S1367578822001006) section 3.4. Kraus maps and numerical schemes for diffusive SME).

PR #6, opened 2023-02-12
Add ODE integrator

gautierronan:
This PR adds a home-made ODE integrator and test functions for it.

For now, this supports:

- A single runga-kutta solver with adaptive time step, Dormand Prince of order 5.
- Support for outsourced solvers (meaning that `odeint` outsources the computation of `y(t+dt)` to the user). To be used for the Rouchon method.
- Automatic differentiation through PyTorch's backward AD.
- Automatic differentiation of linear functions with the adjoint method.

To be added in other PRs:

- Other ODE solvers.
- Test functions for the gradients with backward- and adjoint-based AD that compare to analytical gradients.
- In-place version of the solvers in case AD is not required by the user (higher performance).


torchqdynamics/ode/adjoint.py
```diff
@@ -0,0 +1,110 @@
...
+        
+        # Initialize the augmented state as a flat tensor
+        # y is the state, a is the adjoint, g is the gradient w.r.t. parameters
+        yT, aT, gT = y_saved[-1], grad_y[-1][-1], torch.zeros_like(params)
+        y_len, a_len, g_len = yT.numel(), aT.numel(), gT.numel()
+        y_shape, a_shape, g_shape = yT.shape, aT.shape, gT.shape
+        aug_y = torch.cat([yT.view(-1), aT.view(-1), gT.view(-1)])
+
+        solver_type = ode.solver_to_solvertype(solver)
+        if solver_type == "rk":
```
abocquet, line 52: Better declare an enumeration here 
```
class Solver:
    RK = "rk"
    ROUCHON = "rouchon"
    [etc...]
```
and then later we can refactor easily and not be stuck with strings

torchqdynamics/ode/adjoint.py
```diff
@@ -0,0 +1,110 @@
...
+                    # Compute dg/dt = - (d(da/dt)/dtheta @ y)
+                    dg = torch.autograd.grad(da, params_, y.view(-1), 
+                                            allow_unused=True, retain_graph=True)
+                    # Convert back dg to flat tensor
+                    dg = torch.cat([torch.zeros_like(param).view(-1) if dgi is None else dgi.view(-1)
+                                for param, dgi in zip(params_, dg)]).conj()
+
+                # Repack and return
+                return torch.cat([dy, da, dg])
+        elif solver_type == "out":
```
abocquet, line 72: Same here

torchqdynamics/ode/odeint.py
```diff
@@ -0,0 +1,240 @@
...
+    # Initialize the ODE routine
+    f0 = f(t0, y0)
+    dt = init_step(f, f0, y0, t0, solver.order, atol, rtol)
+    
+    # Run the ODE routine
+    t, y, ft = t0, y0, f0
+    while t < T:         
+        # Check if checkpointing required, or final time is reached
+        if len(save_at) > 0 and (ckpt_counter < len(save_at)) and (t + dt >= save_at[ckpt_counter]):
+            dt_old, ckpt_flag = dt, True
```
abocquet, line 116: I guess `ckpt` is `checkpoint` but maybe it's better to put explicit name for non common words

torchqdynamics/ode/solver.py
```diff
@@ -0,0 +1,144 @@
+import torch
+import torch.nn as nn
+from .tableaus import construct_dopri5
+from math import sqrt
+
+# -------------------------------------------------------------------------------------------------
+#     Solver common functions
+# -------------------------------------------------------------------------------------------------
+
+DTYPE_EQUIV = {
```
abocquet, line 10: Don't forget to install the pre-commit hooks to have black formatting 😃

torchqdynamics/ode/adjoint.py
gautierronan: I'm not familiar with that. How could I implement that in this particular exemple? Like, where are the two functions actually defined?

abocquet:
Pytorch does use black (https://github.com/pytorch/pytorch/blob/master/pyproject.toml#L18) and Tensorflow uses pylint. I don't really mind which formatter we use as long as there is one everyone can agree on so we have a uniform code style in the whole codebase

PierreGuilmin:
Wow, that's a big chunk, amazing work! Is it ready for review @gautierronan? If yes, it will need a lil' bit of time to read everything, but I can find some time to do a thorough review.

gautierronan:
> Wow, that's a big chunk, amazing work! Is it ready for review @gautierronan? If yes, it will need a lil' bit of time to read everything, but I can find some time to do a thorough review.

Yes I believe it's all good. I'm changing a few things here and there as I keep working on it, but it's tested and seems to work fine !

torchqdynamics/ode/adjoint.py
abocquet: See abocquet/odeint-enums

gautierronan:
Closing this PR. I will be rewriting it and splitting it in multiple smaller PRs.

PR #7, opened 2023-02-15
Formatting with yapf

gautierronan:
Moving the discussion from https://github.com/PierreGuilmin/torchqdynamics/pull/6 to this PR. I played around with the formatters a little bit, and I quite like the yapf formatting with facebook style and a few other specific arguments. Mostly, it avoids wrapping long function definitions on many lines (which I find quite unreadable). Would this formatting be fine with you guys?

Sorry for being annoying on this haha. I feel like it's quite important to have as readable code as possible, since contrary to popular libraries, we won't have extensive documentation (for now at least).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,15 +4,20 @@ repos:
     rev: v3.2.0
     hooks:
       - id: trailing-whitespace
-      - id: end-of-file-fixer
-      - id: check-yaml
+      - id: check-docstring-first
+      - id: check-json
       - id: check-added-large-files
-
-  - repo: https://github.com/google/yapf
-    rev: v0.32.0
+      - id: check-yaml
+      - id: debug-statements
+      - id: requirements-txt-fixer
+      - id: check-merge-conflict
+      - id: double-quote-string-fixer
+      - id: end-of-file-fixer
+      - id: sort-simple-yaml
+  - repo: meta
     hooks:
-      - id: yapf
-
+      - id: check-hooks-apply
+      - id: check-useless-excludes
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,10 +8,10 @@ repos:
       - id: check-yaml
       - id: check-added-large-files
 
-  - repo: https://github.com/psf/black
-    rev: 22.10.0
+  - repo: https://github.com/google/yapf
+    rev: v0.32.0
     hooks:
-      - id: black
+      - id: yapf
 
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,8 @@ repos:
     hooks:
       - id: yapf
         name: yapf
-        language: python
-        entry: yapf
+        additional_dependencies: [toml]
         args: [-i]
-        types: [python]
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
         name: yapf
         language: python
         entry: yapf
-        args: [-i, -vv]
+        args: [-i]
         types: [python]
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,14 @@
 files: torchqdynamics
 repos:
+  - repo: https://github.com/google/yapf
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -18,13 +18,17 @@ build-backend = "setuptools.build_meta"
...
-target-version = ["py38"]
-skip-string-normalization = true
+[tool.yapf]
+based_on_style = "facebook"
+column_limit = 88
+indent_width = 4
+SPLIT_BEFORE_NAMED_ASSIGNS = false
+DEDENT_CLOSING_BRACKETS = true
+SPLIT_ARGUMENTS_WHEN_COMMA_TERMINATED = false
+INDENT_DICTIONARY_VALUE = true
```
PierreGuilmin, line 28: I think these three guys are by default in the `"facebook"` style, so unnecessary to specify.

pyproject.toml
```diff
@@ -18,13 +18,17 @@ build-backend = "setuptools.build_meta"
...
 
 [tool.isort]
 profile = "black"
-line_length = 80  # to be consistent with black
+line_length = 88  # to be consistent with yapf
 skip = ["__init__.py"]
+
```
PierreGuilmin, line 34: Could we remove the extra line?

pyproject.toml
```diff
@@ -18,13 +18,17 @@ build-backend = "setuptools.build_meta"
...
-[tool.black]
-line-length = 80
-preview = true
-target-version = ["py38"]
-skip-string-normalization = true
+[tool.yapf]
+based_on_style = "facebook"
+column_limit = 88
+indent_width = 4
+SPLIT_BEFORE_NAMED_ASSIGNS = false
```
PierreGuilmin, line 25: Could we pick a uniform capitalization style?

pyproject.toml
PierreGuilmin: You can tune your IDE so it does all this little things for you, for example in VSCode user configuration
```json
"files.trimTrailingWhitespace": true,
"files.insertFinalNewline": true,
"files.trimFinalNewlines": true,
"files.eol": "\n",
```

PierreGuilmin (changes requested):
I'm fine with any formatter as long as we have one, I've been wanting to try yapf for a while, so I'm happy to go with it! 😊 Note that you also need to update `requirements-dev.txt`, `.pre-commit-config.yaml` and the `README.md`. Also, now would be a good time to reformat all the files and commit the modifications, as we don't have a lot of code and not too many PRs relying on existing code.

pyproject.toml
gautierronan: The top and bottom ones yes, but not the middle one. Changing accordingly.

pyproject.toml
gautierronan: Added that in my VSCode settings, thanks !

pyproject.toml
PierreGuilmin: The middle one is in PEP8 style from which Facebook inherits.

gautierronan:
Should be good now. I had to add a `setup.cfg` file to work around https://github.com/google/yapf/issues/936.

README.md
```diff
@@ -22,17 +22,17 @@ conda develop /path/to/torchqdynamics
...
 ```shell
-pip install isort black
+pip install isort yapf
 ```
 
 Run the following before each commit:
 ```shell
 isort torchqdynamics
-black torchqdynamics
+yapf --style={based_on_style: facebook, column_limit: 88, split_before_named_assigns: False} torchqdynamics
```
PierreGuilmin, line 31: Weird, yapf should pick the `pyproject.toml` settings without needing to specify all of them. Maybe it's due to

> Should be good now. I had to add a setup.cfg file to work around https://github.com/google/yapf/issues/936.

I think a better fix is just to add `toml` as a dependency.

.pre-commit-config.yaml
```diff
@@ -1,18 +1,32 @@
 files: torchqdynamics
 repos:
+  - repo: https://github.com/google/yapf
+    rev: v0.32.0
+    hooks:
+      - id: yapf
+        name: yapf
+        language: python
+        entry: yapf
+        args: [-i, -vv]
```
PierreGuilmin, line 10: I think we can remove `-vv`, it will get very noisy when having more files.

PierreGuilmin (changes requested):
Alright, almost done, nice little addons in the pre-commit config 🥓

PierreGuilmin:
I will add some CI soon now with GitHub Actions now that we have so many cool checks

README.md
gautierronan: Maybe passing the style here is useless indeed. I was not sure, so I did it anyway :D

PierreGuilmin: I think we need `yapf -i -r torchqdynamics`.

gautierronan: Ok I'm not quite sure about all this tbh. Can I let you make the proper changes on the pre-commit stuff?

PierreGuilmin (approved):

PR #8, opened 2023-02-16
Change solver type from string to enum

abocquet:


gautierronan:
Thanks for showing how this can be done ! I think it's indeed a better option. However, in your implementation you're mixing two things (which maybe were not so clear on my end) and so the solver won't work anymore.

There's the actual solver (i.e. dopri5, dopri8, rk2, ...) for which there should be an actual class (sub-class of `BaseSolver`) that implements how the next step is computed. And there's the solver "type" (not sure it's the best name) that tells us whether the solver is within the odeint ecosystem, or outsourced to the user. So, in the adjoint method, there's only two functions to be defined (ever), which depend on this solver "type" (in or out). Whereas there's a priori an infinite number of solvers that can be implemented by us.

torchqdynamics/ode/odeint.py
```diff
@@ -12,14 +13,28 @@
...
+    OUT = 'out'
+
+
 def odeint(
-    f, y0, tspan, solver='dopri5', save_at=(), sensitivity=None, model_params=None,
-    atol=1e-8, rtol=1e-6, backward_mode=False
+    f,
+    y0,
+    tspan,
+    solver=Solver.DOPRI5,
```
gautierronan, line 26: The end user should be supplying a string as the `odeint` argument, and not an object. It's much simpler on the user end since there's no need to import an additional object.

gautierronan: So we need to convert afterwards from the string to the Solver argument afterwards

PierreGuilmin:
Closing this because of #10.

PR #11, opened 2023-02-22
Add tests directory

PierreGuilmin:
I didn't add `pytest` to `pre-commit` as I expect us to have some quite lengthy tests, and as a user I wouldn't want to run them for _each_ commit. I'm not a user of `pre-commit`, so I don't know what's best, @abocquet you tell me 😉

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,4 +29,4 @@ repos:
     rev: 5.12.0
     hooks:
       - id: isort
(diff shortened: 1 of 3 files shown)


abocquet (approved):
I don't have a strong opinion on running the tests in at commit but I can see how it can be annoying, so I'm fine with this.

Now we could consider running the CI using Github workflow if we feel the need at some point

PR #12, opened 2023-02-28
Framework for mesolve and odeint

gautierronan:
In this PR, I make an attempt at the overall framework for mesolve (and sesolve, smesolve) and odeint, following the discussion we had the other day.

I introduce a class named `QSolver` which implements a single ODE iteration of any given solver. For now, it is written for first and second-order Rouchon methods.  It can later also be derived into subclasses for other solvers, both fixed and adaptive time steps. It can also easily be derived with the 1.5 order Rouchon scheme of https://github.com/PierreGuilmin/torchqdynamics/pull/5. All Qsolvers will eventually support the `forward_adjoint` method for adjoint-based autodiff. 

Inside `mesolve`, an instance of `QSolver` is created using the solver options dataclass of https://github.com/PierreGuilmin/torchqdynamics/pull/10, and the ODE integrator `odeint` is called. In the end, `odeint` is meant to be common to all main functions of the library, such as `sesolve` and `smesolve`.

I intentionnally made the Hamiltonian `H` time-dependent for now, and the `jump_ops` time-independent, as I believe this should be the most common scenario. Also, `H` is a callable for now, but meant to also accept a `torch.Tensor` or a `List[torch.Tensor]` with time-dependent coefficients in the end. This should be done in a separate PR.

Anyway, I'll let you read this. I tried to be as basic as possible so we can all agree on the framework and then build on it. I found this was a good way to meet all of our demands. I will personnally be able to build on this for the adjoint-based odeint.

### DIFF
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -5,7 +5,7 @@ from .solver import Rouchon
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', variables=None
+    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
 ):
     if solver is None:
         # TODO: The default dt should not be choosen in such an arbitrary
@@ -22,7 +22,7 @@ def mesolve(
         raise NotImplementedError
 
     # compute the result
-    return odeint(qsolver, rho0, tsave, sensitivity=sensitivity, variables=variables)
+    return odeint(qsolver, rho0, tsave, sensitivity=sensitivity, parameters=parameters)
 
 
 class QSolver:
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -66,21 +66,7 @@ class MERouchon1(MERouchon):
 
 class MERouchon2(MERouchon):
     def forward(self, t, dt, rho):
-        """Compute rho(t+dt) using a Rouchon method of order 2."""
-        # non-hermitian Hamiltonian at time t
-        H_nh = self.H(t) - 0.5j * self.sum_nojump
-
-        # build time-dependent Kraus operators
-        # TODO: Add the missing time derivative term in -0.5j * dt**2 * \dot{H}
-        M0 = self.I - 1j * dt * H_nh - 0.5 * dt**2 * H_nh @ H_nh
-        M1s = 0.5 * self.Ls @ M0
-        M1s += M1s.adjoint()
-
-        # compute rho(t+dt)
-        drho_ = dt * (M1s @ rho.unsqueeze(0) @ M1s.adjoint()).sum(dim=0)
-        drho = M0 @ rho @ M0.adjoint() + drho_
-        drho += 0.5 * dt * (M1s @ drho_.unsqueeze(0) @ M1s.adjoint()).sum(dim=0)
-        return drho
+        raise NotImplementedError
 
     def forward_adjoint(self, t, dt, phi):
         raise NotImplementedError
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -10,7 +10,8 @@ from .odeint import odeint
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver, sensitivity='autograd', model_params=None
+    H, jump_ops, rho0, tsave, solver=Rouchon(), sensitivity='autograd',
+    model_params=None
 ):
     # Define the QSolver
     if isinstance(solver, Rouchon):
@@ -97,5 +98,6 @@ class MERouchon2(MERouchon):
 
 @dataclass
 class Rouchon:
-    dt: float
+    dt: float = 1e-2
+    order: int = 1
     stepclass: str = 'fixed'
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -56,8 +56,10 @@ class MERouchon1(MERouchon):
         M0 = self.I - 1j * dt * H_nh
 
         # compute rho(t+dt)
-        rho = M0 @ rho @ M0.adjoint()
-        rho += dt * (self.jump_ops @ rho.unsqueeze(0) @ self.jumpdag_ops).sum(dim=0)
+        rho = (
+            M0 @ rho @ M0.adjoint() + dt *
+            (self.jump_ops @ rho.unsqueeze(0) @ self.jumpdag_ops).sum(dim=0)
+        )
         rho = rho / rho.trace()
         return rho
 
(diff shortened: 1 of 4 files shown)


torchqdynamics/mesolve.py
```diff
@@ -1,63 +1,103 @@
...
+# --------------------------------------------------------------------------------------
+
+# See the PR by @abocquet at https://github.com/PierreGuilmin/torchqdynamics/pull/10
+
+
+@dataclass
+class Rouchon:
+    dt: float = 1e-2
+    order: int = 1
+    stepclass: str = 'fixed'
```
abocquet, line 103: Using strings for enumeration is a pattern I'm ill at ease with because we don't know what are the accepted values. Here I'm guessing `stepclass` could be `fixed` or "fixed but variable in time"?

abocquet (commented):
Only small nits for now. I have to think a little more about the architecture, I think we had different things in mind

torchqdynamics/mesolve.py
```diff
@@ -1,63 +1,103 @@
...
-        drho = _lindbladian(H, Ls, Lsdag, rho)
-        return torch.view_as_real(drho)
-
-    y0 = torch.view_as_real(rho0)
-    t = torch.from_numpy(tlist)
-    options = {'step_size': options['step_size']}
-
-    ys = torchdiffeq.odeint(f, y0, t, method='rk4', options=options)
-    return torch.view_as_complex(ys)
+    H, jump_ops, rho0, tsave, solver=Rouchon(), sensitivity='autograd',
```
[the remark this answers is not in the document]
gautierronan: Yes I tried to stay as simple as possible for this. But it for sure should be added later.

torchqdynamics/mesolve.py
gautierronan: I was thinking that `stepclass` should be either `fixed` or `adaptive` here. It's just an information so that we know whether we should dispatch to the `fixed_odeint` or `adaptive_odeint` later on. In your solver options PR you have the two separate classes for fixed and adaptive, so we could use just the parent class name instead of a string class argument?

For time-dependence, I think it should be implemented directly in the QSolver. Something like `MERouchon1` and `MERouchon1_TD` maybe? Or it could be the same class, but just pre-compute the Kraus operators in the `Qsolver.__init__`?

abocquet: +1 on the separate classes if needed, but I thought the Rouchon method could only support fixed time steps (or time dependent but fixed in advanced).

If all fixed timestep method support both a simple dt and a time depedent timestep then we could simply say that the `step` argument accepts a float or a list of floats?

gautierronan: No that's not what I meant the `stepclass` argument for. I meant it to be used here: https://github.com/PierreGuilmin/torchqdynamics/blob/f00ac9e980515c434891cc2c394e81eecbbd2198/torchqdynamics/odeint.py#L31-L34
just to dispatch to the related fixed or adaptive odeint solver. It's unrelated to time-dependence or varying dt.

abocquet: Yes, agreed. Are there methods that can be both used with a fixed timestep or an adaptative one?
For instance I think that Rouchon method is fixed only and DOPRI6 is adaptative only?

gautierronan: No I think it's always one or the other.

abocquet: Then can't be `stepclass` inferred from the schema?

gautierronan: Yes but I didn't know how to, and I also wanted a simple implementation for this PR. But it should be implemented properly from your PR instead.

abocquet: Ok good to know, I just wanted to make sure I didn't miss something.
Let's do this after we merge https://github.com/PierreGuilmin/torchqdynamics/pull/10

abocquet (approved):

torchqdynamics/mesolve.py
```diff
@@ -1,63 +1,102 @@
...
+    def forward(self, t, dt, rho):
+        """Compute rho(t+dt) using a Rouchon method of order 2"""
+        # Non-hermitian Hamiltonian at time t
+        H_nh = self.H(t) - 0.5j * self.sum_nojump
+
+        # Build time-dependent Kraus operators
+        # TODO: Add the missing time derivative term in -0.5j * dt**2 * \dot{H}
+        M0 = self.I - 1j * dt * H_nh - 0.5 * dt**2 * H_nh @ H_nh
+        M1s = 0.5 * self.Ls @ M0
+        M1s += M1s.adjoint()
```
gautierronan, line 79: (NB: Bug here needs to be fixed. Line 79 should be in line 78 without the adjoint)

torchqdynamics/odeint.py
```diff
@@ -0,0 +1,95 @@
...
+
+def odeint(qsolver, y0, tsave, sensitivity='autograd', model_params=None):
+    # Check arguments
+    tsave = check_tsave(tsave)
+    if (model_params is not None) and (sensitivity in [None, 'autograd']):
+        warnings.warn('Argument `model_params` was supplied in `odeint` but not used.')
+
+    # Dispatch to appropriate odeint subroutine
+    if sensitivity is None:
+        return odeint_inplace(qsolver, y0, tsave)
```
gautierronan, line 18: A good solution here might be to simply use
```python
if sensitivity is None:
    with torch.no_grad():
        return odeint_main(qsolver, y0, tsave)
```
This way torch does not store gradients. It still might be slower than an in-place solver, but it should give a performance boost.

PierreGuilmin (dismissed):
Thanks a lot for this great PR @gautierronan! It looks like a very good starting point for the rest 🥳 I did mostly minor changes, I advise you to check the commits one by one. I am happy to discuss anything that looks weird.

Various comments:
- I merged the `main` branch after having accepted #10. The recommended workflow is to directly base your work on top of #10 instead of implementing dummy code to fake it 😉 For that you just need to checkout the branch `abocquet/solvers-options` of #10, then create a new branch from there, and then when opening a PR on GitHub ask to merge into the branch `abocquet/solvers-options` instead of `main`. Then when we merge #10 GitHub will automatically change the target branch of the PR to the `main` branch.
- Commit 758f8e752cf095d2d83a6caf8d77c5e3c43993e8 (hover to see commit title) fixes various style issues with the comments (to match common Python style of the main libraries), namely
  - Remove file comments (the docstrings and file names/organisations **must be** sufficient).
  - Inline comments start with a lowercase, except if it's a long explanation needing multiple sentences spanning multiple lines.
  - Add final dots to docstring comments.
  - Add inline code formatting when needed.
- 1ab48f55373bf2b21aa8dc2f0de086d6b2e98cbc because the second order is bugged as you pointed out, I'd rather have this in a separate clean PR.

PierreGuilmin:
I'm good to merge this PR if you agree @gautierronan and @abocquet, and then we can port the nice extras from Adrien's Rouchon solver implementation (#5).

PierreGuilmin (approved):
Let's go 🥳

PR #13, opened 2023-03-09
Add ME Rouchon of order 1.5 and 2

gautierronan:
Order 1.5 corresponds to @abocquet's PR https://github.com/PierreGuilmin/torchqdynamics/pull/5.

### DIFF
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -61,7 +61,7 @@ class MERouchon1(MERouchon):
 
         # compute rho(t+dt)
         rho = kraus_map(rho, Ms)
-        return rho / trace(rho).real
+        return rho / trace(rho)[..., None, None].real
 
     def forward_adjoint(self, t, dt, phi):
         raise NotImplementedError
@@ -108,7 +108,7 @@ class MERouchon2(MERouchon):
         # compute rho(t+dt)
         rho_ = kraus_map(rho, M1s)
         rho = kraus_map(rho, M0[None, ...]) + rho_ + 0.5 * kraus_map(rho_, M1s)
-        return rho / trace(rho).real
+        return rho / trace(rho)[..., None, None].real
 
     def forward_adjoint(self, t, dt, phi):
         raise NotImplementedError
@@ -131,9 +131,8 @@ def kraus_map(rho, operators):
 
 
 def trace(rho):
-    """Compute the batched trace of a tensor over its last two dimensions, and
-    return a tensor of the same number of dimensions as rho."""
-    return torch.einsum('...ii', rho)[..., None, None]
+    """Compute the batched trace of a tensor over its last two dimensions."""
+    return torch.einsum('...ii', rho)
 
 
 def inv_sqrtm(mat):
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -5,7 +5,7 @@ from .solver import Rouchon
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver=Rouchon(), sensitivity='autograd', parameters=None
+    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
 ):
     if solver is None:
         # TODO: The default dt should not be choosen in such an arbitrary
@@ -110,8 +110,8 @@ class MERouchon2(MERouchon):
         """Compute rho(t+dt) using a Rouchon method of order 2.
 
         NOTE: For fast time-varying Hamiltonians, this method is not order 2 because the
-        second-order time derivative term is neglected. This term should be added in the
-        zero-th order Kraus operator, as M0 += -0.5j * dt**2 * \dot{H}.
+        second-order time derivative term is neglected. This term could be added in the
+        zero-th order Kraus operator if needed, as M0 += -0.5j * dt**2 * \dot{H}.
         """
         # non-hermitian Hamiltonian at time t
         H_nh = self.H(t) - 0.5j * self.sum_nojump
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -133,7 +133,7 @@ def kraus_map(rho, *operators):
 def btrace(rho):
     """Compute the batched trace of a tensor over its last two dimensions, and return a
     tensor of the same number of dimensions as rho."""
-    return torch.einsum('...ii', rho).real[(..., ) + (None, ) * 2]
+    return torch.einsum('...ii', rho).real[..., None, None]
 
 
 def inv_sqrtm(mat):
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -8,9 +8,7 @@ def mesolve(
     H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
 ):
     if solver is None:
-        # TODO: The default dt should not be choosen in such an arbitrary
-        # fashion, which depends on the time unit used by the user.
-        # Will be replaced by adaptive time step solver when implemented.
+        # TODO: Replace by adaptive time step solver when implemented.
         solver = Rouchon(dt=1e-2)
 
     # define the QSolver
@@ -63,8 +61,7 @@ class MERouchon1(MERouchon):
             M0 @ rho @ M0.adjoint() + dt *
             (self.jump_ops @ rho.unsqueeze(0) @ self.jumpdag_ops).sum(dim=0)
         )
-        rho = rho / rho.trace().real
-        return rho
+        return rho / btrace(rho)
 
     def forward_adjoint(self, t, dt, phi):
         raise NotImplementedError
@@ -79,7 +76,7 @@ class MERouchon1_5(MERouchon):
         # build time-dependent Kraus operators
         M0 = self.I - 1j * dt * H_nh
         S = M0.adjoint() @ M0 + dt * self.sum_nojump
-        S_inv_sqrtm = self.inv_sqrtm(S)
+        S_inv_sqrtm = inv_sqrtm(S)
 
         # compute rho(t+dt)
         rho = S_inv_sqrtm @ rho @ S_inv_sqrtm.adjoint()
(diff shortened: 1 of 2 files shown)


abocquet:
I'm refactoring the code due to all the previous merges. Can we wait for these changes to be merged before we add this scheme? I'll push a first version soon

gautierronan:
Of course ! Would you prefer me to remove the order 1.5 from this PR and keep only the order 2? I tried to stay as close to your implementation, but if you have other ideas please do :)

abocquet:
No it's perfect, actually I might still your code and directly include it into my PR if you don't mind 🥷

PierreGuilmin:
Alright, so I'll maybe wait for @abocquet PR before reviewing this one 😇

torchqdynamics/mesolve.py
```diff
@@ -60,16 +63,71 @@ def forward(self, t, dt, rho):
...
+        # non-hermitian Hamiltonian at time t
+        H_nh = self.H(t) - 0.5j * self.sum_nojump
+
+        # build time-dependent Kraus operators
+        M0 = self.I - 1j * dt * H_nh
+        S = M0.adjoint() @ M0 + dt * self.sum_nojump
+        S_inv_sqrtm = self.inv_sqrtm(S)
+
+        # compute rho(t+dt)
+        rho = S_inv_sqrtm @ rho @ S_inv_sqrtm.adjoint()
```
PierreGuilmin, line 85: This is a quite un-natural way to write the normalisation, but I see why you did. Anyway this will be udpated when we specify the code for time-dep vs time-indep H.

PierreGuilmin (changes requested):
I'm ok for this PR, I'll take care of properly merging this work with @abocquet last PR afterwards (namely replacing the `QSolver` class by the `ForwardQSolver` defining the interface needed by `odeint`).

torchqdynamics/mesolve.py
gautierronan: Yes if everything is time-independent, this is not the right way. But in the general time-dependent way it's faster.

torchqdynamics/mesolve.py
```diff
@@ -60,16 +63,71 @@ def forward(self, t, dt, rho):
...
+        rho = (
+            M0 @ rho @ M0.adjoint() + dt *
+            (self.jump_ops @ rho.unsqueeze(0) @ self.jumpdag_ops).sum(dim=0)
+        )
```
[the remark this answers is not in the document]
gautierronan: Isn't this already taken care of by the S_inv_sqrt renormalization?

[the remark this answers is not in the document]
PierreGuilmin: It is up to order 2, but the trace can diverge rather quickly if we don't renormalize so it's necessary to fix it even though we normalized our Kraus operators.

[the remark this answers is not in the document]
gautierronan: I don't think it's up to order 2, is it? If you compute `\sum \tilde{M}_i^dag \tilde{M}_i` where `\tilde{M_i} = M_i S^{-1/2}`, then I think you exactly get 1.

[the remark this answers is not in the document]
gautierronan: What is order 2 is the fact that this Kraus map gives the Lindblad dynamics exactly. But the point of the renormalization is to get an exact trace-perserving Kraus map.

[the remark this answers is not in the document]
PierreGuilmin: That's right, so it's up to numerical precision sorry 👍🏼 We should try again on some example then, I have memories of trace-exploding order 1.5 schemes but maybe what I said is not true. I propose to *not* correct for now then, and we can fix it later if we find it necessary.

torchqdynamics/mesolve.py
```diff
@@ -126,8 +111,26 @@ def forward(self, t, dt, rho):
...
+        return rho / btrace(rho)
 
     def forward_adjoint(self, t, dt, phi):
         raise NotImplementedError
+
+
+def btrace(input):
+    """Compute the batched trace of a tensor over its last two dimensions, and return a
+    tensor of the same number of dimensions as input."""
+    return input.diagonal(dim1=-1, dim2=-2).sum(-1)[(..., ) + (None, ) * 2]
```
PierreGuilmin, line 123: I suggest the use of the beautiful `einsum` 😉 
```python
def btrace(x):
    # x: (..., d, d)
    # -> (..., d)
    return torch.einsum('...ii', x)
```

torchqdynamics/mesolve.py
```diff
@@ -63,8 +61,7 @@ def forward(self, t, dt, rho):
             M0 @ rho @ M0.adjoint() + dt *
             (self.jump_ops @ rho.unsqueeze(0) @ self.jumpdag_ops).sum(dim=0)
         )
-        rho = rho / rho.trace().real
-        return rho
+        return rho / btrace(rho)
```
PierreGuilmin, line 64: I think we should keep the `.real`?

PierreGuilmin (changes requested):

gautierronan:
Note I also added a batched trace function in the last commit, because I think `mesolve` should support batched `rho0` inputs (i.e. tensors of shape (n, N, N)) . This is much faster than solving `mesolve` multiple times in case you have multiple initial values. This is very useful for instance for the full tomography of a gate, or readout simulations, etc.

torchqdynamics/mesolve.py
gautierronan: I have this tendency to avoid `einsum` because I don't trust its performance. I have seen examples where it's very slow compared to normal functions.

gautierronan: But maybe it's only in numpy, I'm not sure

gautierronan: Ok it seems fixed in https://github.com/pytorch/pytorch/issues/32591. I will change for einsum then.

gautierronan:
Alright so I tested einsum for computing `M @ rho @ M.adjoint()`. Here's the results:
```python
def kraus1(Ms, rho):
    return (Ms.unsqueeze(1) @ rho.unsqueeze(0) @ Ms.unsqueeze(1).adjoint()).sum(dim=0)

def kraus2(Ms, rho):
    return torch.einsum('kin,lnm,kjm->lij', Ms, rho, Ms.conj())

N = 100
Ms = torch.randn(3, N, N, dtype=torch.complex128)
rho = torch.randn(5, N, N, dtype=torch.complex128)

print(torch.all(torch.isclose(kraus1(Ms, rho), kraus2(Ms, rho))))
%timeit kraus1(Ms, rho)
%timeit kraus2(Ms, rho)
```
For N = 10 (cpu):
```
kraus1: 104 µs ± 5.4 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
kraus2: 163 µs ± 3.66 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
For N = 100 (cpu):
```
kraus1: 2.34 ms ± 333 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
kraus2: 1.94 ms ± 94 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
For N = 1000 (cpu):
```
kraus1: 1.71 s ± 64.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
kraus2: 1.42 s ± 22.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So it's only slower on smaller sizes, so I will switch to `einsum` implementation if that's fine with you @abocquet @PierreGuilmin.

torchqdynamics/mesolve.py
```diff
@@ -69,7 +88,62 @@ def forward_adjoint(self, t, dt, phi):
...
+    ops_batch = torch.cat(
+        tuple(op if op.ndim == 3 else op.unsqueeze(0) for op in operators)
+    )
+    return torch.einsum('kin,...nm,kjm->...ij', ops_batch, rho, ops_batch.conj())
+
+
+def btrace(rho):
+    """Compute the batched trace of a tensor over its last two dimensions, and return a
+    tensor of the same number of dimensions as rho."""
+    return torch.einsum('...ii', rho).real[..., None, None]
```
PierreGuilmin, line 136: Functions should always do what they say and nothing more, so we should either rename the method (`btrace_real`, `dmtrace`, ...) and edit the doc to say we just take the real part, or move the `.real` elsewhere to avoid possible misuse later.

torchqdynamics/mesolve.py
```diff
@@ -69,7 +88,62 @@ def forward_adjoint(self, t, dt, phi):
...
+    """
+    if any(op.ndim > 3 or op.ndim < 2 for op in operators):
+        raise ValueError(
+            'Kraus operators should be 2-D tensors, or 3-D tensors if batched.'
+        )
+
+    ops_batch = torch.cat(
+        tuple(op if op.ndim == 3 else op.unsqueeze(0) for op in operators)
+    )
+    return torch.einsum('kin,...nm,kjm->...ij', ops_batch, rho, ops_batch.conj())
```
PierreGuilmin, line 130: Niiiice! I love the benchmark, einsum is actually x2 faster on my CPU on this example, even with small matrix sizes. 🤓  I take the chance to give some insights of what I think is good practice for our codebase.

- First, we should be careful with this kind of tricks where the argument tensors can have different size, it feels easier at first but it's most of the time better to make strict functions that don't do fancy stuff in the background. We will use more pytorch tricks here and there (like `x[None,...]`), but the code will be very readable for those who know the basics.
- Then I think we should always indicate when we make a generic function working with tensors the dimensions of the arguments and the returned tensor (including batching). If we do so, I don't think we need to check the tensors dimensions everytime, we should trust the user.
- For the einstein summation, I like the `i, j, k, l,...` indices for the matrix multiplications and `n, m, ...` indices for the repeated dimensions to sum over. For example here writing `'mij,...jk,mkl ->...il'` we see nicely the `ij, jk, kl` matrix multiplication contraction and the `m` is just summed over.
- I think we want `adjoint` and not `conj`?

So I suggest:
```python
def kraus_map(rho, ops):
    """ ... """
    # rho: `(..., d, d)`
    # ops: `(n, d, d)`
    # -> (..., d, d)
    return torch.einsum('mij,...jk,mkl ->...il', ops, rho, ops.adjoint())
```
and in e.g. Rouchon1
```python
ops = torch.cat((M0[None,...], sqrt(dt) * self.jump_ops))
rho = kraus_map(rho, ops)
```

PierreGuilmin (changes requested):

PierreGuilmin (approved):
LGTM 🥳

PR #15, opened 2023-03-14
Support different t_save and t_step + exp_ops

abocquet:
Short term roadmap
- [x] support different t_save and t_step
- [x] add `exp_op
- [ ] Add Rouchon 2 method
- [ ] use solver options
- [ ] Rouchon 1.5
- [ ] Support for richer data return (observations in `smesolve` for instance)

### DIFF
--- a/torchqdynamics/odeint.py
+++ b/torchqdynamics/odeint.py
@@ -1,80 +1,117 @@
 import warnings
+from abc import ABC
+from enum import Enum
+from typing import List
 
 import torch
 
 from .solver import AdaptativeStep, FixedStep
 
 
-def odeint(qsolver, y0, tsave, sensitivity='autograd', parameters=None):
+class ForwardQSolver(ABC):
+    def forward(self, t, dt, rho):
+        pass
+
+
+class GradientAlgorithm(Enum):
+    AUTOGRAD = 'autograd'  # gradient computed by torch
+    NONE = 'none'  # don't compute the gradients
+    ADJOINT = 'adjoint'  # compute the gradient using the adjoint method
+
+
+def odeint(
+    qsolver,
+    y0,
+    t_save: torch.Tensor,
+    t_step: torch.Tensor = None,
+    exp_ops: List[torch.Tensor] = None,
+    save_states: bool = True,
+    gradient_algorithm=GradientAlgorithm.AUTOGRAD,
+):
+    exp_ops = exp_ops or []
     # check arguments
-    tsave = torch.tensor(tsave)
-    check_tsave(tsave)
-    if (parameters is not None) and (sensitivity in [None, 'autograd']):
-        warnings.warn('Argument `parameters` was supplied in `odeint` but not used.')
+    check_t(t_save)
+    t_step is None or check_t(t_step)
 
     # dispatch to appropriate odeint subroutine
-    if sensitivity is None:
-        return odeint_inplace(qsolver, y0, tsave)
-    elif sensitivity == 'autograd':
-        return odeint_main(qsolver, y0, tsave)
-    elif sensitivity == 'adjoint':
-        return odeint_adjoint(qsolver, y0, tsave, parameters)
+    params = (qsolver, y0, t_save, t_step, exp_ops, save_states)
+    if gradient_algorithm == GradientAlgorithm.NONE:
+        return _odeint_inplace(*params)
+    elif gradient_algorithm == GradientAlgorithm.AUTOGRAD:
+        return _odeint_main(*params)
+    elif gradient_algorithm == GradientAlgorithm.ADJOINT:
+        return _odeint_adjoint(*params)
+    else:
+        raise ValueError(f'Gradient algorithm {gradient_algorithm} not defined')
 
 
-def odeint_main(qsolver, y0, tsave):
+def _odeint_main(qsolver, y0, t_save, t_step, exp_ops, save_states):
     if isinstance(qsolver.options, FixedStep):
-        return _fixed_odeint(qsolver, y0, tsave)
+        return _fixed_odeint(qsolver, y0, t_save, t_step, exp_ops, save_states)
     elif isinstance(qsolver.options, AdaptativeStep):
-        return _adaptive_odeint(qsolver, y0, tsave)
+        warnings.warn(
+            't_step argument is not taken into account for adaptative step solvers'
+        )
+        return _adaptive_odeint(qsolver, y0, t_save, exp_ops, save_states)
 
 
-def odeint_inplace(qsolver, y0, tsave):
+# For now we use *args and **kwargs for helper methods that are not implemented to ease the potential API
+# changes that could occur later. When a method is implemented the methods should take the same arguments
+# as all others
+def _odeint_inplace(*args, **kwargs):
     # TODO: Simple solution for now so torch does not store gradients. This
     #       is probably slower than a genuine in-place solver.
     with torch.no_grad():
-        return odeint_main(qsolver, y0, tsave)
+        return _odeint_main(*args, **kwargs)
 
 
-def odeint_adjoint(qsolver, y0, tsave, parameters):
+def _odeint_adjoint(*_args, **_kwargs):
     raise NotImplementedError
 
 
-def _adaptive_odeint(qsolver, y0, tsave):
+def _adaptive_odeint(*_args, **_kwargs):
     raise NotImplementedError
 
 
-def _fixed_odeint(qsolver, y0, tsave):
+def _fixed_odeint(qsolver, y0, t_save, t_step, exp_ops, save_states):
     # initialize save tensor
-    ysave = torch.zeros(len(tsave), *y0.shape).to(y0)
+    y_save = None
+    exp_save = []
+    if save_states:
+        y_save = torch.zeros(len(t_save), *y0.shape).to(y0)
+
     save_counter = 0
-    if tsave[0] == 0.0:
-        ysave[0] = y0
+    if t_save[0] == 0.0:
+        if save_states:
+            y_save[0] = y0
+        exp_save.append([torch.trace(op @ y0) for op in exp_ops])
         save_counter += 1
 
-    # get qsolver fixed time step
-    dt = qsolver.options.dt
-
     # run the ODE routine
-    t, y = 0.0, y0
-    while t < tsave[-1]:
+    y = y0
+    for i, t in t_step:
         # iterate solution
-        y = qsolver.forward(t, dt, y)
-        t = t + dt
+        y = qsolver.forward(t, t[i + 1] - t, y)
 
         # save solution
-        if t >= tsave[save_counter]:
-            ysave[save_counter] = y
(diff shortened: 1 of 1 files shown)


abocquet:
Ok I'll change this

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
+    ADJOINT = 'adjoint'  # compute the gradient using the adjoint method
+
+
+def odeint(
+    qsolver,
+    y0,
+    t_save: torch.Tensor,
+    exp_ops: List[torch.Tensor] = None,
+    save_states: bool = True,
+    gradient_algorithm=GradientAlgorithm.AUTOGRAD,
```
PierreGuilmin, line 29: I don't know if `sensitivity` was a better name or if they are synonyms, maybe @gautierronan knows better?

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
-def _adaptive_odeint(qsolver, y0, tsave):
+def _adaptive_odeint(*_args, **_kwargs):
     raise NotImplementedError
 
 
-def _fixed_odeint(qsolver, y0, tsave):
-    # initialize save tensor
-    ysave = torch.zeros(len(tsave), *y0.shape).to(y0)
+def _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states):
+    # Initialize save tensor
```
PierreGuilmin, line 74: Pls keep lowercase first letter for inline comments (here and everywhere).

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
             save_counter += 1
 
-    return ysave
+    return y_save, exp_save
 
 
-def check_tsave(tsave):
-    """Check that `tsave` is valid (it must be a non-empty 1D tensor sorted in
+def check_t(t):
+    """Check that `t_save` is valid (it must be a non-empty 1D tensor sorted in
```
PierreGuilmin, line 115: We should either make the method generic (i.e. it checks a generic time tensor), and never refer to `tsave` within the method, or explicitly keep the name `check_tsave` IMO.

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
+    y_save = None
+    if save_states:
+        y_save = torch.zeros(len(t_save), *y0.shape).to(y0)
+
+    if len(exp_ops) > 0:
+        exp_save = torch.zeros(len(t_save), len(exp_ops)).to(
+            device=y0.get_device(), dtype=torch.float
+        )
+    else:
+        exp_save = None
```
PierreGuilmin, line 84: It would be better to place `exp_save = None` just above the `if` like you did with `y_save`.

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
-    t, y = 0.0, y0
-    while t < tsave[-1]:
-        # iterate solution
+    # Run the ODE routine
+    y = y0
+    t, t_max = 0, max(t_save)
+    while t < t_max:
+        # Iterate solution
         y = qsolver.forward(t, dt, y)
-        t = t + dt
```
PierreGuilmin, line 62: We need to increment the time smwh.

PierreGuilmin (changes requested):
Only minor comments. Could you fix the PR title too pls?

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
+    AUTOGRAD = 'autograd'  # gradient computed by torch
+    NONE = 'none'  # don't compute the gradients
+    ADJOINT = 'adjoint'  # compute the gradient using the adjoint method
+
+
+def odeint(
+    qsolver,
+    y0,
+    t_save: torch.Tensor,
+    exp_ops: List[torch.Tensor] = None,
```
PierreGuilmin, line 27: Also, we should type this as `Optional`.

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
+class GradientAlgorithm(Enum):
+    AUTOGRAD = 'autograd'  # gradient computed by torch
+    NONE = 'none'  # don't compute the gradients
+    ADJOINT = 'adjoint'  # compute the gradient using the adjoint method
+
+
+def odeint(
+    qsolver,
+    y0,
+    t_save: torch.Tensor,
```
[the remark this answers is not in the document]
abocquet: I prefer this because it's consistent with the general "rule" of using snake case for variable names, and IMHO new exceptions add to the cognitive load. But if I'm the only one we can change CC @gautierronan @eliegenois

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,124 @@
...
+    exp_ops: List[torch.Tensor] = None,
+    save_states: bool = True,
+    gradient_algorithm=GradientAlgorithm.AUTOGRAD,
+):
```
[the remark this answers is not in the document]
gautierronan: `parameters` should be a tuple of nn.Parameters !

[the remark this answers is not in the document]
abocquet: Maybe we can keep it for when we implement the adjoint computation

torchqdynamics/odeint.py
gautierronan: I have seen `sensealg` and `sensitivity`. I think they are the usual terminology. IMO `gradient_algorithm` suggests something like Adam or SGD (a gradient descent alg.). An alternative could be `autodiff_algorithm` or something of this sort.

torchqdynamics/odeint.py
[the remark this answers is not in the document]
gautierronan: I prefer `tsave`, but if there's a general rule that should be followed then I'm fine with `t_save` as well.

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,122 @@
...
+    exp_ops: Optional[List[torch.Tensor]] = None,
+    save_states: bool = True,
+    gradient_algorithm=GradientAlgorithm.AUTOGRAD,
+):
+    # check arguments
+    check_t_save(t_save)
 
-def odeint_main(qsolver, y0, tsave):
+    # dispatch to appropriate odeint subroutine
+    params = (qsolver, y0, t_save, exp_ops, save_states)
```
gautierronan, line 34: Can we keep the terminology `parameters` for the model parameters to compute the gradients on? Maybe replace this with `args`?

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,122 @@
...
-        if t >= tsave[save_counter]:
-            ysave[save_counter] = y
+        if t >= t_save[save_counter]:
+            if save_states:
+                y_save[save_counter] = y
+
+            for j, op in enumerate(exp_ops):
+                exp_save[save_counter, j] = torch.trace(op @ y)
             save_counter += 1
+        t += dt
```
gautierronan, line 107: If `t_max / dt` is not an integer, the solution will be iterated for a longer time than `t_max`. At some point I had put something like
```python
if t + dt > t_max:
   dt = t_max - t
```
at the beginning of the loop to avoid this problem. However, I agree that this is not ideal for time-independent problems for which it's nice to have `dt` constant.

I would argue that the best solution would be to update `dt` to `dt = t_max / math.ceil(t_max / dt)` somewhere in the code. 

Actually, even a stronger criterion requires that all values inside `t_save` are multiples of `dt`. Otherwise, the solution is saved to `y_save[i]` at a different time than `t_save[i]`.

torchqdynamics/odeint.py
```diff
@@ -1,80 +1,122 @@
...
 
-    return ysave
+    return y_save, exp_save
 
 
-def check_tsave(tsave):
-    """Check that `tsave` is valid (it must be a non-empty 1D tensor sorted in
+def check_t_save(tsave):
```
gautierronan, line 112: Inconsistent `tsave` and `t_save` in this function.

torchqdynamics/odeint.py
PierreGuilmin: I think the solution is indeed to require that all values inside `t_save` are multiples of `dt`. However, if one just wants to see the global evolution and doesn't care about details (just plot Wigners for example) he wouldn't mind that `t_save` values are not perfectly aligned (if `dt` is small). IMO we should just make the doc clear to explain this, later. 
 And also I think we don't really care that the scheme is iterated for a tiny bit longer in case the final `t_save` value is not a multiple of `dt`.

gautierronan: I don't really agree with this. It's quite important that we iterate for exactly the right time. If you're simulating a gate and iterate for 1% longer, you get 1% fidelity error, whereas your gate might otherwise have 1e-5 error.

Also, for the adjoint method, I need to run "smaller" odeints between every value of `t_save`. So it's quite important that all values of `t_save` are well aligned with `dt`. So, in my mind there are three solutions: (1) we update `dt` at the beginning of `_odeint_fixed`, (2) we enforce that all values of `t_save` are aligned with `dt` in `check_t_save`, or (3) we allow for the varying `dt` inside the loop, which is dynamically changed at the last time step in the loop.

gautierronan: Maybe one solution would be, inside `check_t_save`, to:
(1) update `dt` such that `dt = t_max / math.ceil(t_max / dt)`
(2) update `t_save` such that `t_save[i] = dt * k_i` for some integer `k_i` (we just take the closest value)
And then:
(3) return the actual `t_save` to the user at the end, within the `result` instance.

PierreGuilmin: I agree with you, I think we should favor the simplest option (2) for now then (which we can do in another PR).

PierreGuilmin: Concerning
> It's quite important that we iterate for exactly the right time. If you're simulating a gate and iterate for 1% longer, you get 1% fidelity error, whereas your gate might otherwise have 1e-5 error.

This is very dependent indeed on the simulation you're running:
- quick qualitative simulation: you don't care about `t_save` being perfectly aligned
- quantitative simulation: you do care

Fixing the `dt` would fit both use cases BUT for testing and benchmarking numerical schemes + user experience I don't think you don't want some dark magic to update the `dt` in the background. Thus option (2) checking that the user chose compatible `dt` and `t_save` is the best choice I think.

gautierronan: Yes I think option (2) is good. Plus it's probably what most people do naturally anyway.
So we can just add inside `check_t_save`:
```python
if torch.any(torch.remainder(t_save, dt) != 0.0):
    raise ValueError('Every value of argument `t_save` must be a multiple of the time step `dt`.')
```
I can do it in another PR, or we add it here, as you wish !

gautierronan: Actually the above solution doesn't work. Can use this instead:
```python
if not torch.all(torch.isclose((tsave / dt).int().float(), tsave / dt)):
```

PierreGuilmin (approved):
Nice!

PR #19, opened 2023-03-17
Fix mesolve to have a minimal working version

PierreGuilmin:


torchqdynamics/mesolve.py
```diff
@@ -1,14 +1,21 @@
...
+from .solver_utils import inv_sqrtm, kraus_map
+from .utils import trace
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
+    H, jump_ops, rho0, tsave, exp_ops: Optional[List[torch.Tensor]] = None, solver=None,
+    sensitivity='autograd', parameters=None
 ):
+    tsave = torch.tensor(tsave)
```
gautierronan, line 16: Should be
```python
if isinstance(tsave, (list, np.ndarray)):
    tsave = torch.tensor(tsave)
```
Otherwise torch raises a warning every time if tsave is already a torch.tensor.

torchqdynamics/mesolve.py
```diff
@@ -1,14 +1,21 @@
...
+from .odeint import AdjointQSolver, odeint
 from .solver import Rouchon
+from .solver_utils import inv_sqrtm, kraus_map
+from .utils import trace
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
+    H, jump_ops, rho0, tsave, exp_ops: Optional[List[torch.Tensor]] = None, solver=None,
+    sensitivity='autograd', parameters=None
```
gautierronan, line 14: I think `sensitivity` should default to `None` (no gradients computed).

torchqdynamics/mesolve.py
```diff
@@ -1,14 +1,21 @@
...
 ):
+    tsave = torch.tensor(tsave)
+    if exp_ops is None:
+        exp_ops = exp_ops
```
gautierronan, line 18: What does this do? Is it a typo?

torchqdynamics/odeint.py
```diff
@@ -83,35 +90,24 @@ def _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states):
...
-    # save first step
-    save_counter = 0
-    if t_save[0] == 0.0:
-        if save_states:
-            y_save[0] = y0
-        for j, op in enumerate(exp_ops):
-            exp_save[save_counter, j] = torch.trace(op @ y0)
-        save_counter += 1
+        exp_save = torch.zeros(len(exp_ops),
+                               len(t_save)).to(device=y0.device, dtype=torch.float)
```
gautierronan, line 94: The `dtype` should not be `torch.float` by default. If you do cat qubits and look at bitflips typically, you want machine precision on the expectation value of the sigma_x invariant. Also, the `dtype` is not necessarely a real-valued element. If you evaluate the annihilation operator a for instance, it will be complex-valued. So I propose taking the same dtype as the `y0`.

torchqdynamics/odeint.py
```diff
@@ -83,35 +90,24 @@ def _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states):
...
+    save_counter = 0
+    for t in tqdm(times):
         # save solution
         if t >= t_save[save_counter]:
             if save_states:
                 y_save[save_counter] = y
-
             for j, op in enumerate(exp_ops):
-                exp_save[save_counter, j] = torch.trace(op @ y)
+                exp_save[j, save_counter] = torch.trace(op @ y).real
```
gautierronan, line 106: Same comment here. It doesn't have to be a real-valued expectation value.

torchqdynamics/odeint.py
```diff
@@ -83,35 +90,24 @@ def _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states):
...
             save_counter += 1
-        t += dt
+
+        # iterate solution
+        y = qsolver.forward(t, dt, y)
```
gautierronan, line 110: The loop is iterated one too many times. I understand you did that to not bother with saving the solution one more time, but I think it's better not to do that (`forward` could be slow, it could diverge after the final time step, or whatever).

So I would either go back to saving at `t=0.0` before the loop, do an automatic saving of `t=tsave[-1]` after the loop (my preferred option), or add a `break` inside the loop when the final time step is reached.

torchqdynamics/mesolve.py
```diff
@@ -1,14 +1,21 @@
...
-from .odeint import odeint
+from .odeint import AdjointQSolver, odeint
 from .solver import Rouchon
+from .solver_utils import inv_sqrtm, kraus_map
+from .utils import trace
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
+    H, jump_ops, rho0, tsave, exp_ops: Optional[List[torch.Tensor]] = None, solver=None,
```
gautierronan, line 13: For consistency with the rest of the code, `tsave` should be `t_save`.

gautierronan (changes requested):

torchqdynamics/mesolve.py
PierreGuilmin: Yes I meant the following sorry, I'll fix it
```suggestion
        exp_ops = []
```

torchqdynamics/odeint.py
PierreGuilmin: Right 👍

PierreGuilmin:
Thanks for the careful review, I love it 🤓

torchqdynamics/odeint.py
PierreGuilmin: 100% agree

gautierronan (approved):

torchqdynamics/mesolve.py
```diff
@@ -1,14 +1,23 @@
...
 from .solver import Rouchon
+from .solver_utils import inv_sqrtm, kraus_map
+from .utils import trace
 
 
 def mesolve(
-    H, jump_ops, rho0, tsave, solver=None, sensitivity='autograd', parameters=None
+    H, jump_ops, rho0, t_save, exp_ops: Optional[List[torch.Tensor]] = None,
+    solver=None, sensitivity=None, parameters=None
 ):
```
gautierronan, line 16: Actually I'm thinking. Do we want the following function signature instead?
```python
def mesolve(H, jump_ops, rho0, t_save, *, exp_ops=None, solver=None, sensitivity=None, parameters=None)
```
(with the star between args and kwargs arguments)
Pytorch does it, it forces keyword-only arguments after the star. I think it's nice to avoid mistakes in the arguments ordering by users.

PR #21, opened 2023-03-18
Improve readability of mesolve/odeint

gautierronan:
Here is a small PR with a few small modifications on mesolve/odeint before we move on to bigger things. I am building on https://github.com/PierreGuilmin/torchqdynamics/pull/19 and also a bit on https://github.com/PierreGuilmin/torchqdynamics/pull/16. The rationale behind the changes is:

- Added `expect` function to allow batching over `y0`.
- Modified the `times` definition in `fixed_odeint` for robustness. If `dt = 0.099999` and `t_save[-1] =1`, one too many time step was defined by the previous solution.
- Added `save_state` argument all the way up in `mesolve` definition.
- Added a star * in `mesolve` definition to force keyword arguments.
- Added typing in `mesolve` and `odeint` with the latest Python 3.10 format (e.g. `list` instead of `typing.List`, and `x | y` instead of `Union[x,y]`). 
- Finally, on a more controversial topic (sorry I'm annoying @abocquet), I removed the `AutoDiffAlgorithm` enumeration, because I think it was cluttering the readability for not much gained. Plus the user has to import and define an `AutoDiffAlgorithm` upon launching `mesolve` which I think is not a good thing. Unless I'm missing something you intended, I would argue to remove it.

### DIFF
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -19,7 +19,7 @@ def mesolve(
     *,
     exp_ops: list[torch.Tensor] = None,
     save_states: bool = True,
-    autodiff: str = None,
+    autodiff_alg: str = None,
     parameters: tuple[nn.Parameter, ...] = None,
     solver: SolverOption = None,
 ):
@@ -43,7 +43,7 @@ def mesolve(
         raise NotImplementedError
 
     # compute the result
-    return odeint(qsolver, rho0, t_save, exp_ops, save_states, autodiff)
+    return odeint(qsolver, rho0, t_save, exp_ops, save_states, autodiff_alg)
 
 
 class MERouchon(AdjointQSolver):
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -1,5 +1,5 @@
 from math import sqrt
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -1,26 +1,27 @@
 from math import sqrt
-from typing import List, Optional
+from typing import Callable
 
 import numpy as np
 import torch
+import torch.nn as nn
 
 from .odeint import AdjointQSolver, odeint
-from .solver import Rouchon
+from .solver import Rouchon, SolverOption
 from .solver_utils import inv_sqrtm, kraus_map
 from .utils import trace
 
 
 def mesolve(
-    H,
-    jump_ops,
-    rho0,
-    t_save,
+    H: torch.Tensor | Callable[[float], torch.Tensor],
+    jump_ops: list[torch.Tensor],
+    rho0: torch.Tensor,
+    t_save: torch.Tensor,
     *,
-    exp_ops: Optional[List[torch.Tensor]] = None,
+    exp_ops: list[torch.Tensor] = None,
     save_states: bool = True,
-    solver=None,
-    sensitivity=None,
-    parameters=None,
+    autodiff: str = None,
+    parameters: tuple[nn.Parameter, ...] = None,
+    solver: SolverOption = None,
 ):
     if isinstance(t_save, (list, np.ndarray)):
         t_save = torch.tensor(t_save)
@@ -42,7 +43,7 @@ def mesolve(
         raise NotImplementedError
 
     # compute the result
-    return odeint(qsolver, rho0, t_save, exp_ops, save_states)
+    return odeint(qsolver, rho0, t_save, exp_ops, save_states, autodiff)
 
 
 class MERouchon(AdjointQSolver):
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -1,5 +1,5 @@
 from math import sqrt
-from typing import Callable
+from typing import Callable, List, Optional, Tuple
 
 import numpy as np
 import torch
@@ -12,16 +12,16 @@ from .utils import trace
 
 
 def mesolve(
-    H: torch.Tensor | Callable[[float], torch.Tensor],
-    jump_ops: list[torch.Tensor],
+    H: Union[torch.Tensor, Callable[[float], torch.Tensor]],
+    jump_ops: List[torch.Tensor],
     rho0: torch.Tensor,
     t_save: torch.Tensor,
     *,
-    exp_ops: list[torch.Tensor] = None,
+    exp_ops: Optional[List[torch.Tensor]] = None,
     save_states: bool = True,
-    autodiff_alg: str = None,
-    parameters: tuple[nn.Parameter, ...] = None,
-    solver: SolverOption = None,
+    gradient_alg: Optional[str] = None,
+    parameters: Optional[Tuple[nn.Parameter, ...]] = None,
+    solver: Optional[SolverOption] = None,
 ):
     if isinstance(t_save, (list, np.ndarray)):
         t_save = torch.tensor(t_save)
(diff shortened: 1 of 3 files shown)


torchqdynamics/odeint.py
```diff
@@ -94,15 +88,16 @@ def _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states):
 
     # run the ode routine
     y = y0
-    times = torch.arange(0.0, t_save[-1], dt)
+    num_times = round(t_save[-1].item() / dt) + 1
+    times = torch.linspace(0.0, t_save[-1], num_times)
```
PierreGuilmin, line 92: > Modified the `times` definition in `fixed_odeint` for robustness. If `dt = 0.099999` and `t_save[-1] =1`, one too many time step was defined by the previous solution.

I disagree with this change. The proposed solution **does not** respect the `dt` asked by the user, it tries to fix it because `t_save` does not respect the condition that all elements are multiples of `dt`. This all comes from the fact that `allclose` has a limited `rtol` at `1e-5` (which we can change if you want to be more picky).

```python
torch.set_printoptions(precision=10)

dt = 0.0999999
t_save = torch.tensor([0.0, 0.5, 1.0])

# previous code: dt is correct but the last time step goes too far
times = torch.arange(0.0, t_save[-1], dt)
print(times)
tensor([0.0000000000, 0.0999998972, 0.1999997944, 0.2999997139, 0.3999995887,
        0.4999994934, 0.5999994278, 0.6999992728, 0.7999991775, 0.8999990821,
        0.9999989867])

# proposed change: the last time step is just right but dt is incorrect
num_times = round(t_save[-1].item() / dt) + 1
times = torch.linspace(0.0, t_save[-1], num_times)
print(times[:-1])
tensor([0.0000000000, 0.1000000015, 0.2000000030, 0.3000000119, 0.4000000060,
        0.5000000000, 0.6000000238, 0.6999999881, 0.8000000119, 0.8999999762])
```

IMO we should keep the first version to guarantee that `dt` is the one asked for, and if you want we can be more strict on `t_save` values being exact multiples of `dt` (up to `float32` numerical precision for example).

PierreGuilmin (changes requested):
It's good to see all these improvements!

- Concerning the typing I think we should stay compatible with at least Python >=3.8 (see https://devguide.python.org/versions/ and https://www.jetbrains.com/lp/devecosystem-2022/python/) so we should IMO **not** use the new typing features, even if they are nice.
- I'm favourable to the removal of the `AutoDiffAlgorithm` enumeration. Concerning the naming, is the term *automatic differentiation* containing both the *automatic gradient* and *gradient by the adjoint* methods? (I'm a bit scared the naming will introduce a confusion btw _automatic differentiation_ and _automatic gradient_). Maybe _sensitivity_ like we had first would be better, or _gradient_alg_ which would talk to the most I guess (everyone knows what's the gradient, but not everyone knows the meaning of autodiff, adjoint, autograd and sensitivity).

torchqdynamics/odeint.py
gautierronan: I understand what you mean, but going one step too far is really bad behavior. I believe it's much better to make a 1e-5 mistake on the dt required by the user, than making a `dt` error on the total time of evolution.

torchqdynamics/mesolve.py
```diff
@@ -1,18 +1,27 @@
...
+    H: torch.Tensor | Callable[[float], torch.Tensor],
+    jump_ops: list[torch.Tensor],
+    rho0: torch.Tensor,
+    t_save: torch.Tensor,
+    *,
+    exp_ops: list[torch.Tensor] = None,
+    save_states: bool = True,
+    autodiff_alg: str = None,
+    parameters: tuple[nn.Parameter, ...] = None,
+    solver: SolverOption = None,
```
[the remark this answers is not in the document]
gautierronan: Alright, I will just go to Python 3.8 typing and leave the rest.

torchqdynamics/odeint.py
gautierronan: FYI all this is due to the imperfect behavior of torch.linspace (which will be called by the user):
```python
import torch
torch.set_printoptions(precision=10)
t_save = torch.linspace(0, 1, 11)
print(t_save)
# tensor([0.0000000000, 0.1000000015, 0.2000000030, 0.3000000119, 0.4000000060,
#         0.5000000000, 0.6000000238, 0.6999999881, 0.8000000119, 0.8999999762,
#         1.0000000000])
```

torchqdynamics/odeint.py
```diff
@@ -79,7 +74,7 @@ def _adaptive_odeint(*_args, **_kwargs):
 
 
 def _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states):
-    if not torch.all(torch.isclose(torch.round(t_save / dt), t_save / dt)):
+    if not torch.allclose(torch.round(t_save / dt), t_save / dt):
```
PierreGuilmin, line 77: Could we add the following comment on top:
```python
    # assert that `t_save` values are multiples of `dt` (with the default
    # `rtol=1e-5` they differ by at most 0.001% from a multiple of `dt`)
```

torchqdynamics/odeint.py
PierreGuilmin: You convinced me, you're right 👍 Could we add the following comment just below? (we will have forgotten all of this in six months, I think it's important to explain when there are some quirks)
```python
    # Note that defining the solver times as `torch.arange(0.0, t_save[-1], dt)`
    # could result in an error of order `dt` in the save time (e.g. for
    # `t_save[-1]=1.0` and `dt=0.999999` -> `times=[0.000000, 0.999999]`, so
    # the final state would be saved one time step too far at
    # `0.999999 + dt ~ 2.0`). The previous definition ensures that we never
    # perform an extra step, so the error on the correct save time is of
    # `0.001% * dt` instead of `dt`.
```

PierreGuilmin (changes requested):

PierreGuilmin (approved):
Nice!! Please **don't merge** in the target branch `fix-mesolve`, we'll wait for `fix-mesolve` to be merged in `main`, which will automatically change the target of this branch to `main` and then we will merge this PR to `main`. This is called cascading PRs, by doing this each PR will appear in the `main` branch commit history, while if we would merge this PR in `fix-mesolve` directly and then merge `fix-mesolve` in `main`, both PRs would appear as a single commit.

abocquet:
Nice PR!
To address the enumeration: the user **does not** necessarily have to use it. They can simply put the string value, which is simpler indeed. But internally, it really fixes the ideas for which values are acceptable or not and make refactoring easier and more reliable later

abocquet:
Also I'm not sure how I'm feeling about staying compatible with python 3.8. Could you explicit which typing features you have in mind? If they are available from python 3.11, it's probably best not to use them, if it's python 3.9 there shouldn't be any problem, since by the time we generally release it'll have gained traction

PierreGuilmin:
Wooops sorry, because we're doing "squash and merge" this didn't go as expected, let met fix it.

PierreGuilmin (approved):

PierreGuilmin:
> Wooops sorry, because we're doing "squash and merge" this didn't go as expected, let met fix it.

PierreGuilmin:
> Wooops sorry, because we're doing "squash and merge" this didn't go as expected, let met fix it.

To be more explicit, when squashing and merging the target branch `A` into `main` a new single commit is created on `main` gathering all the changes, and the target branch for this PR switches from `A` to `main`. But then because all the common commits with `A` are not on `main` anymore, they reappear in the PR. The (annoying) solution is to rebase on `main` by selectively picking only the new commits. Do you have any nice solution to this @abocquet? (with the constraints to keep the "Squash and merge", I think it fits nicely our workflow with commits that are a bit messy in each PR).

PierreGuilmin:
Let me merge this and the subsequent PRs and keep aside the discussion on the use of enumeration for our next call.

PR #22, opened 2023-03-18
Add basic typing

PierreGuilmin:
Very basic fixes and type aliases.

A few comments:
- Precisely defining the acceptable `List` type for `TensorLike` requires recursive types and all (for nested lists), I don't think we should go into such trouble.
- We will need to replace `torch.Tensor` by `QTensor` in many places when the class will be ready, I prefer to keep the original torch tensor for now.
- @gautierronan what's the typing for `phi` in `forward_adjoint`?
- Concerning the type aliases, we're not really used to them in typical libraries, but IMO it really makes the big functions signatures like `mesolve` much more readable, and we can use them in all solvers instead of copy pasting lengthy union types like `Union[torch.Tensor, Callable[[float], torch.Tensor]]`. I think for users unaccustomed to Python typing, seeing `TimeDependentOperator` is less scary even if the exact typing is hidden, and we just have to make sure that the doc explains clearly what are the types accepted for time-dependent Hamiltonian.
- Following @gautierronan's PR `odeint` only accepts `torch.Tensor` and not `TensorLike` types for example, assuming all the conversions are done by the callers (i.e. `ssolve`, `mesolve`, etc...). In a way we consider it as a backend method, and we only work with torch tensors in the back.

### DIFF
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -72,7 +72,7 @@ class MERouchon1(MERouchon):
         rho = kraus_map(rho, Ms)
         return rho / trace(rho)[..., None, None].real
 
-    def forward_adjoint(self, t: float, dt: float, phi):
+    def forward_adjoint(self, t: float, dt: float, phi: torch.Tensor):
         raise NotImplementedError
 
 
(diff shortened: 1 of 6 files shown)


torchqdynamics/mesolve.py
```diff
@@ -70,12 +72,12 @@ def forward(self, t, dt, rho):
         rho = kraus_map(rho, Ms)
         return rho / trace(rho)[..., None, None].real
 
-    def forward_adjoint(self, t, dt, phi):
+    def forward_adjoint(self, t: float, dt: float, phi):
```
gautierronan, line 75: `phi` is also a `torch.Tensor`.

gautierronan (approved):
Looks good to me !

PR #28, opened 2023-03-22
Batching of `H` and `rho0`

PierreGuilmin:
Here is a proposed implementation for the batching. `H` is assumed time-independent in order not to mess with @abocquet work. I suggest reading commit by commit.

- The proposed philosophy is to have `odeint` agnostic of the batching (that is `y0` can be arbitrarily batched `(..., m, n)`), because the code is fundamentally unaware of the problem structure. This is necessary because it will be used by `ssolve` for which `m=d, n=1` and by `smesolve` for which we will probably batch on the number of trajectories. So I don't think we can have a homogeneous batching for all these guys, thus the need for `odeint` to handle any batch sizes. It's more of a proposition, I don't know if it's doable, but it would be nice. This implies any operation in `odeint` should be any-batch-size compatible (which is the case today).
- The tensors `H` and `rho0` can be batched or not when calling `mesolve`, and the returned result is batched accordingly, as discussed together on Slack.
- For the batching of `H` I think we have two options to compute `M0` and all, either we copy the jump operators along a new dimension of the batch size of `H` (like I did here, it consumes a little moremmemory but no computation time I think), either we write `einsum` everywhere to handle the extra dimension of `H`. Do you see a better alternative?
- I propose the following bunch of convention for tensor shapes and dimensions: we denote tensor shapes between `()`, favor lowercase one-letter dimension, use `b` or `b_...` for batch size, use `m`, `n` for unknown final two dimensions and `n, n` if it's a square matrix, use `i,j,k,l` for matrix indices in einsum, favor the use of `.shape` over `.size()` except when we need a specific dimension then we use e.g. `.size(2)`.


gautierronan:
I don't have time to do a full review yet. But just as a first remark from quickly reading through your PR, it seems like batching over `rho0` is quite simple (i.e. your first three commits), but batching over `H` adds a lot of complexity and memory cost (the fourth commit). Should we just do `rho0` batching for now, and implement the `H` batching way later?

It's just a suggestion, but I will also try to think more carefully about H batching when I do the full review.

torchqdynamics/mesolve.py
```diff
@@ -81,20 +121,31 @@ def forward_adjoint(self, t: float, dt: float, phi: torch.Tensor):
...
-        Ms = torch.cat((M0[None, ...], sqrt(dt) * self.jump_ops))
+        M0 = self.I - 1j * dt * H_nh  # (b_H, 1, n, n)
+        Ms = sqrt(dt) * self.jump_ops  # (1, len(jump_ops), n, n)
 
         # build normalization matrix
-        S = M0.adjoint() @ M0 + dt * self.sum_nojump
-        S_inv_sqrtm = inv_sqrtm(S)
+        S = M0.adjoint() @ M0 + dt * self.sum_nojump  # (b_H, 1, n, n)
+        # TODO: fix `inv_sqrtm` (size not compatible and linalg.solve RuntimeError)
+        S_inv_sqrtm = inv_sqrtm(S)  # (b_H, 1, n, n)
```
PierreGuilmin, line 140: The `inv_sqrtm` function was bugged, I propose to fix this in a later PR. Then we also need to make it so it handles properly the new tensor shape.
```python
RuntimeError: linalg.solve: Expected A and B to have the same dtype, but found A of type ComplexDouble and B of type Double instead
```

torchqdynamics/solver_utils.py
```diff
@@ -1,20 +1,21 @@
...
-        rho: Density matrix of shape (..., n, n).
-        operators: Kraus operators of shape (b, n, n).
+        rho: Density matrix of shape (a, ..., n, n).
+        operators: Kraus operators of shape (a, b, n, n).
     Returns:
-        Density matrix of shape (..., n, n) with the Kraus map applied.
+        Density matrix of shape (a, ..., n, n) with the Kraus map applied.
     """
-    return torch.einsum('mij,...jk,mkl->...il', operators, rho, operators.adjoint())
+    # TODO: fix doc
```
PierreGuilmin, line 17: Once you agree overall I will fix the doc (specify what are `a` and `b`, add the fact that the kraus map is not normalized).

torchqdynamics/mesolve.py
PierreGuilmin: I will soon add tests to `mesolve` because it feels necessary.

torchqdynamics/solver_utils.py
PierreGuilmin: Ho and the renaming of `operators` to `O` is a proposed new convention: when objects are batched we don't use the plural or prefix them with `b` or `batched`, we just use the singular regular name (like we do for `rho`) and the doc tells what's the size. I also propose to use `O` for any generic operator argument.

torchqdynamics/mesolve.py
```diff
@@ -23,6 +23,21 @@ def mesolve(
...
+    #     H: (b_H?, n, n)
+    #     rho0: (b_rho0?, n, n)
+    #
+    # Returns:
+    #     (y_save, exp_save) with
+    #     - y_save: (b_H?, b_rho0?, len(t_save), n, n)
+    #     - exp_save: (b_H?, b_rho0?, len(exp_ops), len(t_save))
+
+    # batch H and rho0 by default
+    H_batched = H[None, ...] if H.dim() == 2 else H
```
abocquet, line 36: Nit: the use of `unsqueeze` would be more consistent with `squeeze` below

abocquet:
It's nice that you put tensors dimension in comment everywhere 👍

torchqdynamics/mesolve.py
PierreGuilmin: Right, I chose the current syntax because it's what we do in `odeint`, I think it's easier to understand what happens like that than with `unsqueeze`. Unfortunately, I don't think there's such a shorthand syntax for `squeeze`.

PierreGuilmin: Actually there is, we could do `y_save[0]` but it's far less expressive than `y_save.squeeze(0)` ...

torchqdynamics/mesolve.py
gautierronan: Yes sorry I have to admit I coded without checking too much. I was like "it's fine we'll make some test function at some point and debug everything". I don't know if that's a good way of working or not haha.

torchqdynamics/mesolve.py
gautierronan: I agree that we should use `squeeze` and `[None, ...]`. They are the most readable forms.

torchqdynamics/solver_utils.py
gautierronan: Ok i'm fine with this convention.

torchqdynamics/mesolve.py
```diff
@@ -35,44 +50,69 @@ def mesolve(
...
-        self.sum_nojump = (self.jumpdag_ops @ self.jump_ops).sum(dim=0)
-        self.I = torch.eye(H(0).shape[-1]).to(H(0))
+        # Args:
+        #     H: (b_H, n, n)
+
+        # we convert H and jump_ops to compatible sizes, i.e. tensors of size
+        # (a, b, n, n) where a is b_H or 1 and b is len(jump_ops) or 1
+        self.H = H[:, None]  # (b_H, 1, n, n)
+        self.jump_ops = jump_ops[None, ...]  # (1, len(jump_ops), n, n)
+        self.sum_nojump = (jump_ops @ jump_ops.adjoint()).sum(dim=0)  # (n, n)
```
gautierronan, line 87: The `adjoint()` should be on the first operator and not the second.

torchqdynamics/mesolve.py
```diff
@@ -35,44 +50,69 @@ def mesolve(
...
-        self.jump_ops = jump_ops
-        self.jumpdag_ops = jump_ops.adjoint()
-        self.sum_nojump = (self.jumpdag_ops @ self.jump_ops).sum(dim=0)
-        self.I = torch.eye(H(0).shape[-1]).to(H(0))
+        # Args:
+        #     H: (b_H, n, n)
+
+        # we convert H and jump_ops to compatible sizes, i.e. tensors of size
+        # (a, b, n, n) where a is b_H or 1 and b is len(jump_ops) or 1
+        self.H = H[:, None]  # (b_H, 1, n, n)
```
gautierronan, line 85: Can we write `self.H = H[:, None, :, :]` for readability? Or `self.H = H[:, None, ...]`.

torchqdynamics/mesolve.py
```diff
@@ -35,44 +50,69 @@ def mesolve(
...
-        self.H = H
-        self.jump_ops = jump_ops
-        self.jumpdag_ops = jump_ops.adjoint()
-        self.sum_nojump = (self.jumpdag_ops @ self.jump_ops).sum(dim=0)
-        self.I = torch.eye(H(0).shape[-1]).to(H(0))
+        # Args:
+        #     H: (b_H, n, n)
+
+        # we convert H and jump_ops to compatible sizes, i.e. tensors of size
+        # (a, b, n, n) where a is b_H or 1 and b is len(jump_ops) or 1
```
gautierronan, line 84: `# (b_H, len(jump_ops), n, n)`
No need to introduce (a,b) notations I think. It's confusing here since you don't use it later.

torchqdynamics/mesolve.py
```diff
@@ -35,44 +50,69 @@ def mesolve(
...
+
         # non-hermitian Hamiltonian at time t
-        H_nh = self.H(t) - 0.5j * self.sum_nojump
+        H_nh = self.H - 0.5j * self.sum_nojump  # (b_H, 1, n, n)
 
         # build time-dependent Kraus operators
-        M0 = self.I - 1j * dt * H_nh
-        Ms = torch.cat((M0[None, ...], sqrt(dt) * self.jump_ops))
+        M0 = self.I - 1j * dt * H_nh  # (b_H, 1, n, n)
+        Ms = sqrt(dt) * self.jump_ops  # (1, len(jump_ops), n, n)
```
gautierronan, line 107: I would define it as `M1s` instead of `Ms`, like in the MERouchon2

gautierronan (changes requested):
Thanks for the great PR. I really like where this is going ! Just a few small comments and its good for me.

PR #29, opened 2023-03-23
Convert `jump_ops` to a tensor and raise error if empty list

PierreGuilmin:


PierreGuilmin:
Because then one should use `ssolve` no?

gautierronan (approved):
I agree that since `jump_ops` is a mandatory argument in `mesolve`, it should be non-empty. Otherwise the user should use `ssolve` which should accept both density matrices and pure states. Although we can implement just pure states for now, since I think `ssolve` with mixed states is a bit more niche. 

Also, I believe we should not be redirecting from `mesolve` to `ssolve` like qutip does. It clutters the code and the understanding.

PR #31, opened 2023-03-23
Accept qutip `QObj` in `mesolve` and ket for `rho0`

PierreGuilmin:
This is a game changer for adoption, our mesolve can now be a drop-in replacement for qutip mesolve.

<details>
  <summary>Spiraling cavity example</summary>
  
  ```python
# parameters
d = 32
kappa = 1.5
delta = 4 * np.pi
alpha0 = 1.0 + 1.0j

# operators
a = qt.destroy(d)
adag = a.dag()
H = delta * adag * a
jump_ops = [np.sqrt(kappa) * a]
exp_ops = [(a+adag)/np.sqrt(2), (a-adag)/(np.sqrt(2) * 1j)]

# other arguments
rho0 = qt.coherent(d, alpha0)
tsave = np.linspace(0.0, 1.0, 11)
solver = tq.solver.Rouchon(dt=1e-4, order=1)

# solve
states, exp = tq.mesolve(H, jump_ops, rho0, tsave, exp_ops=exp_ops, solver=solver)

# result
print(states.shape)
print(exp.shape)
```
</details>

Note that will also make test writing a breeze, we can focus on physical examples.

I think we maybe also need an optional layer of conversion back from `torch.Tensor` to `qutip.Qobj` for the returned objects, depending on a parameter set by the user?

### DIFF
--- a/torchqdynamics/types.py
+++ b/torchqdynamics/types.py
@@ -1,4 +1,4 @@
-from typing import Callable, List, Union, get_args
+from typing import Callable, List, Optional, Union, get_args
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by PierreGuilmin** [comment]
@abocquet I set up a little the scene for the time-dependence by introducing the appropriate type, but all `mesolve` is written for time-independent `H` now. I believe it won't be so easy to add the time-dependency, but once done we're all set 🙌🏼 For the `TimeDependentOperatorLike` type introduced in this PR, I was thinking to have specific `to_tensor` methods like:
```python
def time_dependent_to_tensor(x: TimeDependentOperatorLike) -> TimeDependentOperator:
    if isinstance(x, get_args(OperatorLike)):
        return to_tensor(x)
    return lambda t: to_tensor(x(t))
```
(Btw, warning bc I discovered that `QObj` is a callable so all tests `callable(x)` are sometimes returning `True` unexpectedly)

torchqdynamics/types.py
```diff
@@ -12,25 +12,37 @@
...
+           of these types. If `None` or empty list, returns an empty tensor of size (0).
 
     Returns:
         PyTorch tensor.
     """
+    if x is None:
+        return torch.tensor([])
+    if isinstance(x, list):
+        if len(x) == 0:
+            return torch.tensor([])
```
gautierronan, line 37: I believe this is redundent with the line below, since `torch.stack([])` should return an empty tensor.

gautierronan (changes requested):
So, I'm not quite sure what to think of this PR. Originally I had made a mean review saying that I thought it was not quite the right time to interface so hard with qutip. But maybe it's actually fine, and I quite like it in the end.

torchqdynamics/types.py
PierreGuilmin: Unfortunately:
```python
RuntimeError: stack expects a non-empty TensorList
```

PierreGuilmin:
> So, I'm not quite sure what to think of this PR. Originally I had made a mean review saying that I thought it was not quite the right time to interface so hard with qutip. But maybe it's actually fine, and I quite like it in the end.

I must confess I got tired of writing `torch.stack` and `tq.to_tensor` everywhere in my tests notebooks 😅  I agree it's a bit early, but this really ease writing examples and test cases so 🤷‍♂️  In the end we should _not rely_ on QuTiP even for the testing, but we need to write a few convenience functions that we don't have yet.

abocquet (approved):

gautierronan (approved):

PR #32, opened 2023-03-23
Test `mesolve` batching

PierreGuilmin:
Built on top of #31.

This is not **at all** a complete test suite for `mesolve`, it's just to test that the batching works and that the Rouchon 1 method runs with no bug (which does not mean that the result is correct 😋). I also took the chance to have isort and yapf run on the tests directory.

### DIFF
--- a/tests/test_mesolve.py
+++ b/tests/test_mesolve.py
@@ -16,7 +16,7 @@ def test_mesolve_batching():
     a = qt.destroy(n)
     adag = a.dag()
     H = delta * adag * a
-    H_batched = [0.5 * delta * adag * a, delta * adag * a, 2 * delta * adag * a]
+    H_batched = [0.5 * H, H, 2 * H]
     b_H = len(H_batched)
     jump_ops = [np.sqrt(kappa) * a, np.sqrt(kappa) * a]
     exp_ops = [(a + adag) / np.sqrt(2), (a - adag) / (np.sqrt(2) * 1j)]
@@ -32,7 +32,7 @@ def test_mesolve_batching():
     ]
     b_rho0 = len(rho0_batched)
     num_t_save = 51
-    t_save = np.linspace(0.0, delta / (2 * np.pi), num_t_save)  # a full tour
+    t_save = np.linspace(0.0, delta / (2 * np.pi), num_t_save)  # a full rotation
     solver = tq.solver.Rouchon(dt=1e-4, order=1)
 
     run_mesolve = lambda H, rho0: tq.mesolve(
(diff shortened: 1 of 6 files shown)


abocquet (dismissed):

tests/test_mesolve.py
```diff
@@ -0,0 +1,61 @@
...
+    rho0 = qt.coherent(n, alpha0)
+    rho0_batched = [
+        qt.coherent(n, alpha0),
+        qt.coherent(n, 1j * alpha0),
+        qt.coherent(n, -alpha0),
+        qt.coherent(n, -1j * alpha0)
+    ]
+    b_rho0 = len(rho0_batched)
+    num_t_save = 51
+    t_save = np.linspace(0.0, delta / (2 * np.pi), num_t_save)  # a full tour
```
gautierronan, line 35: I think you mean `# a full revolution` or `a full rotation` :)

tests/test_mesolve.py
```diff
@@ -0,0 +1,61 @@
...
+    num_exp_ops = len(exp_ops)
+
+    # other arguments
+    rho0 = qt.coherent(n, alpha0)
+    rho0_batched = [
+        qt.coherent(n, alpha0),
+        qt.coherent(n, 1j * alpha0),
+        qt.coherent(n, -alpha0),
+        qt.coherent(n, -1j * alpha0)
+    ]
```
gautierronan, line 32: It would probably be better to not use qutip objects in this test, and to write another test that checks if qutip is supported by `mesolve`. But since we're lacking proper object definitions, this is probably fine for now.

tests/test_mesolve.py
```diff
@@ -0,0 +1,61 @@
...
+    n = 8
+    kappa = 1.0
+    delta = 2 * np.pi
+    alpha0 = 1.0
+
+    # operators
+    a = qt.destroy(n)
+    adag = a.dag()
+    H = delta * adag * a
+    H_batched = [0.5 * delta * adag * a, delta * adag * a, 2 * delta * adag * a]
```
gautierronan, line 19: Better to write `H_batched = [0.5 * H, H, 2 * H]` like you did below.

tests/test_mesolve.py
```diff
@@ -0,0 +1,61 @@
...
+    assert states.shape == (b_H, num_t_save, n, n)
+    assert exp.shape == (b_H, num_exp_ops, num_t_save)
+
+    # batched rho0
+    states, exp = run_mesolve(H, rho0_batched)
+    assert states.shape == (b_rho0, num_t_save, n, n)
+    assert exp.shape == (b_rho0, num_exp_ops, num_t_save)
+
+    # batched H and rho0
+    H_batched = [0.5 * H, H, 2 * H]
```
gautierronan, line 58: Already defined above.

gautierronan (dismissed):

tests/test_mesolve.py
PierreGuilmin: I think the same 😊

PierreGuilmin:
I think I have a nice idea to streamline the writing of our "physically-inspired" tests, I will introduce it after the Euler method PRs.

tests/test_mesolve.py
PierreGuilmin: It will just be a matter of replacing every `qt` by a `tq` 😋

PR #33, opened 2023-03-24
Move `mesolve` to a specific folder

PierreGuilmin:
Built on top of #32.

Replace #16. ~With the proposed code there is still a (future) name clash between the `Rouchon1` of `mesolve` and the `Rouchon1` solver of `smesolve`.~ I fixed this, let's discuss how the user picks the solvers next meeting (with this PR it's `tq.mesolver`).

I remove the `order` parameter of `Rouchon` because it felt one could choose a continuum, while it was really just a way to pick btw 1, 1.5 and 2. Another option would have been to set it as a `Literal`, but I think this is the best.


tests/mesolve/test_mesolve.py
```diff
@@ -33,7 +33,7 @@ def test_mesolve_batching():
     b_rho0 = len(rho0_batched)
     num_t_save = 51
     t_save = np.linspace(0.0, delta / (2 * np.pi), num_t_save)  # a full tour
-    solver = tq.solver.Rouchon(dt=1e-4, order=1)
+    solver = tq.mesolver.Rouchon1(dt=1e-4)
```
gautierronan, line 36: I don't like that we should call `tq.mesolver` instead of `tq.solver`. The solver options should be common to all functions, and then within the actual function it is dispatched to the appropriate QSolver. It's much more simple on the user-end.

torchqdynamics/mesolve/solver_options.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@dataclass
+class Rouchon1_5(FixedStep):
+    pass
+
+
+@dataclass
+class Rouchon2(FixedStep):
+    pass
```
gautierronan, line 18: As I was suggesting, I think it would be nice to have a default `Rouchon` solver that redirects to either `Rouchon1` or `Rouchon2` by default. This way users can simply call `Rouchon` without having to bother about the backend order of the solver.

gautierronan (changes requested):

tests/mesolve/test_mesolve.py
PierreGuilmin: Agreed 👍 The correct solution is to have a common namespace for solver options coming from different places, from what I know this can only be done by creating a new file importing all the guys from the different places, and then exposing the file namespace -> see the proposed changes.

PierreGuilmin: It was also the chance to rename `solver` to the more accurate `solver_options`, while keeping the simple `tq.solver` API.

gautierronan (approved):

PR #34, opened 2023-03-24
Rename `ssolve` to `sesolve` and add basic API

PierreGuilmin:
Built on top of #33.

Quite some copied code for the `sesolve` API, I think we can factor that later after having `smesolve` to do it cleanly.

Ho and I renamed to `sesolve` because we have Master Equation => `mesolve`, so I thought Schrödinger Equation => `sesolve`.


abocquet (dismissed):

torchqdynamics/sesolve/sesolve.py
```diff
@@ -0,0 +1,66 @@
...
+    # TODO: placeholder, to remove
+    if solver is None:
+        pass
+
+    # TODO: placeholder, to remove
+    qsolver = None
+
+    # compute the result
+    b_H = H_batched.size(0)
+    y0 = psi_batched[None, ...].repeat(b_H, 1, 1, 1)  # (b_H, b_psi, n, n)
```
gautierronan, line 55: Like I did in my PR for `mesolve`, these two lines should be above under the `# convert psi to a tensor and batch by default` comment. Also `y0` should be renamed to `psi0_batched`.

torchqdynamics/sesolve/sesolve.py
```diff
@@ -0,0 +1,66 @@
...
+import torch.nn as nn
+
+from ..odeint import odeint
+from ..solver import SolverOption
+from ..types import OperatorLike, TensorLike, TimeDependentOperatorLike, to_tensor
+
+
+def sesolve(
+    H: TimeDependentOperatorLike,
+    psi: OperatorLike,
```
gautierronan, line 13: `psi0` to be consistent with `mesolve`.

torchqdynamics/sesolve/sesolve.py
```diff
@@ -0,0 +1,66 @@
...
+    t_save: TensorLike,
+    *,
+    exp_ops: Optional[List[OperatorLike]] = None,
+    save_states: bool = True,
+    gradient_alg: Literal[None, 'autograd', 'adjoint'] = None,
+    parameters: Optional[Tuple[nn.Parameter, ...]] = None,
+    solver: Optional[SolverOption] = None,
```
gautierronan, line 20: IMO a more logical ordering of keyword arguments is:
 - save_states
 - exp_ops
 - solver
 - gradient_alg
 - parameters

(of course it's just for readability since we enforce non-positional arguments with the *)

torchqdynamics/sesolve/sesolve.py
```diff
@@ -0,0 +1,66 @@
...
+    #     psi: (b_psi?, n, 1)
+    #
+    # Returns:
+    #     (y_save, exp_save) with
+    #     - y_save: (b_H?, b_psi?, len(t_save), n, 1)
+    #     - exp_save: (b_H?, b_psi?, len(exp_ops), len(t_save))
+
+    # TODO: support density matrices too
+    # TODO: H is assumed to be time-independent from here (temporary)
+
```
gautierronan, line 33: I think we should add a test on the shape of `psi0` at the beginning of the function. Users might be tempted to give initial kets of shape (n,) or (b,n). Both of these should raise an error.

torchqdynamics/sesolve/sesolve.py
```diff
@@ -0,0 +1,66 @@
...
+
+def sesolve(
+    H: TimeDependentOperatorLike,
+    psi: OperatorLike,
+    t_save: TensorLike,
+    *,
+    exp_ops: Optional[List[OperatorLike]] = None,
+    save_states: bool = True,
+    gradient_alg: Literal[None, 'autograd', 'adjoint'] = None,
```
gautierronan, line 18: I prefer `Optional[Literal['autograd', 'adjoint']]` here, to make clear that the sensitivity algorithm is optional.

gautierronan (changes requested):
A few small comments, but thanks for the PR :)

torchqdynamics/sesolve/sesolve.py
PierreGuilmin: I added a TODO 👍🏼  I prefer to merge this quickly without waiting for a review of the small changes because we have quite a lot of stacked PRs waiting. Also I think we should discuss about arguments validation, namely what are the checks we want to implement on all arguments (the one you proposed is definitely necessary 👍🏼)

PR #35, opened 2023-03-24
Time-dependent Hamiltonian | Part 1

abocquet:
Adds support for time dependent Hamiltonians. The main focus is to generally ease the addition of variables that depend on other variables, Hamiltonian or other

This PR is a bit long in number of lines but:
- lines > 100 in `decorators.py` is taken from functools (https://github.com/PierreGuilmin/torchqdynamics/pull/35/files#diff-768f0f6962f303a6e209998ce18c325558976db31d86d4326c61447a18cb811cR104)
- The test file is very verbose but simple to read. I can add it in a next PR if necessary.

### DIFF
--- a/torchqdynamics/decorators.py
+++ b/torchqdynamics/decorators.py
@@ -1,0 +1,153 @@
+"""
+Code freely inspired from functools.cached_property
+"""
+import inspect
+from copy import copy
+from typing import Dict
+
+CACHE_KEY = '__cache__'
+CACHED_VALUE_KEY = 'cached_value'
+PREVIOUS_KEY = 'previous_key'
+HAS_CHANGED_KEY = 'has_changed'
+
+
+def cached_main(func):
+    """
+    Decorator that implements a size 1 LRU and propagates changes to children,
+    defined using `cached_depends_on`. The function must return a boolean to indicate whether the
+    value changed or not since last call
+    """
+    func_name = func.__name__
+
+    def inner(instance, *args, **kwargs):
+        instance_dict = instance.__dict__
+        if CACHE_KEY not in instance_dict:
+            instance_dict[CACHE_KEY] = {}
+
+        cache_dict = instance_dict[CACHE_KEY].setdefault(func_name, {})
+
+        key = _make_key(args, kwargs, False)
+        has_changed = False
+
+        if cache_dict.get(CACHED_VALUE_KEY
+                         ) is None or cache_dict.get(PREVIOUS_KEY) != key:
+            val = func(instance, *args, **kwargs)
+            assert isinstance(
+                val[-1], bool
+            ), 'The last return param of a cached_main function must be a boolean indicating whether the return value ' \
+               'should change'
+            has_changed = val[-1] or cache_dict.get(CACHED_VALUE_KEY) is None
+            cache_dict[PREVIOUS_KEY] = key
+            if has_changed:
+                cache_dict[CACHED_VALUE_KEY] = val[0:-1]
+
+        cache_dict[HAS_CHANGED_KEY] = has_changed
+        return cache_dict[CACHED_VALUE_KEY]
+
+    return inner
+
+
+def cached_depends_on(dependencies: Dict[str, str]):
+    """
+    Decorator that implements a size 1 LRU on the function. All the parameters of the functions
+    are hashed as the LRU key **except** for the dependencies.
+    We call 'source function' a function that computes a dependence.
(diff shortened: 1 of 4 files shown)


PierreGuilmin:
Can you motivate why we bring such machinery? I understand you want to propose a very general way to handle dependent-variables, but this is only needed in our case for `H` so I don't know if it justifies some dark magic. The `decorators.py` code will be especially hard to review/maintain, and I'm not sure why we need it. For example, what was not working with the previous much simpler implementation?

abocquet:
It's a matter of separation of concerns: we move the complexity of caching away from the logic of the solvers.
Once it's there adding caching is basically free so we avoid boilerplate and potential mistakes. 

I do agree that this code is not trivial but it's tested and should not move much in the future so it's fine to have that IMHO. If you insist we can going back but it would be a shame to, as now we have a tool that works well.

gautierronan:
> It's a matter of separation of concerns: we move the complexity of caching away from the logic of the solvers. Once it's there adding caching is basically free so we avoid boilerplate and potential mistakes.
> 
> I do agree that this code is not trivial but it's tested and should not move much in the future so it's fine to have that IMHO. If you insist we can going back but it would be a shame to, as now we have a tool that works well.

I agree it's nice to separate this kind of machinery from the main code, but I have mainly two concerns about your solution:

1. Readability. It's difficult to understand what's happening, especially for the poor physicist who just wants to know what's happening to it's time-dependent matrix under the hood. Readability is a big factor in people adopting the code. I know of a lot of theoreticians that do not use qutip because they don't know what's happening and they don't want to spend time going deep into the code to understand. Instead they use home-made solvers.
2. Performance. I'm concerned that this additional machinery may not bode well with pytorch, and especially with `torch.compile()`. This is in a very performance-sensitive part of the code, so I would argue that simpler is better. But it might not change anything, I'm not sure.

torchqdynamics/decorators.py
PierreGuilmin, line 1: So I won't review this file, I trust you 😉

torchqdynamics/mesolve.py
```diff
@@ -54,13 +55,18 @@ def forward(self, t, dt, rho):
         H_nh = self.H(t) - 0.5j * self.sum_nojump
```
PierreGuilmin, line 55: This should be calculated once for non time-dependent tensors.

PierreGuilmin (changes requested):
Here's a first review. Let's rebase this on `main` once we agree on the code, to take `H` batching and the file reorganization into account (it may be easier to open a new PR and manually copy-pasting the modifications). We also need to do the order Rouchon orders (but maybe it's coming next).

On another note, I understand the interest of a completely isolated little helper to cache functions depending on some variables, but let me advocate the following alternative option. The reason is that I think `H` is the only time-dependent object we will have, and that the caching is mostly going to be used for Rouchon methods, so I'm not sure if we need such a generic solution. Moreover, even if the option I propose is a tiny bit more manual, it really adds very little code and does not, IMO, affect features or readability.
<details>
  <summary>Alternative option (drafty)</summary>

  Add to `ForwardQSolver`:

  ```python
  def __init__(self, H):
      self._H = H

  def H(self, t):
      if isinstance(self._H, torch.Tensor):
          self._H_has_changed = False
          return self._H
      elif callable(self._H):
          self._H_has_changed = True
          return self._H(t)
      else:
          # will come in a next PR
          raise NotImplementedError('Piecewise constant Hamiltonian is not supported yet.')
  ```

  Add to `MERouchon`

  ```python
  def forward(self, t, dt, rho):
      """Compute rho(t+dt) using a Rouchon method of order 1."""
      rho = kraus_map(rho, self.Ms(t, dt))
      return rho / trace(rho)[..., None, None].real

  def Ms(self, t, dt):
      if not self._H_has_changed:
          return self._Ms

      # non-hermitian Hamiltonian at time t
      H_nh = self.H(t) - 0.5j * self.sum_nojump

      M0 = self.I - 1j * dt * H_nh
      self._Ms = torch.cat((M0[None, ...], sqrt(dt) * self.jump_ops))
      return self._Ms
  ```

</details>

abocquet:
I agree your solution is simple enough so it's scalable. However for `dt`s that depend on time, you need to update `M0(H, dt)` on change of `H` or `dt`. It's still doable locally but it adds more potential bugs. We could also decide that for time-variable schemas we always recompute `M0`

PierreGuilmin:
That's a good point. I don't know whether adaptative solvers change their `dt` at each step or not. 🤷‍♂️

PR #36, opened 2023-03-24
Euler method for `sesolve`

PierreGuilmin:
Built on top of #34.

Useful to test the API, for benchmark, to show-off how easy it is to add a solver, for documentation, ~and to fry some bacon~.

Another PR will follow for `mesolve` Euler dumb method.


torchqdynamics/sesolve/euler.py
```diff
@@ -0,0 +1,25 @@
+import torch
+
+from ..odeint import ForwardQSolver
+from ..solver import SolverOption
+from ..types import TimeDependentOperator
+
+
+class SEEuler(ForwardQSolver):
```
PierreGuilmin, line 8: Actually it's maybe an `AdjointQSolver`, but let's not care about implementing the adjoint on this guy for now, except if you find it useful for some reason @gautierronan (maybe as a playground example?).

abocquet (approved):

torchqdynamics/sesolve/euler.py
gautierronan: So I did not think about it too much tbh. It can definitly be implemented for `sesolve`, the question is more: is it useful?

The adjoint method is an answer to the large memory requirement of regular backward autodiff for `mesolve`. For `sesolve`, ket takes much less memory, but the Hamiltonian is also stored in the graph of operations I believe. In this case, one could still run quite quickly in memory problems for time-dependent problems. So I guess it could also be useful, and it's quite easy to implement.

torchqdynamics/sesolve/euler.py
```diff
@@ -0,0 +1,25 @@
...
+        self.options = solver_options
+
+    def forward(self, t: float, dt: float, psi: torch.Tensor):
+        # Args:
+        #     psi: (b_H, b_psi, n, 1)
+        #
+        # Returns:
+        #     (b_H, b_psi, n, 1)
+
+        return psi - 1j * self.H @ psi * dt
```
gautierronan, line 25: Can you put the `dt` in front of the matmul? I find it more normal ordering.

gautierronan (dismissed):
All gucci.

torchqdynamics/sesolve/euler.py
PierreGuilmin: I agree, at least it's a good and simple example to explain how the adjoint works.

PR #39, opened 2023-03-25
Implement adjoint autodiff for `mesolve`

gautierronan:
I suggest reviewing commit by commit. 

I am not happy about everything, but the main idea is here. In particular, I'm not sure about commit c04e4b289d23d7e3ad56c715730f441ff7c7f500 and the `QSolver` organization (with the flag inside `ODEIntAdjoint.backward`). I'm down for suggestions.

Also I have not tested it, so a few bugs here and there might remain. But I've got a similar implementation for my own project that works. I will write a test function at some point.

---

EDIT: I have refactored the code in a better way. I'm now much more happy with the implementation. Instead of the backward pass going back into `odeint_main`, it now has a dedicated backward ode integrator. It helps a lot readability, but also avoids overloading the various `odeint` functions.

When calling `odeint(..., gradient_alg='adjoint',...)`, the following calls are sequentially made:
 - **Forward pass**: `odeint` --> `_odeint_adjoint` --> `ODEIntAdjoint.forward` --> `_odeint_inplace` --> `_odeint_main` --> `_fixed_odeint`/`_adaptive_odeint` --> `qsolver.forward`
 - **Backward pass**: `odeint` --> `_odeint_adjoint` --> `ODEIntAdjoint.backward` --> `_odeint_augmented_main` --> `_fixed_odeint_augmented`/`_adaptive_odeint_augmented` --> `qsolver.backward_augmented`

For reviewing, it might now be better to look at all commits at once, and to start from `odeint.py` before reading `mesolve.py`.

### DIFF
--- a/torchqdynamics/mesolve.py
+++ b/torchqdynamics/mesolve.py
@@ -7,7 +7,7 @@ from torch import Tensor
 
 from .odeint import AdjointQSolver, odeint
 from .solver import Rouchon, SolverOption
-from .solver_utils import inv_sqrtm, kraus_map
+from .solver_utils import inv_sqrtm, kraus_map, none_to_zeros_like
 from .types import TDOperator, TensorLike
 from .utils import trace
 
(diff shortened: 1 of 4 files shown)


PierreGuilmin:
I will review this along #41 tomorrow @gautierronan.

PR #44, opened 2023-03-26
Euler method for `mesolve`

PierreGuilmin:


torchqdynamics/solver_utils.py
```diff
@@ -45,3 +45,15 @@ def bexpect(operators: Tensor, state: Tensor) -> Tensor:
     # TODO: Once QTensor is implemented, check if state is a density matrix or ket.
     # For now, we assume it is a density matrix.
     return torch.einsum('bij,...ji->...b', operators, state)
+
+
+def dissipator(rho: torch.Tensor, O: torch.Tensor) -> torch.Tensor:
+    return (
+        torch.einsum('abij,a...jk,abkl->a...il', O, rho, O.adjoint()) -
+        0.5 * torch.einsum('abij,abjk,a...kl->a...il', O.adjoint(), O, rho) -
+        0.5 * torch.einsum('a...ij,abjk,abkl->a...il', rho, O.adjoint(), O)
```
gautierronan, line 54: A faster way to do this is to:
 - Compute `tmp = Ldag @ L @ rho`
 - Sum with its adjoint `tmp = tmp + tmp.adjoint()`
 - Add the jump term `rho = L @ rho @ Ldag - 0.5 * tmp`

All of this using einsum as much as possible of course. This runs only 4 matmuls, whereas your method runs 6. Also you can combine the 2nd and 3rd step in one line.

torchqdynamics/solver_utils.py
```diff
@@ -45,3 +45,15 @@ def bexpect(operators: Tensor, state: Tensor) -> Tensor:
...
+    )
+
+
+def lindbladian(rho: torch.Tensor, H: torch.Tensor, O: torch.Tensor) -> torch.Tensor:
+    return -1j * (H @ rho - rho @ H) + dissipator(rho, O)
```
gautierronan, line 59: Same here. Faster way is to compute `H @ rho` and then sum with its adjoint. H and rho have to be hermitian however.

Actually, it would be even better to compute the full non-hermitian Hamiltonian (including the sum of Ldag @ L), and do the same trick as above.

tests/mesolve/test_euler.py
```diff
@@ -0,0 +1,32 @@
...
+    n = 8
+    kappa = 1.0
+    delta = 2 * np.pi
+    alpha0 = 1.0
+
+    # operators
+    a = qt.destroy(n)
+    adag = a.dag()
+    H = delta * adag * a
+    jump_ops = [np.sqrt(kappa) * a, np.sqrt(kappa) * a]
```
gautierronan, line 19: Maybe better to use two different jump operators? You could use the identity for the second jump operator actually, just for sanity check that it doesn't do anything. And this way you still check the `jump_ops` batching.

tests/mesolve/test_euler.py
```diff
@@ -0,0 +1,32 @@
...
+    # other arguments
+    rho0 = qt.coherent(n, alpha0)
+    num_t_save = 51
+    t_save = np.linspace(0.0, delta / (2 * np.pi), num_t_save)  # a full tour
+    solver = tq.solver.Euler(dt=1e-3)
+
+    # run solver
+    states, exp = tq.mesolve(H, jump_ops, rho0, t_save, exp_ops=exp_ops, solver=solver)
+    assert states.shape == (num_t_save, n, n)
+    assert exp.shape == (num_exp_ops, num_t_save)
```
gautierronan, line 32: Since we're not checking the result, this function should be renamed `test_mesolve_euler_batching` or something. Also, if we have a lot of different test functions that all test for batching, it could be nice to group them into a single one, and to use `@pytest.parametrize` for all different parameters/solvers.

gautierronan (changes requested):

tests/mesolve/test_euler.py
PierreGuilmin: I'll rename it with `_cheap` because it's not even testing the full batching, it will change soon anyway. 😋

torchqdynamics/solver_utils.py
```diff
@@ -45,3 +45,14 @@ def bexpect(operators: Tensor, state: Tensor) -> Tensor:
     # TODO: Once QTensor is implemented, check if state is a density matrix or ket.
     # For now, we assume it is a density matrix.
     return torch.einsum('bij,...ji->...b', operators, state)
+
+
+def lindbladian(rho: torch.Tensor, H: torch.Tensor, L: torch.Tensor) -> torch.Tensor:
```
gautierronan, line 50: For consistency with the rest, maybe we could type with `Tensor` instead of `torch.Tensor`. Unless you prefer the other one, I'm fine with both.

torchqdynamics/solver_utils.py
```diff
@@ -45,3 +45,14 @@ def bexpect(operators: Tensor, state: Tensor) -> Tensor:
...
+    sum_nojump = (L.adjoint() @ L).sum(dim=0)
+
+    # non-hermitian Hamiltonian
+    H_nh = H - 0.5j * sum_nojump
+
+    H_nh_rho = H_nh @ rho
```
gautierronan, line 56: Is there any performance gain by putting all of these in a single line? Since this will be a performance-sensitive part of the code (called at every time step for adaptive solvers), maybe we could use `torch.einsum` as much as possible, and group as much as we can in single calls.

gautierronan (changes requested):
Sorry for the second review, but there's especially the `lindbladian` function which I think is import to define well for performance.

torchqdynamics/solver_utils.py
PierreGuilmin: You're 100% right. (never be sorry for reviewing twice, it's better! 😋)

And **_arg_**, even for simple matmul it feels like we are going to need `einsum` everywhere 😭 
```python
import torch
import torch.utils.benchmark as benchmark

n = 50
b_H = 5
b_rho = 4
H = torch.rand(b_H, n, n, dtype=torch.complex64)
rho = torch.rand(b_H, b_rho, n, n, dtype=torch.complex64)
Hb = H[:, None, ...]  # (b_H, 1, n, n)

stmt1 = "Hb @ rho"
stmt2 = "torch.einsum('aij,a...jk->a...ik', H, rho)"

t1 = benchmark.Timer(stmt=stmt1, globals={'Hb': Hb, 'rho': rho})  # 402.55 us
t2 = benchmark.Timer(stmt=stmt2, globals={'H': H, 'rho': rho})  # 238.43 us

print(t1.timeit(100))
print(t2.timeit(100))
```

So I think we loose every time we augment a tensor to have a specific batching (e.g. everytime we do smthg like `self.H = H[:, None, ...]` and then use `H` in a matrix multiplication).

I propose to merge this PR, and then I'll do a new PR to einsum everything correctly. And then we absolutely need to look at https://optimized-einsum.readthedocs.io/en/stable/.

gautierronan: On another note, I tried the pytorch sparse COO the other day. It gives some crazy GPU speedup on large matrices (x50 for N=1000), but supports very little actual useful operations. So I'm really convinced that a proper DIA matrix implementation will give us a crazy speedup.

gautierronan (approved):

PR #47, opened 2023-03-26
Clean up code and minimal documentation for odeint/mesolve

gautierronan:
Copy of https://github.com/PierreGuilmin/torchqdynamics/pull/41 after folder reorganization.

---

A little bit of clean up and documentation.

Also, I removed the argument `dt` accross various `_odeint` functions, since the time step should be private to the qsolver (the fixed odeint subroutine never changes the value of `dt` called). It also makes the `qsolver.forward` functions signature consistent between fixed and adaptive solvers (see next PR).


torchqdynamics/odeint.py
```diff
@@ -61,44 +90,60 @@ def odeint(
...
-        return _fixed_odeint(qsolver, y0, t_save, dt, exp_ops, save_states)
+        return _fixed_odeint(qsolver, y0, t_save, save_states, exp_ops)
     elif isinstance(qsolver.options, AdaptativeStep):
-        return _adaptive_odeint(qsolver, y0, t_save, exp_ops, save_states)
+        return _adaptive_odeint(qsolver, y0, t_save, save_states, exp_ops)
 
 
-# For now we use *args and **kwargs for helper methods that are not implemented to ease
-# the potential api changes that could occur later. When a method is implemented the
-# methods should take the same arguments as all others.
```
PierreGuilmin, line 76: We should keep this.

PierreGuilmin (changes requested):
A few comments that apply everywhere:
- Please use backticks everywhere you specify the shape of a tensor.
- You don't need to repeat the typing in the documentation, it will be automatically set by mkdocs later by reading the function signature. The only case where you'd want to add it explicitly using parentheses is if you want to simplify the typing with words (e.g. say `H (tensor or function)`).
- I suggest to use a `self.dt = self.options.dt` in `ForwardQSolver` to (a) fix a design issue (odeint does not know where to get the `dt` from a solver) and (b) remove from the solvers implementations the `dt = qsolver.options.dt` and just use `self.dt`.

PierreGuilmin:
@gautierronan can you merge/rebase on main?

PR #49, opened 2023-03-27
Adaptive time step solvers

gautierronan:
This PR adds the framework for adaptive time step solvers, and a single solver for now (Dormand-Prince of order 5). It is implemented for both `mesolve` and `sesolve`.

The function `lindbladian` is not implemented, but will be in https://github.com/PierreGuilmin/torchqdynamics/pull/44.

To integrate the ODE, I have defined a new class `AdaptiveSolver` which takes care of all the "non-quantum" stuff about adaptive time step integration. It is common to all quantum solvers, `mesolve, `sesolve` and others. It can be sub-classed with as many integrators as we would like.

NB: I have tested nothing so a test function and some debugging will be required at some point.

NB2: The PR includes some formatting that is unrelated to the PR, caused by the change from yapf to black. It will be fixed once we merge with the parent branch, so don't mind this.


torchqdynamics/adaptive.py
```diff
@@ -0,0 +1,166 @@
...
+        return (alpha, beta, csol, csol - cerr)
+
+    def step(self, f0: Tensor, y0: Tensor, t0: float,
+             dt: float) -> Tuple[Tensor, Tensor, Tensor]:
+        """Compute a single step of the ODE integration."""
+        # import butcher tableau
+        alpha, beta, csol, cerr = self.tableau
+
+        # compute iterated Runge-Kutta values
+        k = torch.zeros(7, *f0.shape, dtype=f0.dtype, device=f0.device)
```
abocquet, line 155: Shouldn't we do `torch.tensor(alpha).to(t0)` instead? It would be more straightfoward than storing the types aside + currently it is possible to have actually a different type and device for `alpha` and `t0` 

Same comment for `y` and `beta` etc...

abocquet (changes requested):
Great work, I don't understand all of the numerics specifics but the code looks good, except for one point in comment

torchqdynamics/adaptive.py
gautierronan: Thanks for raising this. Actually I have two partial answers to this:

1. It's theoretically faster/cleaner to initialize the tensor directly by specifying the dtype and device, than to initialize it and then transfer it to the proper dtype/device using `.to(...)`. It avoids GPU/CPU transfers.
2. I think there's no reason for the `AdaptiveSolver` class to know about `t0` or `y0`. It should not be an attribute of the class, so I don't really want to pass it as argument.

However my first point is quite of whatever because this is a not a performance-sensitive part of the code. The only the second point is left. Do you think I should pass t0 as argument?

abocquet: Both arguments are good arguments. To take them into account I propose `build_tableau(self)` becomes `build_tableau(self, target)`, we define the variables as `torch.tensor(alpha, dtype=target.dtype, device=target.device)` and then in `step`:
```
if self.tableau is None:
    self.tableau = self.build_tableau(f0)
```

What do you think?

gautierronan: Sounds good to me !

abocquet (dismissed):

torchqdynamics/adaptive.py
```diff
@@ -0,0 +1,177 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+
+import torch
```
PierreGuilmin, line 4: Missing `from torch import Tensor`.

PierreGuilmin (changes requested):
This library is moving so fast 🏃‍♂️ 

Could you merge/rebase on main?

I have not fully read `_adaptive_odeint()` and `adaptive.py` yet.

torchqdynamics/solver_options.py
```diff
@@ -24,10 +27,5 @@ class Euler(FixedStep):
 
 
 @dataclass
-class RK4(FixedStep):
-    pass
-
-
-@dataclass
-class DOPRI6(AdaptativeStep):
+class Dopri45(AdaptiveStep):
```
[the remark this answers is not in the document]
gautierronan: Well scipy uses `dopri45` with lower case letters. I thought I would use `Dopri45` to have consistent letter casing with `Rouchon` and `Euler` methods.

torchqdynamics/odeint.py
```diff
@@ -121,9 +122,109 @@ def _odeint_inplace(*args, **kwargs):
...
+
+        # perform a single solver step of size dt
+        ft_new, y_new, y_err = solver.step(ft, y, t, dt)
+
+        # compute estimated error of this step
+        error = solver.get_error(y_err, y, y_new)
+        accept_step = error <= 1
+
+        # update results if step is accepted
+        if accept_step:
```
PierreGuilmin, line 200:
```suggestion
        # update results if step is accepted
        if error <= 1:
```

torchqdynamics/odeint.py
```diff
@@ -121,9 +122,109 @@ def _odeint_inplace(*args, **kwargs):
...
+
+    if len(exp_ops) > 0:
+        exp_save = torch.zeros(*batch_sizes, len(exp_ops), len(t_save)).to(y0)
+
+    # save initial solution
+    save_counter = 0
+    if t_save[0] == 0.0:
+        if save_states:
+            y_save[..., save_counter, :, :] = y0
+
```
PierreGuilmin, line 155:
```suggestion
```

torchqdynamics/odeint.py
```diff
@@ -121,9 +122,109 @@ def _odeint_inplace(*args, **kwargs):
...
+    batch_sizes, (m, n) = y0.shape[:-2], y0.shape[-2:]
+    if save_states:
+        y_save = torch.zeros(*batch_sizes, len(t_save), m, n).to(y0)
+
+    if len(exp_ops) > 0:
+        exp_save = torch.zeros(*batch_sizes, len(exp_ops), len(t_save)).to(y0)
+
+    # save initial solution
+    save_counter = 0
+    if t_save[0] == 0.0:
```
PierreGuilmin, line 152: I think we need to update `save_counter` in this if branch.

torchqdynamics/adaptive.py
```diff
@@ -0,0 +1,182 @@
...
+        real_dtype = to_float(dtype)
+        device = target.device
+
+        # initialize tensors
+        alpha = torch.tensor(alpha, dtype=real_dtype, device=device)
+        beta = torch.tensor(beta, dtype=dtype, device=device)
+        csol = torch.tensor(csol, dtype=dtype, device=device)
+        cerr = torch.tensor(cerr, dtype=dtype, device=device)
+
+        return (alpha, beta, csol, csol - cerr)
```
PierreGuilmin, line 157:
```suggestion
        return alpha, beta, csol, csol - cerr
```

torchqdynamics/adaptive.py
```diff
@@ -0,0 +1,182 @@
...
+    def step(
+        self, f0: Tensor, y0: Tensor, t0: float, dt: float
+    ) -> Tuple[Tensor, Tensor, Tensor]:
+        """Compute a single step of the ODE integration."""
+        # create butcher tableau if not already done
+        if self.tableau is None:
+            self.tableau = self.build_tableau(f0)
+
+        # import butcher tableau
+        alpha, beta, csol, cerr = self.tableau
```
PierreGuilmin, line 168: It's a bit weird that `csol-cerr` turned into `cerr` here.

PierreGuilmin (changes requested):
This is beautifully written @gautierronan 😍 Very minor comments overall, I didn't check the numerics but rather the overall logic and implementation, it looks good.

torchqdynamics/adaptive.py
gautierronan: The point is Dormand-Prince computes 4th order and 5th order solution estimations using the same runge kutta points, and then the difference is an estimate of the error. But the 4th order is only used to compute the error. So anyway, I renamed the above variables for clarity.

PierreGuilmin (approved):
Let's go 🔥

PR #50, opened 2023-03-28
Fix `bexpect` for ket

PierreGuilmin:
Fix expectation values computation for `sesolve`.

New proposed conventions (I write them all down in a doc that I'll share, so we have a complete and unambiguous set of rule):
- Use _size_ (instead of _shape_ or _dimension_) when referring to a tensor size.
- Use _state vector_ and _density matrix_ to specify ket $\psi$ vs density matrix $\rho$.
- Don't specify what's explicitly batchable, the tensor size speaks for itself.

Once we're satisfied with the set of conventions we can go through the code and fix inconsistencies everywhere.

### DIFF
--- a/torchqdynamics/solver_utils.py
+++ b/torchqdynamics/solver_utils.py
@@ -37,8 +37,8 @@ def inv_sqrtm(mat: Tensor) -> Tensor:
 
 
 def bexpect(O: Tensor, x: Tensor) -> Tensor:
-    """Compute the expectation values of many operators on a state vector or a density
-    matrix.
+    """Compute the expectation values of batched operators on a state vector or a
+    density matrix.
 
     The expectation value $\braket{O}$ of a single operator $O$ is computed
     - as $\braket{O}=\braket{\psi|O|\psi}$ if `x` is a state vector $\psi$,
--- a/torchqdynamics/solver_utils.py
+++ b/torchqdynamics/solver_utils.py
@@ -3,6 +3,8 @@ from typing import Tuple
 import torch
 from torch import Tensor
 
+from .utils import is_ket
+
 
 def kraus_map(rho: Tensor, O: Tensor) -> Tensor:
     """Compute the application of a Kraus map on an input density matrix.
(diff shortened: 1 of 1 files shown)


torchqdynamics/solver_utils.py
```diff
@@ -34,19 +36,27 @@ def inv_sqrtm(mat: Tensor) -> Tensor:
...
-def bexpect(operators: Tensor, state: Tensor) -> Tensor:
-    """Compute the expectation values of many operators on a quantum state or
-    density matrix. The method is batchable over the operators and the state.
+def bexpect(O: Tensor, x: Tensor) -> Tensor:
+    """Compute the expectation values of many operators on a state vector or a density
+    matrix.
+
+    The expectation value $\braket{O}$ of a single operator $O$ is computed
+    - as $\braket{O}=\braket{\psi|O|\psi}$ if `x` is a state vector $\psi$,
+    - as $\braket{O}=\tr(O\rho)$ if `x` is a density matrix $\rho$.
```
gautierronan, line 45: A proposal, but not necessarely a good idea (not sure): use `$\braket{\hat{O}}$` instead of `$\braket{O}$`. It would be better for the documentation, but it's definitly less readable for the docstrings.

torchqdynamics/solver_utils.py
```diff
@@ -34,19 +36,27 @@ def inv_sqrtm(mat: Tensor) -> Tensor:
     return vecs @ torch.linalg.solve(vecs, torch.diag(vals ** (-0.5)), left=False)
 
 
-def bexpect(operators: Tensor, state: Tensor) -> Tensor:
-    """Compute the expectation values of many operators on a quantum state or
-    density matrix. The method is batchable over the operators and the state.
+def bexpect(O: Tensor, x: Tensor) -> Tensor:
+    """Compute the expectation values of many operators on a state vector or a density
+    matrix.
```
gautierronan, line 41:
```
Compute the expectation values of batched operators on a state vector or a density matrix.
```
is more clear I think (we understand what the `b` of `bexpect` stands for)

gautierronan (dismissed):
Two small suggestions, otherwise it's good. I like the conventions you propose !

torchqdynamics/solver_utils.py
PierreGuilmin: Hahaha I _hate_ the hats on top of operators, but that's a long discussion 😅 Let's discuss this next meeting 😈

gautierronan (approved):

PR #52, opened 2023-03-29
New dev tools

PierreGuilmin:
I picked a cute simple-config task runner, I don't think we need more. I will add `pylint`, `mypy` and the gears to build the doc later.


gautierronan (approved):

gautierronan (approved):
Looks good ! :)

PR #53, opened 2023-03-30
Proposal for a testing framework

PierreGuilmin:
Here's what I had in mind for the test framework, it may look like a big PR but it's not, a lot of the code is really straightforward/similar to read. I suggest reading `testssesolve/closed_system.py`, then `tests/sesolve/sesolver_test.py`, then `tests/sesolve/test_euler.py` to understand the logic, then the rest.

The proposal is to decouple the physical systems from the tested solvers. So we would have
- A few physical systems defining `H`, `jump_ops`, etc... (a qubit, a cavity, two coupled systems)
- Base test classes for the solver with helper methods.
- A test class for each solver that picks the relevant physical system to test something (e.g. batching, computed states correctness, stability, etc...).

I think what's important is that everything is optional: you can define a physical system with no analytical solution (but then you can't use it to check computed states correctness, obviously), you can test batching only for a specific solver, etc... Another important aspect is that we can here separate long-running tests that we will probably not run for each PR (for example checking a solver solution for long-time evolution) from quick tests that are just checking the logic.

And then I think we can use the physical systems to implement benchmarks between solvers (to monitor performance when a PR modifies the solvers code or the odeint code). I don't think we need this for now, but in a few months ~when a random guy from Patagonia makes a PR to propose some dark magic torch optimization~, it will be nice to have.

### DIFF
--- a/tests/mesolve/mesolver_test.py
+++ b/tests/mesolve/mesolver_test.py
@@ -54,8 +54,7 @@ class MESolverTest:
         system: OpenSystem,
         *,
         num_t_save: int,
-        rtol: float = 1e-05,
-        atol: float = 1e-08,
+        norm_atol: float = 1e-2,
     ):
         t_save = system.t_save(num_t_save)
 
@@ -68,7 +67,7 @@ class MESolverTest:
             solver=solver,
         )
 
-        assert torch.allclose(rho_save, system.rhos(t_save), rtol=rtol, atol=atol)
+        assert torch.norm(rho_save - system.rhos(t_save)) <= norm_atol
 
     def test_rho_save(self):
         pass
--- a/tests/mesolve/mesolver_test.py
+++ b/tests/mesolve/mesolver_test.py
@@ -1,0 +1,74 @@
+import torch
+
+import torchqdynamics as tq
+from torchqdynamics.solver_options import SolverOption
+
+from .open_system import OpenSystem
+
+
+class MESolverTest:
+    def _test_batching(self, solver: SolverOption, system: OpenSystem):
+        """Test the batching of `H` and `rho0` in `mesolve`, and the returned object
+        sizes."""
+        n = system.n
+        n_exp_ops = len(system.exp_ops)
+        b_H = len(system.H_batched)
+        b_rho0 = len(system.rho0_batched)
+        nt = 11
+
+        run_mesolve = lambda H, rho0: tq.mesolve(
+            H,
+            system.jump_ops,
+            rho0,
+            system.t_save(nt),
+            exp_ops=system.exp_ops,
+            solver=solver,
+        )
+
+        # no batching
+        rho_save, exp_save = run_mesolve(system.H, system.rho0)
+        assert rho_save.shape == (nt, n, n)
+        assert exp_save.shape == (n_exp_ops, nt)
+
+        # batched H
+        rho_save, exp_save = run_mesolve(system.H_batched, system.rho0)
+        assert rho_save.shape == (b_H, nt, n, n)
+        assert exp_save.shape == (b_H, n_exp_ops, nt)
+
+        # batched rho0
+        rho_save, exp_save = run_mesolve(system.H, system.rho0_batched)
+        assert rho_save.shape == (b_rho0, nt, n, n)
+        assert exp_save.shape == (b_rho0, n_exp_ops, nt)
+
+        # batched H and rho0
+        rho_save, exp_save = run_mesolve(system.H_batched, system.rho0_batched)
+        assert rho_save.shape == (b_H, b_rho0, nt, n, n)
+        assert exp_save.shape == (b_H, b_rho0, n_exp_ops, nt)
+
+    def test_batching(self):
+        pass
+
+    def _test_rho_save(
+        self,
+        solver: SolverOption,
+        system: OpenSystem,
+        *,
+        nt: int,
+        rtol: float = 1e-05,
+        atol: float = 1e-08,
+    ):
+        t_save = system.t_save(nt)
+
+        rho_save, _ = tq.mesolve(
+            system.H,
+            system.jump_ops,
+            system.rho0,
+            t_save,
+            exp_ops=system.exp_ops,
+            solver=solver,
+        )
+
+        assert torch.allclose(rho_save, system.rhos(t_save), rtol=rtol, atol=atol)
+
+    def test_rho_save(self):
+        pass
--- a/tests/mesolve/mesolver_test.py
+++ b/tests/mesolve/mesolver_test.py
@@ -14,36 +14,36 @@ class MESolverTest:
         n_exp_ops = len(system.exp_ops)
         b_H = len(system.H_batched)
         b_rho0 = len(system.rho0_batched)
-        nt = 11
+        num_t_save = 11
 
         run_mesolve = lambda H, rho0: tq.mesolve(
             H,
             system.jump_ops,
             rho0,
-            system.t_save(nt),
+            system.t_save(num_t_save),
             exp_ops=system.exp_ops,
             solver=solver,
         )
 
         # no batching
         rho_save, exp_save = run_mesolve(system.H, system.rho0)
-        assert rho_save.shape == (nt, n, n)
-        assert exp_save.shape == (n_exp_ops, nt)
+        assert rho_save.shape == (num_t_save, n, n)
+        assert exp_save.shape == (n_exp_ops, num_t_save)
 
         # batched H
         rho_save, exp_save = run_mesolve(system.H_batched, system.rho0)
-        assert rho_save.shape == (b_H, nt, n, n)
-        assert exp_save.shape == (b_H, n_exp_ops, nt)
+        assert rho_save.shape == (b_H, num_t_save, n, n)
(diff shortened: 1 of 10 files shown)


gautierronan:
> Our methods have quite a bad accuracy, I wonder if I take a Hamiltonian rotating too fast maybe?

It's because you use `torch.linspace`. See https://github.com/PierreGuilmin/torchqdynamics/issues/27#issuecomment-1487751801

gautierronan:
> I don't think so, it is much worse than the errors caused by the linspace `1e-6` error.

Well I don't know exactly what you're simulating, but if you have a `1e-6` error on `dt` and run for `1e3` time steps, then your final error builds up to `1e-3`. But there's for sure some problems in our `_fixed_odeint` function in this regard.

tests/mesolve/test_euler.py
```diff
@@ -1,32 +1,17 @@
...
-    solver = tq.solver.Euler(dt=1e-3)
-
-    # run solver
-    states, exp = tq.mesolve(H, jump_ops, rho0, t_save, exp_ops=exp_ops, solver=solver)
-    assert states.shape == (num_t_save, n, n)
-    assert exp.shape == (num_exp_ops, num_t_save)
+    @pytest.mark.long
+    def test_rho_save(self):
+        solver = tq.solver.Euler(dt=1e-5)
+        self._test_rho_save(solver, leaky_cavity_8, nt=51, rtol=1e0)
```
gautierronan, line 17: Why do you need `rtol=1e0` here? It looks a bit cheaty for a test

gautierronan (changes requested):
Very nice PR once more. Some minor comments and questions here and there.

Also, on a more general note it would be nice to have a set of utility functions for `tq`, such as coherent, basis, create, destroy, etc. This way we would avoid going back and forth between `tq`, `qt` and `np` during these tests. I will write a PR in the coming days.

tests/sesolve/closed_system.py
```diff
@@ -0,0 +1,64 @@
...
+        self.H_batched = None
+        self.psi0 = None
+        self.psi0_batched = None
+        self.exp_ops = None
+
+    @abstractmethod
+    def t_save(self, n: int) -> Tensor:
+        pass
+
+    def psi(self, t: float) -> Tensor:
```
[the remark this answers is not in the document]
PierreGuilmin: If it were then it would be mandatory to define this method (one can't instantiate an abstract class), but I thought we may have some physical example with no analytical solutions, for which we don't define this guy.

tests/mesolve/test_euler.py
PierreGuilmin: Haha it's just the lowest value that makes the test pass, probably due to the linspace issue as you explained 🤷‍♂️ We definitely need to fix this.

tests/sesolve/closed_system.py
```diff
@@ -0,0 +1,64 @@
...
+import numpy as np
+import qutip as qt
+import torch
+from torch import Tensor
+
+import torchqdynamics as tq
+
+
+class ClosedSystem(ABC):
+    def __init__(self):
```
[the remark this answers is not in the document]
PierreGuilmin: Right, I think we may want to promote the use of `torch.set_default_dtype` rather than passing dtypes everywhere, we can discuss this next meeting, I'll add it to the agenda. For tests it might be better to have a `dtype` argument, I chose not to care for now as we care nowhere in the code, but we definitely need to fix this at some point.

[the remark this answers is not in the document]
PierreGuilmin: TL;DR I think it needs a global discussion rather than just a fix here, so I propose to postpone 😉

tests/sesolve/closed_system.py
```diff
@@ -0,0 +1,64 @@
...
+        self.psi0_batched = [
+            qt.coherent(n, alpha0),
+            qt.coherent(n, 1j * alpha0),
+            qt.coherent(n, -alpha0),
+            qt.coherent(n, -1j * alpha0),
+        ]
+
+    def t_save(self, n: int) -> Tensor:
+        t_end = self.delta / (2 * np.pi)  # a full rotation
+        return torch.linspace(0.0, t_end, n)
```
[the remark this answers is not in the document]
PierreGuilmin: Same story as previous comment IMO, we need to have a proper discussion and fix throughout the code rather than just here.

tests/sesolve/closed_system.py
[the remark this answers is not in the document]
gautierronan: But then it could be the solution given by qutip for instance, no? In case we don't have an analytical solution.

tests/sesolve/closed_system.py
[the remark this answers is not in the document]
gautierronan: Alright let's organize a meeting next week if you guys are available. There's quite a few things to discuss by now I think.

tests/mesolve/mesolver_test.py
```diff
@@ -0,0 +1,74 @@
...
+        rho_save, _ = tq.mesolve(
+            system.H,
+            system.jump_ops,
+            system.rho0,
+            t_save,
+            exp_ops=system.exp_ops,
+            solver=solver,
+        )
+
+        assert torch.allclose(rho_save, system.rhos(t_save), rtol=rtol, atol=atol)
```
gautierronan, line 71: I don't think this is a fair test. A better test would be
```python
assert torch.norm(rho_save - system.rhos(t_save)) < 1e-2 * nt
```
Otherwise the relative tolerence of `torch.allclose` amplifies the small differences with small amplitudes (i.e. far off-diagonal components of the density matrix).

tests/mesolve/mesolver_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_batching(self):
+        pass
+
+    def _test_rho_save(
+        self,
+        solver: SolverOption,
+        system: OpenSystem,
+        *,
+        nt: int,
```
gautierronan, line 56: Can you rename `nt` to `num_tsteps` or something of this sort throughout the code? This avoids confusion between the Hilbert size `n` and `nt`.

tests/mesolve/open_system.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def rho(self, t: float) -> Tensor:
+        alpha_t = (
+            self.alpha0 * np.exp(-1j * self.delta * t) * np.exp(-self.kappa / 2 * t)
+        )
+        rho_t = qt.coherent(self.n, alpha_t)
+        return tq.ket_to_dm(tq.from_qutip(rho_t))
+
+
+leaky_cavity_8 = LeakyCavity(n=8, kappa=1.0, delta=2 * np.pi, alpha0=1.0)
```
gautierronan, line 69: I think this should be defined within the scope of each `TestMEEuler` and `TestMERouchonX` class. This way we can read and change all test parameters in one place.

tests/mesolve/test_euler.py
gautierronan, line 17: I think `nt=51` is a bit large for a test. Also, `dt=1e-5` is a bit small. With the test I propose above, you can take `tol=1e-1` and raise `dt=1e-4` I think.

gautierronan (changes requested):
Here's a second batch of comments on the PR. 

I investigated a bit more why your test was failing without `rtol=1e0` by testing both against qutip and against the analytical formula. The code inside `mesolve` works as intended, it's just that Rouchon1 converges slowly (it's just an order 1 solver, so it's quite expected), and that testing with `torch.allclose` is not completely fair (see my comment below).

Also, note that Rouchon1 and Rouchon2 are quite slow right now compared to qutip. I think we'll definitly have to look into torch.compile and sparse matrices at some point. Also, the high-order adaptive time step solver might help performance-wise.

tests/sesolve/closed_system.py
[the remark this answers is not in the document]
PierreGuilmin: Yes, but I prefer not to make it mandatory, I think we may want to test specific things on some systems without needing the full state (expectation values for example, or just that the solver is stable), so we should not make it mandatory to have a solution for `psi` or `rho`.

tests/mesolve/mesolver_test.py
PierreGuilmin: Right, I'm not sure about the `* nt` though, the correct thing to test would be against `dt` no? See proposed commit otherwise, just checking correctness independent of the time.

PierreGuilmin: Ho actually I see what you wanted to do, let me propose smthg which is better (check that each saved state is `< norm_atol` close to the correct one.

PierreGuilmin:
Thanks a lot for the review @gautierronan, this looks much better!

PierreGuilmin:
Maybe we should test for the common state fidelity (`ket_fidelity()` and `dm_fidelity()`) rather than the torch norm, wdyt?

abocquet (approved):

gautierronan:
> Maybe we should test for the common state fidelity (`ket_fidelity()` and `dm_fidelity()`) rather than the torch norm, wdyt?

I don't think so, it will slow down the test by having to compute the matrix square root, and the norm already checks that all elements of the two matrices are close.

PR #54, opened 2023-03-31
Various fixes

PierreGuilmin:
The `dm_fidelity()` is inspired from [qutip implementation](https://qutip.org/docs/latest/modules/qutip/metrics.html#fidelity) and tested against QuTiP, and for batched matrices:
```python
import qutip as qt
import torch
import torchqdynamics as tq

b1, b2, n = 3, 2, 5
rho = [[qt.rand_dm(n, n) for _ in range(b2)] for _ in range(b1)]  # (b1, b2, n, n)
sigma = [[qt.rand_dm(n, n) for _ in range(b2)] for _ in range(b1)]  # (b1, b2, n, n)
[[qt.fidelity(r, s) ** 2 for r, s in zip(rr, ss)] for rr, ss in zip(rho, sigma)]
# [[0.7320792219355124, 0.629518078418219],
#  [0.5630845960869209, 0.34414775783174417],
#  [0.49023424734517823, 0.5439257254888179]]

torch.set_printoptions(precision=16)
tq.dm_fidelity(tq.types.to_tensor(rho), tq.types.to_tensor(sigma))
# tensor([[0.7320792219355172, 0.6295180784182186],
#         [0.5630845960869211, 0.3441477578319131],
#         [0.4902342473451624, 0.5439257254884318]], dtype=torch.float64)
```


torchqdynamics/solver_options.py
```diff
@@ -1,14 +1,15 @@
...
 @dataclass
 class SolverOption:
-    pass
+    verbose: bool = True
 
 
 @dataclass
 class FixedStep(SolverOption):
-    dt: float
+    dt: Optional[float] = None
```
gautierronan, line 12: Why is this optional? Shouldn't it always be passed by the user?

torchqdynamics/utils.py
```diff
@@ -46,10 +46,28 @@ def ket_overlap(x: Tensor, y: Tensor) -> torch.complex:
...
+        x: Tensor of dimension `(..., n, n)`.
+
+    Returns:
+        Tensor of dimension `(..., n, n)` holding the square root of `x`.
+    """
+    # code copied from
+    # https://github.com/pytorch/pytorch/issues/25481#issuecomment-1032789228
+    L, Q = torch.linalg.eigh(x)
+    zero = torch.zeros((), device=L.device, dtype=L.dtype)
+    threshold = L.max(-1).values * L.size(-1) * torch.finfo(L.dtype).eps
```
gautierronan, line 62: Instead of `torch.finfo`, I think it's more readable to import `from torch._prims_common import corresponding_real_dtype as to_real` and to run `to_real(L.dtype)`.

FYI, I made a PR on the pytorch github to implement `dtype.to_real` and `dtype.to_complex` methods: https://github.com/pytorch/pytorch/pull/97935 for exactly this problem.

gautierronan, line 62: Instead of `torch.finfo`, I think it's more readable to import `from torch._prims_common import corresponding_real_dtype as to_real` and to run `to_real(L.dtype)`.

FYI, I made a PR on the pytorch github to implement `dtype.to_real` and `dtype.to_complex` methods: https://github.com/pytorch/pytorch/pull/97935 for exactly this feature.

torchqdynamics/utils.py
```diff
@@ -46,10 +46,28 @@ def ket_overlap(x: Tensor, y: Tensor) -> torch.complex:
...
+    L = L.where(L > threshold.unsqueeze(-1), zero)  # zero out small components
+    return (Q * L.sqrt().unsqueeze(-2)) @ Q.mH
```
gautierronan, line 64: I don't know how important this is, but an sqrtm implementation based on `eigh` can be a bit slow, and also unstable numerically (probably why you have to zero out small components). Maybe a better implementation would be the following, based on a Newton method:
```python
class MatrixSquareRoot(Function):
    """Square root of a positive definite matrix.
    NOTE: matrix square root is not differentiable for matrices with
          zero eigenvalues.
    """    
    @staticmethod
    def forward(ctx, input):
        dim = input.shape[0]
        norm = torch.norm(input)
        Y = input/norm
        I = torch.eye(dim,dim,device=input.device).type(input.dtype)
        Z = torch.eye(dim,dim,device=input.device).type(input.dtype)
        for i in range(20):
            T = 0.5*(3.0*I - Z.mm(Y))
            Y = Y.mm(T)
            Z = T.mm(Z)
        sqrtm = Y*torch.sqrt(norm)
        ctx.mark_dirty(Y,I,Z)
        ctx.save_for_backward(sqrtm)
        return sqrtm

    @staticmethod
    def backward(ctx, grad_output):
        grad_input = None
        sqrtm, = ctx.saved_tensors
        dim = sqrtm.shape[0]
        norm = torch.norm(sqrtm)
        A = sqrtm/norm
        I = torch.eye(dim, dim, device=sqrtm.device).type(input.dtype)
        Q = grad_output/norm
        for i in range(20):
            Q = 0.5*(Q.mm(3.0*I-A.mm(A))-A.t().mm(A.t().mm(Q)-Q.mm(A)))
            A = 0.5*A.mm(3.0*I-A.mm(A))
        grad_input = 0.5*Q
        return grad_input    

symsqrt = MatrixSquareRoot.apply
```
This is faster, stable, and also supports backward differentiation. However, one has to manage the number of iteration of the process, which may be a bit tedious. Alternatively, we could support both implementations with a keyword argument in `sqrtm`. 

The code is copied from https://github.com/pytorch/pytorch/issues/25481#issuecomment-584896176.

gautierronan, line 64: Just checking, but should it be `(Q @ L.sqrt().unsqueeze(-2)) @ Q.mH` instead?

torchqdynamics/utils.py
```diff
@@ -59,14 +77,54 @@ def ket_fidelity(x: Tensor, y: Tensor) -> float:
...
+
+    Returns:
+        Tensor of size `(...)` holding the real-valued fidelity.
+    """
+    sqrtm_x = sqrtm(x)
+    tmp = sqrtm_x @ y @ sqrtm_x
+
+    # we don't need the whole matrix `sqrtm(tmp)`, just its trace, which can be computed
+    # by summing the square roots of `tmp` eigenvalues
+    eig_vals = torch.linalg.eigvalsh(tmp)
```
gautierronan, line 120: Maybe a small renaming from `eig_vals` to `eigvals_tmp` would be clearer here.

gautierronan (changes requested):
Very nice PR ! Mainly one comment about the way to compute sqrtm.

torchqdynamics/solver_options.py
PierreGuilmin: Because of the way dataclass inheritance is made, now that `SolverOption` has a keyword argument (`verbose`) we **have to** use kw arguments only in the child classes.

> Note that all solver options must be keyword arguments because the base class has a kw args (csqce of dataclasses > https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5).

See also commit description 228276c191209315f477897354787728464117ff.

PierreGuilmin: Because of the way dataclass inheritance is made, now that `SolverOption` has a keyword argument (`verbose`) we **have to** use kw arguments only in the child classes.

> Note that all solver options must be keyword arguments because the base class has a kw args (csqce of dataclasses > https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5).

See also commit description 228276c191209315f477897354787728464117ff.

torchqdynamics/utils.py
PierreGuilmin: Nice! I'll let you add this in another PR 😉 (the proposed version supports backward differentiation with the addition of the incoming fix)

PierreGuilmin: I don't see the difference?

gautierronan: with the @ instead of * in the first multiplication

torchqdynamics/solver_options.py
gautierronan: In this case we need to add a check in the `__init__` of fixed time step solvers for None-like `dt`values, and raise an error in case `dt` is not defined.

torchqdynamics/utils.py
PierreGuilmin: No I think this is correct (`L` has size `(n)` and `Q` has size `(n, n)`.

torchqdynamics/solver_options.py
PierreGuilmin: Actually I think we should not use dataclasses, see new commit.

PierreGuilmin: (it's very confusing for the user to see a field being default initialized to `None`, but then the code yells at him because he didn't set its value 😅 )

abocquet (approved):

gautierronan (approved):
Allright this is all good for me ! I'm keeping the `sqrtm` in my to-do list but it's not really a priority. Also I'm fine with removing the dataclass, although I agree it was kind of neat. Maybe there's a trick with a `from __future__ import ...` as well to fix the inheritence?

PierreGuilmin:
> Maybe there's a trick with a from __future__ import ... as well to fix the inheritence?

I looked a bit online, but couldn't find anything satisfying. 🤷‍♂️

PR #55, opened 2023-03-31
Define a default progress bar format

gautierronan:
With this PR, the text after the tqdm progress bar looks like
```python
37.3% - Time 00:11/00:19
```

### DIFF
--- a/torchqdynamics/progressbar.py
+++ b/torchqdynamics/progressbar.py
@@ -1,0 +1,9 @@
+from functools import partial
+
+from tqdm.auto import tqdm as std_tqdm
+
+# Define a default progress bar format
+PBAR_FORMAT = '{desc}: {percentage:3.1f}% - Time {elapsed}/{remaining}'
+
+# Redefine tqdm with some default arguments
+tqdm = partial(std_tqdm, bar_format=PBAR_FORMAT)
--- a/torchqdynamics/progressbar.py
+++ b/torchqdynamics/progressbar.py
@@ -1,9 +1,0 @@
-from functools import partial
-
-from tqdm.auto import tqdm as std_tqdm
-
-# Define a default progress bar format
-PBAR_FORMAT = '{desc}: {percentage:3.1f}% - Time {elapsed}/{remaining}'
-
-# Redefine tqdm with some default arguments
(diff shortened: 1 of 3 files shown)


abocquet:
Nice initiative. To this point maybe we could directly code a `torchqdiffeq.tqdm` function that calls tqdm with the correct formatting to avoid copy pasting it everywhere and be easier to add more options later ?

gautierronan:
> Nice initiative. To this point maybe we could directly code a `torchqdynamics.tqdm` function that calls tqdm with the correct formatting to avoid copy pasting it everywhere and be easier to add more options later ?

Yes maybe something more like the last commit?

abocquet (dismissed):
Yes, this is exactly what I had in mind

torchqdynamics/progressbar.py
PierreGuilmin, line 1: Could we rename this file `progress_bar.py` to follow our naming convention?

torchqdynamics/progressbar.py
```diff
@@ -0,0 +1,9 @@
+from functools import partial
+
+from tqdm.auto import tqdm as std_tqdm
```
PierreGuilmin, line 3: I'm really not fan of the notebook progress bar viewer (too big and colorfull IMO, plus it doesn't work well with dark mode), could we switch to have the regular progress bar everywhere (`from tqdm import tqdm`)?

Screenshot 2023-04-03 at 16 31 40

torchqdynamics/progressbar.py
```diff
@@ -0,0 +1,9 @@
...
+
+# Define a default progress bar format
+PBAR_FORMAT = '{desc}: {percentage:3.1f}% - Time {elapsed}/{remaining}'
+
+# Redefine tqdm with some default arguments
```
PierreGuilmin, line 8:
```suggestion
# redefine tqdm with some default arguments
```

PierreGuilmin (changes requested):
Nice implem!

gautierronan:
In the end I went with `|██████████| 100.0% - time 00:05/00:00`, and removed the `{desc}` since it's not implemented yet.

PR #57, opened 2023-04-01
Various utility functions

gautierronan:
This PR introduces a bunch of basic utility functions, both for operator and state initialization. Nothing particulary difficult. While reviewing, mostly check if you agree with the naming conventions and the behavior of the various functions introduced.


abocquet (dismissed):

gautierronan:
By mistake I added commit https://github.com/PierreGuilmin/torchqdynamics/pull/57/commits/cb35a45e434ac00e924c69955b31bafc2f4b45b7 to this PR instead of another PR. It's just the typing update to 3.10 annotations, so I guess it's simple enough. Not sure how to revert properly so I keep it inside this PR.

torchqdynamics/operators.py
```diff
@@ -0,0 +1,216 @@
...
+    return torch.tensor([[0.0, -1.0j], [1.0j, 0.0]], dtype=dtype, device=device)
+
+
+def sigmaz(*, dtype: dtype = torch.complex128, device: device | None = None) -> Tensor:
+    """Pauli Z operator."""
+    return torch.tensor([[1.0, 0.0], [0.0, -1.0]], dtype=dtype, device=device)
+
+
+def sigmap(*, dtype: dtype = torch.complex128, device: device | None = None) -> Tensor:
+    """Pauli raising operator."""
```
PierreGuilmin, line 40:
```suggestion
    """Pauli raising operator $\sigma_+$."""
```
Same for lowering.

torchqdynamics/operators.py
```diff
@@ -0,0 +1,216 @@
...
+    return torch.tensor([[0.0, 0.0], [1.0, 0.0]], dtype=dtype, device=device)
+
+
+def qeye(
+    *dims: int, dtype: dtype = torch.complex128, device: device | None = None
+) -> Tensor:
+    """Identity operator."""
+    dims = _extract_tuple_from_varargs(dims)
+    dim = prod(dims)
+    return torch.eye(dim, dtype=dtype, device=device)
```
PierreGuilmin, line 55: For example, after the proposed changes:
```suggestion
def qeye(*dims: int) -> Tensor:
    """Identity operator."""
    return torch.eye(prod(dims))
```

PierreGuilmin: Should we call this guy `eye` directly or `identity`? I never liked `qeye`.

torchqdynamics/operators.py
```diff
@@ -0,0 +1,216 @@
...
+
+def destroy(
+    *dims: int, dtype: dtype = torch.complex128, device: device = None
+) -> Tensor | tuple[Tensor, ...]:
+    """Bosonic annihilation operator.
+
+    If only a single dimension is provided, `destroy` returns the annihilation operator
+    of corresponding dimension. If instead multiples dimensions are provided, `destroy`
+    returns a tuple of each annihilation operator of given dimension, in the Hilbert
+    space given by the product of all dimensions.
```
PierreGuilmin, line 66: Same, is this really a needed feature? We can build all of these very easily with `tensprod` and it's more readable IMO.

torchqdynamics/operators.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def displace(
+    dim: int,
+    alpha: complex,
+    *,
+    dtype: dtype = torch.complex128,
+    device: device | None = None,
+) -> Tensor:
+    """Single-mode displacement operator."""
```
PierreGuilmin, line 201:
```suggestion
    """Displacement operator."""
```
I think it's always single mode by default, no need to precise it everywhere.

PierreGuilmin (changes requested):
Amazing PR thank you 🤩  The new typing annotations are so much easier to read, and we can finally write examples and tests with our own functions 👏

Concerning the `dtype` and `device` arguments, I think we should remove them everywhere. The best way IMO not to clutter the code would be to make a decorator, but anyway I think we should rather use and advertise the global functions `set_default_device` and `set_default_tensor_type` rather than annoying ourselves everywhere with devices and types.

torchqdynamics/states.py
```diff
@@ -0,0 +1,116 @@
...
+                [1.+0.j],
+                [0.+0.j]], dtype=torch.complex128)
+        >>> tq.fock((3, 2), (1, 0))
+        tensor([[0.+0.j],
+                [0.+0.j],
+                [1.+0.j],
+                [0.+0.j],
+                [0.+0.j],
+                [0.+0.j]], dtype=torch.complex128)
+    """
```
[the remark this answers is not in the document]
gautierronan: I think it's a neat feature which I personnaly like using. People can still use your version if they prefer. Especially for the annihilation operators, I really like
```
a, b, c = tq.destroy(Na, Nb, Nc)
```
which is way shorter and readable than
```
ia_ = tq.eye(Na)
ib_ = tq.eye(Nb)
ic_ = tq.eye(Nc)
a_ = tq.destroy(Na)
b_ = tq.destroy(Nb)
c_ = tq.destroy(Nc)
a = tq.tensprod(a_, ib_, ic_)
b = tq.tensprod(ia_, b_, ic_)
c = tq.tensprod(ia_, ib_, c_)
```
(which is literally what I do with qutip personnaly)

torchqdynamics/operators.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def _destroy_single(
+    dim: int,
+    *,
+    dtype: torch.dtype = torch.complex128,
+    device: torch.device | None = None,
+) -> Tensor:
+    """Bosonic annihilation operator."""
+    return torch.diag(torch.sqrt(torch.arange(1, dim, device=device)), 1).to(dtype)
```
PierreGuilmin, line 135:
```suggestion
    return torch.arange(1, dim, device=device).sqrt().diag(1).to(dtype)
```

PierreGuilmin: Same for `_create_single`.

torchqdynamics/operators.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # update ops
+        ops.append(tensprod(identity, _a))
+        for j in range(i + 1):
+            ops[j] = tensprod(ops[j], _id)
+
+        # update identity
+        identity = tensprod(identity, _id)
+
+    return tuple(ops)
```
PierreGuilmin, line 125: Here's a refinement (also work for a single operator):
```python
a = [_destroy_single(dim, dtype=dtype, device=device) for dim in dims]
I = [eye(dim, dtype=dtype, device=device) for dim in dims]
return (tq.tensprod(*[I[i] if i != j else a[j] for j in range(len(dims))]) for i in range(len(dims)))
```

PierreGuilmin (changes requested):
Minor comments.

PierreGuilmin:
I will make a subsequent PR to fix the repeated `dtype` and `device` by using a decorator. It will also set the correct default complex type using `torch.get_default_dtype` rather than overriding it by default. No need to change anything in this PR.

PierreGuilmin (approved):

PR #58, opened 2023-04-03
Time dependent Hamiltonian | Part 1

abocquet:
Replaces #35. The philosophy is very close but the decorators logics is much lighter, as the only accepted variable is now time.

Only `MERouchon1` takes advantage of the caching for now, as an example. Others will come in another PR, but are compatible with callable `H`

I'm not very satisfied with the way the duality Tensor/callable/list of callables is handled right now. Any idea is welcomed

For next PRs:
- [ ] piecewise constant H
- [ ] cache all numerical schemas


.gitignore
```diff
@@ -22,3 +22,6 @@ dist/
 
 # pyenv local Python version
 .python-version
+
+# pytest cache
+.pytest_cache
```
PierreGuilmin, line 27: Missing newline.

PierreGuilmin, line 27:
```suggestion
.pytest_cache/
```

pyproject.toml
```diff
@@ -27,13 +27,14 @@ packages = ["torchqdynamics"]
 
 [tool.isort]
 profile = "black"
-skip = ["__init__.py"]
+skip = ["__init__.py", "venv"]
```
PierreGuilmin, line 30: I think we should rather specify the directories `isort` and `black` should explore (`tests` and `torchqdynamics`) rather than excluding the one they should avoid, let me fix it in a subsequent PR if you agree.

torchqdynamics/mesolve/rouchon.py
```diff
@@ -1,11 +1,11 @@
-from math import sqrt
 from typing import Tuple
 
 import torch
 import torch.nn as nn
+from numpy import sqrt
```
PierreGuilmin, line 5: Could we use `torch.sqrt`?

torchqdynamics/mesolve/rouchon.py
```diff
@@ -41,21 +44,34 @@ def forward(self, t: float, rho: Tensor) -> Tensor:
...
+
+    @H_dependent
+    def H_nh(self, t):
+        # non-hermitian Hamiltonian at time t
+        return self.H(t) - 0.5j * self.sum_nojump
+
+    @H_dependent
+    def M0_adj(self, t):
+        return self.I + 1j * self.dt * self.Hdag_nh(t)
+
```
PierreGuilmin, line 74: Let's type all this (`def ...(self, t: float) -> Tensor`).

torchqdynamics/mesolve/mesolve.py
```diff
@@ -82,9 +82,17 @@ def mesolve(
...
+    if isinstance(H, OperatorLike.__args__):
+        # convert H to a tensor and batch by default
+        H = to_tensor(H)
+        H_batched = H[None, ...] if H.ndim == 2 else H
+    elif callable(H) or (isinstance(H, list) and (len(H) == 0 or callable(H[0]))):
+        # H is callable or list of callables
+        H_batched = H
+    else:
+        raise ValueError(
+            'H must be a tensor-like object or a callable or list of callables'
```
PierreGuilmin, line 94:
```suggestion
            'H must be a tensor-like object or a callable or list of callables.'
```

torchqdynamics/odeint.py
```diff
@@ -11,6 +12,29 @@
...
+
+def H_dependent(func):
+    """
+    Handles caching for functions that only depend on the Hamiltonian.
+    The functions must take as single argument the time and be decorated
+    by `@H_cached`. For instance::
+        @H_cached
+        def H_squared(self, t):
+            return self.H(t) @ self.H(t)
+    """
```
PierreGuilmin, line 26: See https://github.com/PierreGuilmin/torchqdynamics/pull/37#discussion_r1148350366 for the doc style. I need to edit the `CONTRIBUTING.md` with all the new conventions.

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
...
+        self._cache = {}
+
+    @H_dependent
+    def H(self, t):
+        if isinstance(self._H, OperatorLike.__args__):
+            return self._H
+        elif callable(self._H):
+            return self._H(t)
+        else:
+            raise Exception('Piece-wise constant Hamiltonians not supported yet')
```
PierreGuilmin, line 73:
```suggestion
            raise TypeError('Piecewise constant Hamiltonian not supported yet.')
```

torchqdynamics/mesolve/mesolve.py
```diff
@@ -82,9 +82,17 @@ def mesolve(
     """
     # TODO H is assumed to be time-independent from here (temporary)
```
PierreGuilmin, line 83: Remove this.

torchqdynamics/mesolve/mesolve.py
```diff
@@ -82,9 +82,17 @@ def mesolve(
...
 
-    # convert H to a tensor and batch by default
-    H = to_tensor(H)
-    H_batched = H[None, ...] if H.dim() == 2 else H
+    if isinstance(H, OperatorLike.__args__):
+        # convert H to a tensor and batch by default
+        H = to_tensor(H)
+        H_batched = H[None, ...] if H.ndim == 2 else H
+    elif callable(H) or (isinstance(H, list) and (len(H) == 0 or callable(H[0]))):
+        # H is callable or list of callables
```
PierreGuilmin, line 90: If you want to add the bones for PWC H we need to edit the `TDOperatorLike` type.

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
...
+        if isinstance(H, OperatorLike.__args__):
+            if H.ndim == 3:
+                H = H[:, None, ...]  # (b_H, 1, n, n)
+        self._H = H
+        self._previous_t = False
+        self._cache = {}
+
+    @H_dependent
+    def H(self, t):
+        if isinstance(self._H, OperatorLike.__args__):
```
PierreGuilmin, line 68:
```suggestion
        if isinstance(self._H, Tensor):
```

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
         """
         pass
 
+    def __init__(self, H):
+        """
+        Args:
+            H: Hamiltonian, of shape `(b_H, n, n)`.
+        """
+        if isinstance(H, OperatorLike.__args__):
```
PierreGuilmin, line 59: From this point onward `H` is a `torch.Tensor`.
```suggestion
        if isinstance(H, Tensor):
```

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
...
+            if H.ndim == 3:
+                H = H[:, None, ...]  # (b_H, 1, n, n)
+        self._H = H
+        self._previous_t = False
```
PierreGuilmin, line 63: Let's remove this for now, and introduce it in the PWC PR.

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
...
+    def H(self, t):
+        if isinstance(self._H, OperatorLike.__args__):
+            return self._H
+        elif callable(self._H):
+            return self._H(t)
+        else:
+            raise Exception('Piece-wise constant Hamiltonians not supported yet')
+
+    def _H_changed(self, _t):
+        if isinstance(self._H, OperatorLike.__args__):
```
PierreGuilmin, line 76:
```suggestion
        if isinstance(self._H, Tensor):
```

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
...
+    @H_dependent
+    def H(self, t):
+        if isinstance(self._H, OperatorLike.__args__):
+            return self._H
+        elif callable(self._H):
+            return self._H(t)
+        else:
+            raise Exception('Piece-wise constant Hamiltonians not supported yet')
+
+    def _H_changed(self, _t):
```
PierreGuilmin, line 75:
```suggestion
    def _H_changed(self, t):
```

torchqdynamics/odeint.py
```diff
@@ -27,6 +51,35 @@ def forward(self, t: float, y: Tensor) -> Tensor:
         """
         pass
 
+    def __init__(self, H):
```
PierreGuilmin, line 54: Could we type all functions here?

PierreGuilmin (changes requested):
Amazing!! I love the new version, the code will run **so much** faster, and I think the solvers are easy to read. 🥳 

A few remarks and a first batch of comments:
- I think I like better the previous decorator naming `@depends_on_H`.
- We need to rebase this on main at some point, but I don't think it will impact the code too much, we can maybe wait for @gautierronan utils PR's to be merged before.
- The code handling `H` typing at the beginning of `mesolve` should be copied to `sesolve`. Actually, the best would be to move all this logic in the `ForwardQSolver.__init__()` method, so it's not repeated. And even better would be to move the specific gears converting a `TDOperatorLike` to a `TDOperator` in `types.py`. Then we have a clear object with a define type that we can use in `ForwardQSolver`.

pyproject.toml
PierreGuilmin: See https://github.com/PierreGuilmin/torchqdynamics/pull/65.

torchqdynamics/odeint.py
abocquet: This silences the IDE warnings because the variable is unused for now ;)

torchqdynamics/odeint.py
abocquet: Not sure what you want to change here

.gitignore
```diff
@@ -22,3 +22,6 @@ dist/
 
 # pyenv local Python version
 .python-version
+
+# pytest cache
+.pytest_cache/
```
PierreGuilmin, line 27: Missing newline at the end of the file.

torchqdynamics/mesolve/mesolve.py
PierreGuilmin: I suggest removing the part handling the piecewise constant `H~, because there are a few other things to change, and I would prefer that we do everything in the same PR, after this one.

torchqdynamics/mesolve/rouchon.py
```diff
@@ -41,21 +44,32 @@ def forward(self, t: float, rho: Tensor) -> Tensor:
...
         # compute rho(t+dt)
-        rho = kraus_map(rho, M0) + kraus_map(rho, M1s)
+        rho = kraus_map(rho, self.M0(t)) + kraus_map(rho, self.M1s)
 
         # normalize by the trace
         rho = rho / trace(rho)[..., None, None].real
 
         return rho
 
+    @H_dependent
```
PierreGuilmin, line 55: I think you missed one of my previous comment:

> I think I like better the previous decorator naming `@depends_on_H`.

torchqdynamics/mesolve/mesolve.py
```diff
@@ -95,7 +102,7 @@ def mesolve(
     rho0 = to_tensor(rho0)
     if is_ket(rho0):
         rho0 = ket_to_dm(rho0)
-    b_H = H_batched.size(0)
+    b_H = H_batched.size(0) if isinstance(H, TensorLike.__args__) else len(H)
```
PierreGuilmin, line 105:
```suggestion
    b_H = len(H)
```

PierreGuilmin (changes requested):

torchqdynamics/odeint.py
```diff
@@ -27,6 +50,34 @@ def forward(self, t: float, y: Tensor) -> Tensor:
...
+                H = H[:, None, ...]  # (b_H, 1, n, n)
+        self._H = H
+        self._cache = {}
+
+    @H_dependent
+    def H(self, t: float) -> Tensor:
+        if isinstance(self._H, Tensor):
+            return self._H
+        elif callable(self._H):
+            return self._H(t)
```
gautierronan, line 69: I think the batching of H is not handled if it's a callable. So typically if the user sends a time-dependent 2D tensor, it should be converted somewhere to a 4D tensor of size (1, 1, n, n). Just like we do in the time-independent case.

gautierronan (changes requested):
Sorry I'm only reading this now. It looks nice to me, it's a very nice addition I think. I just have one comment about the batching, but maybe you checked and I missed it.

PR #63, opened 2023-04-09
Various fixes and improvements

gautierronan:
Here is a PR with a bunch of small fixes and improvements:

- Improved the error message on non-aligned `t_save` and `dt`
- Allow single operators for `jump_ops` and `exp_ops`
- Change the default solver inside `mesolve` to the adaptive time step solver
- Fix a few small bugs in the adaptive solver
- Added a few comments inside `mesolve` and `sesolve`, and updated the documentation with the new solvers.
- Fixes the naming issue of https://github.com/PierreGuilmin/torchqdynamics/issues/61
- Fixes the wrong time ordering of https://github.com/PierreGuilmin/torchqdynamics/issues/51 (still lacking a proper test however)


torchqdynamics/mesolve/mesolve.py
```diff
@@ -18,12 +18,12 @@
 
 def mesolve(
     H: TDOperatorLike,
-    jump_ops: list[OperatorLike],
+    jump_ops: OperatorLike | list[OperatorLike],
```
PierreGuilmin, line 21: I don't think we should allow a single `jump_ops` or a single `exp_ops`, it makes the object harder to understand for someone reading the code, these guys really should be lists even if they have a single element.

torchqdynamics/mesolve/mesolve.py
```diff
@@ -84,30 +86,37 @@ def mesolve(
     """
     # TODO H is assumed to be time-independent from here (temporary)
 
+    # default solver
+    if solver is None:
+        solver = Dopri45()
```
PierreGuilmin, line 91: The ordering of the different blocks of code (converting `H` to a tensor, then raising errors for `jump_ops`, etc...) was chosen to follow the arguments of the function, I think it's better to preserve it to ease the reading.

torchqdynamics/mesolve/mesolve.py
PierreGuilmin: (same for sesolve)

torchqdynamics/solver_utils.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from math import sqrt
```
PierreGuilmin, line 3: When possible I think we should favour torch functions.

torchqdynamics/mesolve/adaptive.py
```diff
@@ -15,4 +15,10 @@ def __init__(self, H: TDOperator, jump_ops: Tensor, solver_options: SolverOption
...
     def forward(self, t: float, rho: Tensor) -> Tensor:
         """Compute drho / dt = L(rho) at time t."""
-        return lindbladian(rho, self.H, self.jump_ops)
+        # non-hermitian Hamiltonian
+        H_nh = self.H - 0.5j * self.sum_nojump
+
+        # compute Lindblad(t) @ rho
+        H_nh_rho = H_nh @ rho
+        L_rho_Ldag = kraus_map(rho, self.jump_ops)
+        return -1j * (H_nh_rho - H_nh_rho.adjoint()) + L_rho_Ldag
```
PierreGuilmin, line 24: Why not using the Lindbladian directly?

PierreGuilmin (changes requested):

torchqdynamics/solver_utils.py
abocquet: `torch.sqrt` does not support floats

```
In [1]: import torch

In [2]: torch.sqrt(4.0)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 torch.sqrt(4.0)

TypeError: sqrt(): argument 'input' (position 1) must be Tensor, not float
```

PierreGuilmin: Ho ok haha, sorry.

torchqdynamics/mesolve/mesolve.py
gautierronan: I really like having this feature. It copies qutip behavior, it's really handy and it doesn't change much for us inside the code. I'm sure a lot of people will think this is the default behavior, so it's annoying if they get an error every time.

torchqdynamics/mesolve/adaptive.py
gautierronan: Because `lindbladian` was recomputing the `sum_nojump` operator every time which is not needed. So it was faster to rewrite it directly inside here.

torchqdynamics/mesolve/mesolve.py
PierreGuilmin: Alright 👍

PierreGuilmin:
Ho you also need to fix the import in `tests.test_utils`.

PierreGuilmin (approved):

PR #66, opened 2023-04-17
Fix `dtype` and `device` throughout the code

PierreGuilmin:
Written on top of #63. This PR fixes the creation of tensors to rigorously handle `dtype` and `device`. Basically, we must use `dtype` and `device` anywhere we create a tensor, whether it's being directly built or converted from another data type. The `dtype`/`device` is either set by the context (e.g. it has the same type as another tensor) or from arguments passed to the function/class creating the tensor.

- I removed the typing of `dtype` and `device` for objects decorated by `@complex_tensor`, mainly to ease readability (and because static type checking should work through the decorator).
- @gautierronan do we also need the `dtype`/`device` for the `none_to_zeros_like` function? I would say yes.
- I checked that the typing propagates correctly through `mesolve` and `sesolve` (i.e. the returned objects have the correct type, and we can set it independently of the default type).
- We should also fix the creation of tensors in the tests, but I'll do it in an incoming PR where we switch from QuTiP functions to ours.

### DIFF
--- a/torchqdynamics/mesolve/mesolve.py
+++ b/torchqdynamics/mesolve/mesolve.py
@@ -27,6 +27,8 @@ def mesolve(
     solver: SolverOption | None = None,
     gradient_alg: Literal['autograd', 'adjoint'] | None = None,
     parameters: tuple[nn.Parameter, ...] | None = None,
+    dtype: torch.complex64 | torch.complex128 | None = None,
+    device: torch.device | None = None,
 ) -> tuple[Tensor, Tensor]:
     """Solve the Lindblad master equation for a Hamiltonian and set of jump operators.
 
@@ -73,6 +75,10 @@ def mesolve(
             backward pass. Defaults to `None`.
         parameters (tuple of nn.Parameter): Parameters with respect to which gradients
             are computed during the adjoint state backward pass.
+        dtype (torch.dtype, optional): Data type of the complex tensors (i.e. all
+            arguments but `t_save`, whose type is inferred from the underlying data
+            type).
+        device (torch.device, optional): Device on which the tensors are stored.
 
     Returns:
         A tuple `(rho_save, exp_save)` where
@@ -87,7 +93,7 @@ def mesolve(
     # TODO H is assumed to be time-independent from here (temporary)
 
     # convert H to a tensor and batch by default
-    H = to_tensor(H)
+    H = to_tensor(H, dtype=dtype, device=device, is_complex=True)
     H_batched = H[None, ...] if H.ndim == 2 else H
 
     # convert jump_ops to a tensor
@@ -96,11 +102,11 @@ def mesolve(
             'Argument `jump_ops` must be a non-empty list of tensors. Otherwise,'
             ' consider using `sesolve`.'
         )
-    jump_ops = to_tensor(jump_ops)
+    jump_ops = to_tensor(jump_ops, dtype=dtype, device=device, is_complex=True)
     jump_ops = jump_ops[None, ...] if jump_ops.ndim == 2 else jump_ops
 
     # convert rho0 to a tensor and density matrix and batch by default
-    rho0 = to_tensor(rho0)
+    rho0 = to_tensor(rho0, dtype=dtype, device=device, is_complex=True)
     if is_ket(rho0):
         rho0 = ket_to_dm(rho0)
     b_H = H_batched.size(0)
@@ -108,10 +114,10 @@ def mesolve(
     rho0_batched = rho0_batched[None, ...].repeat(b_H, 1, 1, 1)  # (b_H, b_rho0, n, n)
 
     # convert t_save to a tensor
-    t_save = torch.as_tensor(t_save)
+    t_save = torch.as_tensor(t_save, device=device)
 
     # convert exp_ops to a tensor
-    exp_ops = to_tensor(exp_ops)
+    exp_ops = to_tensor(exp_ops, dtype=dtype, device=device, is_complex=True)
     exp_ops = exp_ops[None, ...] if exp_ops.ndim == 2 else exp_ops
 
     # default solver
--- a/torchqdynamics/mesolve/mesolve.py
+++ b/torchqdynamics/mesolve/mesolve.py
@@ -8,7 +8,13 @@ from torch import Tensor
 
 from ..odeint import odeint
 from ..solver_options import AdaptiveStep, Dopri45, Euler, SolverOption
-from ..tensor_types import OperatorLike, TDOperatorLike, TensorLike, to_tensor
+from ..tensor_types import (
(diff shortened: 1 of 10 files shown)


torchqdynamics/mesolve/mesolve.py
```diff
@@ -73,6 +75,10 @@ def mesolve(
             backward pass. Defaults to `None`.
         parameters (tuple of nn.Parameter): Parameters with respect to which gradients
             are computed during the adjoint state backward pass.
+        dtype (torch.dtype, optional): Data type of the complex tensors (i.e. all
+            arguments but `t_save`, whose type is inferred from the underlying data
+            type).
```
gautierronan, line 80: Alternative proposal, maybe a bit more clear:
```
dtype (torch.dtype, optional): Complex data type to which all complex-valued tensors are converted. `t_save` is also converted to a real data type of the same precision.
```

torchqdynamics/mesolve/mesolve.py
```diff
@@ -96,22 +102,22 @@ def mesolve(
...
+    rho0 = to_tensor(rho0, dtype=dtype, device=device, is_complex=True)
     if is_ket(rho0):
         rho0 = ket_to_dm(rho0)
     b_H = H_batched.size(0)
     rho0_batched = rho0[None, ...] if rho0.ndim == 2 else rho0
     rho0_batched = rho0_batched[None, ...].repeat(b_H, 1, 1, 1)  # (b_H, b_rho0, n, n)
 
     # convert t_save to a tensor
-    t_save = torch.as_tensor(t_save)
+    t_save = torch.as_tensor(t_save, device=device)
```
gautierronan, line 117: Also use `dtype=to_real(dtype)` here where `to_real` is imported as `from torch._prims_common import corresponding_real_dtype as to_real` so that all dtypes match.

torchqdynamics/operators.py
```diff
@@ -21,54 +22,45 @@
 ]
 
 
-def sigmax(
-    *, dtype: torch.dtype = torch.complex128, device: torch.device | None = None
-) -> Tensor:
+@complex_tensor
+def sigmax(*, dtype=None, device=None) -> Tensor:
```
gautierronan, line 26: Is it not needed to use typing anymore here? Because of the decorator?

torchqdynamics/sesolve/sesolve.py
```diff
@@ -37,21 +39,21 @@ def sesolve(
...
     # convert psi0 to a tensor and batch by default
     # TODO add test to check that psi0 has the correct shape
     b_H = H_batched.size(0)
-    psi0 = to_tensor(psi0)
+    psi0 = to_tensor(psi0, dtype=dtype, device=device, is_complex=True)
     psi0_batched = psi0[None, ...] if psi0.ndim == 2 else psi0
     psi0_batched = psi0_batched[None, ...].repeat(b_H, 1, 1, 1)  # (b_H, b_psi0, n, 1)
 
     # convert t_save to tensor
     t_save = torch.as_tensor(t_save)
```
gautierronan, line 53: Needs conversion as well to the device and dtype.

torchqdynamics/tensor_types.py
```diff
@@ -22,30 +27,99 @@
...
+) -> torch.complex64 | torch.complex128:
+    if dtype is None:
+        # the default dtype for complex tensors is determined by the default
+        # floating point dtype (torch.complex128 if default is torch.float64,
+        # otherwise torch.complex64)
+        if torch.get_default_dtype() is torch.float64:
+            return torch.complex128
+        else:
+            return torch.complex64
+    elif dtype not in (torch.complex64, torch.complex128):
```
gautierronan, line 81: If you use `to_tensor(x, dtype=torch.float32, is_complex=True)` this will raise an error. Is this the behavior we want? Or do we want to convert float32 to complex64 in this case?

gautierronan (changes requested):
Nice PR !

gautierronan:
>     * @gautierronan do we also need the `dtype`/`device` for the `none_to_zeros_like` function? I would say yes.

I don't think it is required. It's supposed to be an internal function only. But if you want to that's fine also.

torchqdynamics/operators.py
PierreGuilmin: It's never needed to use typing 😋 I think we can afford to be a bit loose with the typing here, because it's a bit technical for not much. Generally the two main advantages of typing the code is (1) for static type checking (e.g. with mypy) and (2) for readability (for users and developers). Here I don't think we need (1), and (2) is rather harmed than helped by the typing so I'd say it's better to remove it.

torchqdynamics/tensor_types.py
PierreGuilmin: Yes that's on purpose, I think one should always specify a complex type for a complex tensor, and we should avoid automatic conversions. But it can be discussed, typically `mesolve` will fail if someone passes `dtype=torch.float32`, he would have to pass `dtype=torch.complex64`, but I think it's better and more clear like that.

gautierronan:
> But the type/device of `dg = none_to_zeros_like(dg, parameters)` will not match the rest of the tensors then during the adjoint pass, maybe it's an issue no? I guess it would be better to have it, but I'm not sure which `dtype` and `device` to use (the one of `y`?).

I guess the issue comes from the `parameters` then. Maybe we should convert them to the appropriate dtype also at the beginning of `mesolve` ?

PierreGuilmin:
> Also use dtype=to_real(dtype) here where to_real is imported as from torch._prims_common import corresponding_real_dtype as to_real so that all dtypes match.

I'm not fan with the idea of using torch internals, I wrote a little one-liner which should work for all practical purpose.

PierreGuilmin:
> I guess the issue comes from the parameters then. Maybe we should convert them to the appropriate dtype also at the beginning of mesolve ?

I don't know haha, are the parameters all floats or can they be complex? I don't think we can convert them without lossing the binding to the real parameter 😕

gautierronan:
> > Also use dtype=to_real(dtype) here where to_real is imported as from torch._prims_common import corresponding_real_dtype as to_real so that all dtypes match.
> 
> I'm not fan with the idea of using torch internals, I wrote a little one-liner which should work for all practical purpose.

I wrote a PR to the PyTorch github here https://github.com/pytorch/pytorch/pull/97935 for exactly this. Once it's merged we can use it.

PierreGuilmin:
> No parameters can be both complex or real. But that doesn't mean they can't be converted to the appropriate precision. They actually should be converted, otherwise torch will do it behind the hood when calling torch.autograd.grad which will slow down everything significantly.

Alright, can I let you do this in another PR then? 😊

gautierronan (approved):

PR #70, opened 2023-04-21
Move save logic from `odeint` to `ForwardQSolver`

PierreGuilmin:
The logic to save state and expectation values at intermediate times is moved from `odeint` to the `ForwardQSolver` class. The goal of this PR is three-fold:
- First this makes the `odeint` code easier to read, and it fixes repeated saving code between fixed and adaptive schemes.
- This is a first step towards fixing the design issue which forces one to use `odeint` when implementing a solver. For example one could compute the exact propagator in `sesolve` or exponentiate the full Lindbladian directly in `mesolve`, without using a time-discretization, and would not need `odeint` at all. Now the saving is fully handled by the `ForwardQSolver` without needing to resort to `odeint` or copy the logic.
- This will allow the SME solvers to save an additional object (the measured signal) by just inheriting cleanly from the `ForwardQSolver`.

Note that the `save_states` boolean is now an argument to pass when defining the solver, rather than directly to `mesolve`.

@gautierronan I'm not 100% sure of the modifications for the adjoint code, could you take a careful look?


torchqdynamics/mesolve/adaptive.py
```diff
@@ -1,14 +1,13 @@
...
-from ..solver_options import SolverOption
 from ..solver_utils import kraus_map
 from ..tensor_types import TDOperator
 
 
 class MEAdaptive(ForwardQSolver):
-    def __init__(self, options: SolverOption, H: TDOperator, jump_ops: Tensor):
-        super().__init__(options)
+    def __init__(self, *args, H: TDOperator, jump_ops: Tensor):
+        super().__init__(*args)
```
gautierronan, line 10: Here and in all other ForwardQSolver children classes, I think it would be nice to have a more meaningful name than `*args` which implies any argument. Maybe `*qsolver_args` ? This way we know it is what is passed to the parent class.

torchqdynamics/mesolve/mesolve.py
```diff
@@ -131,36 +127,33 @@ def mesolve(
         solver = Dopri45()
 
     # define the QSolver
-    args = (solver, H_batched, jump_ops)
+    args = (solver, rho0_batched, exp_ops, t_save)
+    kwargs = dict(H=H_batched, jump_ops=jump_ops)
```
gautierronan, line 131: Similar comment here. It seems strange to separate `args` and `kwargs`, especially as the `kwargs` are not keyword arguments in the actual classes. I think you mean something like `generic_args` and `odeint_args`. These are not very good names but anything with a little bit more meaning would be nice I think !

torchqdynamics/odeint.py
```diff
@@ -16,8 +16,37 @@
...
+                *batch_sizes,
+                len(self.exp_ops),
+                len(self.t_save),
+                dtype=y0.dtype,
+                device=y0.device,
+            )
+        else:
+            self.exp_save = torch.empty(
+                *batch_sizes, len(self.exp_ops), dtype=y0.dtype, device=y0.device
+            )
```
gautierronan, line 49: Unless there's a specific reason for this, I think `exp_save` should be None if no `exp_ops` are passed.

torchqdynamics/odeint.py
```diff
@@ -211,40 +222,23 @@ def _adaptive_odeint(
 
             # save results if flag is raised
             if save_flag:
-                if save_states:
-                    y_save[..., save_counter, :, :] = y
-                if len(exp_ops) > 0:
-                    exp_save[..., save_counter] = bexpect(exp_ops, y)
-                save_counter += 1
+                qsolver.save(t, y)
```
gautierronan, line 225: Wrong function signature here, there is not `t` passed to `qsolver.save`.

torchqdynamics/solver_options.py
```diff
@@ -1,11 +1,19 @@
...
+        """
         self.verbose = verbose
+        self.save_states = save_states
 
 
 class FixedStep(SolverOption):
-    def __init__(self, *, dt: float, verbose: bool = True):
-        super().__init__(verbose=verbose)
+    def __init__(self, *, dt: float, **kwargs):
+        super().__init__(**kwargs)
```
gautierronan, line 16: I'm wondering whether we should be explicit with the arguments here? If people read the code, it's nice to see all the possible arguments to each SolverOption class directly in one place.

gautierronan (changes requested):
This is a very nice PR, great work! I really like the logic behind the changes. Just a few small comments. Concerning the adjoint, I think it's fine but would need to be checked. Will do so when I find the time to implement the tests.

torchqdynamics/mesolve/adaptive.py
PierreGuilmin: It's very canonical to name positional arguments `*args` and keyword-only arguments `**kwargs`, and even more canonical in inheritance-like schemes. From what I read, it's not recommended to pick different names, because everyone in the Python community uses this, so I would argue to keep it like that.

torchqdynamics/mesolve/mesolve.py
PierreGuilmin: Ho but any argument following `*args` is by default a keyword argument (not to be confused with an argument having a default value!). I would argue the same that the generic Python naming are actually the easiest to understand.

torchqdynamics/solver_options.py
PierreGuilmin: This is an endless debate in the community haha, I redirect you to this StackOverlow post which I quite agree with > https://stackoverflow.com/questions/73997582/should-i-repeat-parent-class-init-arguments-in-the-child-classs-init-o. I think it actually makes the code easier to read in these inheritance schemes, and refactoring (renaming one of the base class argument or adding one) so much easier.

torchqdynamics/mesolve/mesolve.py
gautierronan: But I don't get why you would put `args` and `kwargs` on a different level here. They're both arguments of the qsolver classes. You might as well write `args = (solver, rho0_batched, exp_ops, t_save, H_batched, jump_ops)` no?

gautierronan (approved):

torchqdynamics/mesolve/mesolve.py
PierreGuilmin: No it wouldn't work, you can't have a positional argument following `*args`:
```python
>>> def foo(*args, a):
....    pass
>>> foo(1, 2, 3, 4)
TypeError: foo() missing 1 required keyword-only argument: 'a'
```

gautierronan: Yes but it's because you changed the function signature from 
```python
def __init__(self, options: SolverOption, H: TDOperator, jump_ops: Tensor)
```
to 
```python
def __init__(self, *args, H: TDOperator, jump_ops: Tensor)
``` 
But ok I think you're right I'm just mixing myself up. It doesn't really matter anyway ^^

PierreGuilmin: Ha yes yes sorry, it's clearly because of the change 😅

abocquet (approved):

PR #71, opened 2023-04-21
Make solvers independent from `odeint`

PierreGuilmin:
This PR is built on top of #70.

This PR is a fairly major change to the structure that I had in mind for a long time. It basically decouples `sesolve` and `mesolve` from `odeint` and allows to support any kind of solvers. Use cases are _e.g._ explicit exponentiation of the Hamiltonian for `sesolve` (see last commit of this PR), explicit exponentiation of the Lindbladian for `mesolve`, using another PyTorch ODE solver library than our implementation, using analytic formuale for the gradient _à la GRAPE_, implementing some new ideas (funky trajectory sampling to solve the SME, other ways to compute the gradient, etc...). It also helps splitting the code in appropriate separated chuncks. It may feel like a big change, but it's just reorganization, there's nothing new.

Here's the new architecture:
- we have a base class called `QSolver` with a `run()` method called by `sesolve`/`mesolve`
- the implementation of `run()` is responsible for calling the base `QSolver` class `save()` method to save the state and expectation values at the correct time, that's all

Any quantum solver inherits from this base class, typically we have `ODEForwardQSolver` which is the base class for any ODE solver implementing only the forward pas. If you want to see a simple example look at the last commit, which solves Schrödinger by directly computing the propagator, and does not need any `odeint`-related logic. 😉 

Left to do:
- [x] Fix `ODEAdjointQSolver` > I think we can merge some functions directly in the qsolver class, and get rid of parameters that are now class attributes, but I'd prefer that you do it @gautierronan because you know the code much better. If you can commit on top of this PR by just modifying this part it would be great 😁
- [ ] Remove `parameters` from `QSolver` base class > technically `parameters` is a specific argument required only for the ODE solvers implementing gradient by the adjoint, it shouldn't be in the base class.

### DIFF
--- a/torchqdynamics/ode_adjoint_qsolver.py
+++ b/torchqdynamics/ode_adjoint_qsolver.py
@@ -17,9 +17,9 @@ class ODEAdjointQSolver(ODEForwardQSolver):
     GRADIENT_ALG = ['autograd', 'adjoint']
 
     def run(self):
-        super().run()
-
-        if self.gradient_alg == 'adjoint':
+        if self.gradient_alg in [None, 'autograd']:
+            super().run()
+        elif self.gradient_alg == 'adjoint':
             self._odeint_adjoint()
 
     @abstractmethod
--- a/torchqdynamics/ode_adjoint_qsolver.py
+++ b/torchqdynamics/ode_adjoint_qsolver.py
@@ -7,173 +7,13 @@ import torch.nn as nn
 from torch import Tensor
 from torch.autograd.function import FunctionCtx
 
-from .adaptive import DormandPrince45
+from .ode_forward_qsolver import ODEForwardQSolver
 from .progress_bar import tqdm
-from .qsolver import QSolver
-from .solver_options import Dopri45, ODEAdaptiveStep, ODEFixedStep
+from .solver_options import ODEAdaptiveStep, ODEFixedStep
 from .solver_utils import add_tuples, none_to_zeros_like
 
 
-class ForwardQSolver(QSolver):
-    @abstractmethod
-    def forward(self, t: float, y: Tensor) -> Tensor:
-        """Iterate the quantum state forward.
-
-        Args:
-            t: Time.
-            y: Quantum state, of shape `(..., m, n)`.
-
-        Returns:
-            Tensor of shape `(..., m, n)`.
-        """
-        pass
-
-    def run(self):
-        """Integrate a quantum ODE starting from an initial state.
-
-        The ODE is solved from time `t=0.0` to `t=t_save[-1]`.
-        """
-        # check arguments
-        check_t_save(self.t_save)
-
-        # dispatch to appropriate odeint subroutine
-        if self.gradient_alg is None:
-            self._odeint_inplace()
-        elif self.gradient_alg == 'autograd':
-            self._odeint_main()
-
-    def _odeint_inplace(self):
-        """Integrate a quantum ODE with an in-place solver.
-
-        Simple solution for now so torch does not store gradients.
-        TODO Implement a genuine in-place solver.
-        """
-        with torch.no_grad():
-            self._odeint_main()
-
-    def _odeint_main(self):
-        """Dispatch the ODE integration to fixed or adaptive time step subroutines."""
-        if isinstance(self.options, ODEFixedStep):
-            self._fixed_odeint()
-        elif isinstance(self.options, ODEAdaptiveStep):
-            self._adaptive_odeint()
-
-    def _fixed_odeint(self):
-        """Integrate a quantum ODE with a fixed time step solver.
-
-        Note:
-            The solver times are defined using `torch.linspace` which ensures that the
-            overall solution is evolved from the user-defined time (up to an error of
-            `rtol=1e-5`). However, this may induce a small mismatch between the time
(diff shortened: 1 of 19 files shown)


torchqdynamics/qsolver.py
```diff
@@ -0,0 +1,118 @@
...
+            # exp_save: (..., len(exp_ops), len(t_save))
+            self.exp_save = torch.zeros(
+                *batch_sizes,
+                len(self.exp_ops),
+                len(self.t_save),
+                dtype=self.y0.dtype,
+                device=self.y0.device,
+            )
+        else:
+            self.exp_save = torch.empty(
```
gautierronan, line 89: Careful upon merging with the parent branch here that this is converted to None.

torchqdynamics/ode_forward_qsolver.py
```diff
@@ -0,0 +1,167 @@
...
+            # save solution
+            if t >= self.next_tsave():
+                self.save(y)
+
+            # iterate solution
+            y = self.forward(t, y)
+
+        # save final time step (`t` goes `0.0` to `t_save[-1]` excluded)
+        self.save(y)
+        self.save_final(y)
```
gautierronan, line 91: Kind of strange to have `self.save(y)` twice here. It looks like you're saving the state twice. IMO a better logic would be to have only `self.save_final(y)` and within the corresponding function, to check for the `self.options.save_states` boolean.

torchqdynamics/ode_forward_qsolver.py
```diff
@@ -0,0 +1,167 @@
...
+            # update results if step is accepted
+            if error <= 1:
+                t, y, ft = t + dt, y_new, ft_new
+
+                # update the progress bar
+                pbar.update(dt.item())
+
+                # save results if flag is raised
+                if save_flag:
+                    self.save(t, y)
```
gautierronan, line 154: Same comment here: careful when merging since this is in a new file.

torchqdynamics/ode_adjoint_qsolver.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class ODEAdjointQSolver(ODEForwardQSolver):
+    GRADIENT_ALG = ['autograd', 'adjoint']
+
+    def run(self):
+        super().run()
+
+        if self.gradient_alg == 'adjoint':
+            self._odeint_adjoint()
```
gautierronan, line 23: The logic is kinda difficult to understand here. It would be better to copy the code again and to have
```python
if self.gradient_alg is None:
    self._odeint_inplace()
elif self.gradient_alg == 'autograd':
    self._odeint_main()
elif self.gradient_alg == 'adjoint':
    self._odeint_adjoint()
else:
    raise ValueError()
```
I added the error raising but maybe it's already somewhere else I'm not sure.

torchqdynamics/sesolve/exponentiate.py
```diff
@@ -0,0 +1,18 @@
+import torch
+
+from ..progress_bar import tqdm
+from ..qsolver import QSolver
+
+
+class SEExponentiate(QSolver):
```
gautierronan, line 7: I think a better name would be `SEPropagator`. It is more explicit on what the solver is actually doing.

gautierronan (changes requested):
Again some very nice work ! It's a very useful reorganization, the logic is better overall and will be easier to build upon. A few small comments all around and I'll be implementing the adjoint.

torchqdynamics/ode_adjoint_qsolver.py
gautierronan: Or alternatively to have
```python
if self.gradient_alg in [None, 'autograd']:
    super().run()
elif self.gradient_alg == 'adjoint':
    self._odeint_adjoint()
```

gautierronan:
> Fix `ODEAdjointQSolver` > I think we can merge some functions directly in the qsolver class, and get rid of parameters that are now class attributes, but I'd prefer that you do it @gautierronan because you know the code much better. If you can commit on top of this PR by just modifying this part it would be great 😁

I've made the changes. I tried to keep the same logic, but I couldn't do it all the way. Typically, the `ODEIntAdjoint.forward` method *has* to return `y_save` in order to have access to `grad_y` in the backward pass (this is how torch.autograd.Function works). And also, it has to take as arguments at least y0 and the parameters, for similar reasons.

The rest should be fine, but check that you agree with the logic.

abocquet (approved):

PierreGuilmin:
Sorry for the force rebase, it was easier on my side with all these files renaming, the only novelties are the last five commits for the merge and to fix @gautierronan comments.

> I've made the changes. I tried to keep the same logic, but I couldn't do it all the way. Typically, the ODEIntAdjoint.forward method has to return y_save in order to have access to grad_y in the backward pass (this is how torch.autograd.Function works). And also, it has to take as arguments at least y0 and the parameters, for similar reasons.

Yep I expected that you would not be able to use class attributes all the way but that's fine! I didn't check thoroughly but I think it's good, I still consider this part of the code as WIP anyway so it's OK if it's still a little bugged as long as the overall logic works well with the proposed framework.

>  Remove parameters from QSolver base class > technically parameters is a specific argument required only for the ODE solvers implementing gradient by the adjoint, it shouldn't be in the base class.

I'll do this in another PR as this one is getting quite big already.

PR #73, opened 2023-04-24
Small renamings

pierreguilmin:
The namings between `QSolver`, `SolverOption` (referred externally as `solver`) and `AdaptiveSolver` was a bit confusing, we now have three objects:
- `QSolver`: implements a quantum solver.
- `AdaptiveODESolver`: class which has nothing to do with a `QSolver` used by the adaptive ODE `QSolver`.
- `SolverOption`: internal name for `QSolver` options, external exposed name is still `solver` (in `sesolve` and `mesolve` APIS and for imports).


abocquet (dismissed):

gautierronan (approved):

gautierronan:
> There we go 🤓 Now, what do you think about renaming `QSolver` to simply `Solver`? (@abocquet and @gautierronan) We always refer to a `QSolver` object as the _solver_ in the comment and in the doc, and I don't think it's necessary to precise it's a _quantum solver_ (if this even has a meaning), as long as we give a clear name to the other kind of solvers (`ode_solver` for ODE solvers for instance).

Haha are you making proposals incrementally so we are more likely to agree? :D
In any case I'm fine with this renaming, it doesn't really matter for end users anyway !

pierreguilmin:
> Haha are you making proposals incrementally so we are more likely to agree? :D

Hahaha no it's just that I think about them too late 😅

PR #75, opened 2023-04-24
Some more renaming and reorganization

gautierronan:
Following up on https://github.com/pierreguilmin/torchqdynamics/pull/73, this PR proposes a few renamings and mostly a folder reorganization.

- f69a992cfc6c5b8a76c92de3ddf90feda7582ee8 and 01d568bec411092ee2d1c57b8fcc986ff99c5984: Rename test files and test classes for consistency with the library.
- 78bc0c8a4d6a2236e9f27ae6d3d961396e966f2f: Separate files into three new folders: utils (all utility functions), smesolve, and ode (all ode integration functions). I kept `options.py` and `solver.py` in the main folder.
- 9b673facf1237aec5dd586c9f45b5a64ddeaf128: Rename the files to remove "ode" (since it's now in the folder of the same name), and renamed the `AdaptiveODESolver` class to `AdapativeODEIntegrator` to differentiate it from the actual `Solver` class.


torchqdynamics/ode/forward_solver.py
pierreguilmin, line 1: The file name should match the class name, I think we can use `ForwardSolver` and `forward_solver.py` because they are in the `ode` namespace so it's clear enough 👍

torchqdynamics/ode/adaptive_integrator.py
```diff
@@ -6,11 +6,11 @@
...
 
-from .solver_utils import hairer_norm
-from .tensor_types import dtype_complex_to_float
+from ..utils.solver_utils import hairer_norm
+from ..utils.tensor_types import dtype_complex_to_float
 
 
-class AdaptiveODESolver(ABC):
+class AdaptiveODEIntegrator(ABC):
     """A parent class for all adaptive time step ODE solvers.
```
pierreguilmin, line 14:
```suggestion
    """A parent class for all adaptive time step ODE integrators.
```

torchqdynamics/ode/adjoint_solver.py
pierreguilmin, line 1: Let's consistently use the term `integrate` for the ODE solving, line 138
```diff
- # solve the ODE forward without storing the graph of operations
+ # integrate the ODE forward without storing the graph of operations
```

pierreguilmin (changes requested):
Very nice!! Much cleaner ✨

tests/mesolve/test_mesolve.py
pierreguilmin, line 1: I suggest renaming back to `me_solver_test.py` and `MESolverTest`:
- The file name conflicts with the future test file for the real `mesolve.py` file (testing the API, the errors, the arguments conversion, etc...).
- Pytest collects by default all files beginning by `test_...` and all classes beginning by `Test...`, so when we write a base class for some tests (like this one) or some helpers for other tests (like `open_system.py`), it's better to use a different naming. This naming convention also makes very clear for a reader what are the test files and what are the files used to build the tests that are not themselves testing anything. Typically we will have in the `tests/` directory some random files (`utils.py`, `open_system.py`, `me_solver_test.py`, ...) and some files beginning by `test_` and suffixed with a file name from the package (`test_euler.py`, test_mesolve.py`, etc...).

gautierronan: I get your point, but I then I think `MESolverTest` sounds like a test whereas it is a base class for actual test classes. Anyway, I renamed to `MESolverTester` in this case, which I think it a bit more explicit.

PR #83, opened 2023-04-29
Add Wigner utility functions

gautierronan:
Adds the equivalent of `qutip.wigner` and `qutip.plot_wigner`.

### DIFF
--- a/torchqdynamics/utils/states.py
+++ b/torchqdynamics/utils/states.py
@@ -92,7 +92,9 @@ def coherent(dim: int, alpha: complex | Tensor, *, dtype=None, device=None) -> T
                 [0.003+0.j],
                 [0.000+0.j]], dtype=torch.complex128)
     """
-    return displace(dim, alpha, dtype=dtype, device=device) @ fock(dim, 0)
+    return displace(dim, alpha, dtype=dtype, device=device) @ fock(
+        dim, 0, dtype=dtype, device=device
+    )
 
 
 @complex_tensor
--- a/torchqdynamics/utils/tensor_types.py
+++ b/torchqdynamics/utils/tensor_types.py
@@ -86,6 +86,19 @@ def cdtype(
     return dtype
 
 
+def rdtype(
+    dtype: torch.float32 | torch.float64 | None = None,
+) -> torch.float32 | torch.float64:
+    if dtype is None:
+        return torch.get_default_dtype()
+    elif dtype not in (torch.float32, torch.float64):
+        raise TypeError(
+            f'Argument `dtype` ({dtype}) must be `torch.float32`,'
+            ' `torch.float64` or `None` for a real-valued tensor.'
+        )
+    return dtype
+
+
 def complex_tensor(func):
     @functools.wraps(func)
     def wrapper(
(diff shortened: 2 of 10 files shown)


torchqdynamics/utils/states.py
```diff
@@ -92,7 +92,9 @@ def coherent(dim: int, alpha: complex | Tensor, *, dtype=None, device=None) -> T
                 [0.003+0.j],
                 [0.000+0.j]], dtype=torch.complex128)
     """
-    return displace(dim, alpha, dtype=dtype, device=device) @ fock(dim, 0)
+    return displace(dim, alpha, dtype=dtype, device=device) @ fock(
```
abocquet, line 95: I'm guessing that implementing direclty the series expansion of a coherent state would be faster here. Is there a reason why we choose to apply the displacement operator to vacuum (other than being convenient)? 
If not maybe it's  an easy improvement that we could do sometimes

gautierronan: From QuTiP's documentation:

> Select method 'operator' (default) or 'analytic'. With the 'operator' method, the coherent state is generated by displacing the vacuum state using the displacement operator defined in the truncated Hilbert space of size 'N'. This method guarantees that the resulting state is normalized. With 'analytic' method the coherent state is generated using the analytical formula for the coherent state coefficients in the Fock basis. This method does not guarantee that the state is normalized if truncated to a small number of Fock states, but would in that case give more accurate coefficients.

So there isn't a method that is always better than the other, but we could give the choice like qutip does.

abocquet (dismissed):

torchqdynamics/utils/tensor_types.py
```diff
@@ -99,11 +99,25 @@ def wrapper(
...
 ) -> torch.float32 | torch.float64:
     dtype = cdtype(dtype)
-    return torch.float64 if dtype is torch.complex128 else torch.float32
+    return DTYPE_CONVERSION_TABLE[dtype]
+
+
+def dtype_float_to_complex(
+    dtype: torch.float32 | torch.float64 | None = None,
+) -> torch.complex64 | torch.complex128:
+    return DTYPE_CONVERSION_TABLE[dtype]
```
pierreguilmin, line 120: Fails if `None` is passed.

pierreguilmin (changes requested):
I don't have much time this week, so I'm not reviewing the content of `wigner.py`, I think it's quite independent of the rest of the library, so it's fine if we iterate on it later. One small important comment before merging (and removing `plot_wigner` as discussed).

torchqdynamics/utils/tensor_types.py
gautierronan: It's fixed. I also updated the name from `dtype_float_to_complex` to `dtype_real_to_complex` everywhere, to match the naming convention of pytorch.

torchqdynamics/utils/tensor_types.py
```diff
@@ -99,11 +112,32 @@ def wrapper(
...
 
 
-def dtype_complex_to_float(
+DTYPE_TO_REAL = {
+    torch.complex64: torch.float32,
+    torch.complex128: torch.float64,
+    torch.float32: torch.float32,
+    torch.float64: torch.float64,
+}
+DTYPE_TO_CPLX = {
```
pierreguilmin, line 121:
```suggestion
DTYPE_TO_COMPLEX = {
```

torchqdynamics/utils/tensor_types.py
```diff
@@ -99,11 +112,32 @@ def wrapper(
     return wrapper
 
 
-def dtype_complex_to_float(
+DTYPE_TO_REAL = {
+    torch.complex64: torch.float32,
+    torch.complex128: torch.float64,
+    torch.float32: torch.float32,
+    torch.float64: torch.float64,
+}
```
pierreguilmin, line 120:
```suggestion
DTYPE_TO_REAL = {
    torch.complex64: torch.float32,
    torch.complex128: torch.float64,
}
```

torchqdynamics/utils/tensor_types.py
```diff
@@ -99,11 +112,32 @@ def wrapper(
...
+DTYPE_TO_CPLX = {
+    torch.float32: torch.complex64,
+    torch.float64: torch.complex128,
+    torch.complex64: torch.complex64,
+    torch.complex128: torch.complex128,
+}
```
pierreguilmin, line 126:
```suggestion
    torch.float32: torch.complex64,
    torch.float64: torch.complex128,
}
```

PR #86, opened 2023-05-04
Refactor the `Solver` class and all subclasses

gautierronan:
This is quite a big PR that completely reorganizes the `Solver` class. As I was discussing with @pierreguilmin, the current structure did not allow for certain new solvers such as fixed time step Runge Kutta solvers. It was also somewhat incoherent between functions.

I'm quite happy with this proposal, but I'm open to criticism of course. I'd suggest reviewing commit by commit to not have to read too much at once.

In short, the structure goes like:
```
Solver
  >> Propagator

  >> ForwardSolver
     >> Rouchon <with adjoint support>
        >> MERouchon
           >> MERouchon1
           >> MERouchon1_5
           >> MERouchon2
     >> Euler
        >> MEEuler
        >> SEEuler

  >> AdaptiveSolver
     >> MEAdaptive
     >> SEAdaptive

  >> FixedSolver
     >> MEFixed
     >> SEFixed

Integrator
  ** AdaptiveIntegrator
     ** Dopri5
     ** Dopri8 (TBD)
  ** FixedIntegrator
     ** RK4 (TBD)
```
So there is no more `AdjointSolver` class, but the adjoint is implemented directly within each given subclass. For now there is only Rouchon support, but it could easily be extended to Euler as well, and even to Adaptive and Fixed Solvers.

Also, the `ForwardSolver` class now represents any solver that uses a `forward` function that computes directly the next y(t+dt). This is not the case of `AdaptiveSolver` and `FixedSolver` which compute the derivative dy/dt. So I renamed these functions `odefun` instead for coherence.

In following PRs:
- [x] Add typing all around
- [ ] Dopri8
- [ ] RK4
- [ ] Add adjoint to Euler


abocquet:
Very nice PR! It may be because I'm not up to speed in numerical algorithms, but I'm not sure to follow how the distinction is made between
```
  - Propagator
  - ForwardSolver
  - AdaptiveSolver
  - AdaptiveIntegrator
  - FixedSolver
  - FixedIntegrator
```
For instance why a `RouchonSolver` should be a `ForwardSolver` but not `RK4`?

gautierronan:
> 



> Very nice PR! It may be because I'm not up to speed in numerical algorithms, but I'm not sure to follow how the distinction is made between
> 
> ```
>   - Propagator
>   - ForwardSolver
>   - AdaptiveSolver
>   - AdaptiveIntegrator
>   - FixedSolver
>   - FixedIntegrator
> ```
> 
> For instance why a `RouchonSolver` should be a `ForwardSolver` but not `RK4`?

So the problem is that all these methods have their specificities, and they might differ in different places/functions. So it's very hard to give a proper "tree-like" structure simply with classes and sub-classes that encapuslates all these methods.

To be more specific, adaptive methods like Dopri5 and Dopri8 apply to both the SE and ME (and possibly others). So if I were to write solvers MEDopri5, SEDopri5, MEDopri8 and SEDopri8, there would be alot of redundance between MEDopri5 and MEDopri8 (it's the same function we integrate, i.e. the Lindblad equation), but also between MEDopri5 and SEDopri5 (they share the same integration method, same Butcher tableau, etc.). So I decided to separate the Integrator class from the Solver class for these specific methods for instance. Note that `AdaptiveIntegrator` and `FixedIntegrator` are not sub-classes of `Solver` here.

To answer your question, `Rouchon` and `RK4` do share some structure in the sense that we have a simple `for` loop that iterates over time. However, the `forward` functions that compute the next y(t+dt) are very different. For `Rouchon`, we have an explicit formula that applies only to the ME, and that is very "quantum" by nature. For `RK4`, we have an integrator that calls an function that returns dy/dt at all times, and computes the next y(t+dt) from a runge-kutta method. So this is some machinery, and it should still be separated into `MERK4` and `SERK4` depending on the actual ode we integrate.

So in short, `ForwardSolver` is a parent class for all methods that feature a `forward` method that can directly compute y(t+dt) in a "homemade" manner. While the `FixedSolver` is a prent class for all methods based on ODE integration.

pierreguilmin (changes requested):
Nice PR! I think we need to discuss reworking it a bit, but it goes in the right direction.

First, I'd like to state explicitly the philosophy I was trying to impulse with the previous reorganization:
- The code should work for **any kind of solver** (even one having nothing to do with integrating an ODE).
- When implementing a solver, it is the solver which defines how the gradient is computed. Typically, the base solver should only support no gradient and naive autograd. Then specific solvers might have specific ways to compute the gradient, e.g. gradient by the adjoint for ODE-related solvers, analytic gradient for GRAPE, etc...

Now three important points:
1. The naming of ODE solvers is very significant now that they lost their namespace (`ForwardSolver` is an `ODEForwardSolver`, `AdaptiveSolver` is an `ODEAdaptiveSolver` and also `AdaptiveIntegrator` is an `AdaptiveODEIntegrator`). I propose creating an `ode/` subfolder in `solver` gathering `ForwardSolver`, `AdaptiveSolver` and `AdaptiveODEIntegrator` (no need to create a subfolder for this guy). Then the imports will feature the correct namespace, and we don't need to have lengthy class names.
2. There is a design flaw here with the gradient by the adjoint being hooked up so high in the base class hierarchy, we should have an `AdjointGradSolver(Solver)` class that fixes the interface for ODE solvers that implement gradient by the adjoint. Then we can put the `adjoint_solver` file directly in the `ode/` solvers directory, because it is specific to these solvers.
3. So the `Rouchon` solver would be `Rouchon(ForwardSolver, AdjointSolver)`. Concerning the adjoint code now implemented in `solvers/rouchon.py`, is it really specific to the Rouchon solver? Won't the Euler method use the same structure for the adjoint? Then it should go into the `AdjointSolver` class.

pierreguilmin (approved):
Looks good, we can always improve on this later. As discussed together, I'll propose PRs for:
- Moving the gradient algorithm choice (`gradient_alg`) and needed arguments (like `parameters` for the adjoint) to solver `Options`.
- Removing `autograd` from the base solver class to support non-differentiable solvers.

The new adaptive code looks good 👌

PR #87, opened 2023-05-04
Time dependent Hamiltonian | Part 1

abocquet:
Replaces https://github.com/pierreguilmin/torchqdynamics/pull/58. The philosophy is very close but the decorators logics is much lighter, as the only accepted variable is now time.

Only MERouchon1 takes advantage of the caching for now, as an example. Others will come in another PR, but are compatible with callable H

I'm not very satisfied with the way the duality Tensor/callable/list of callables is handled right now. Any idea is welcomed

For next PRs:

- [ ] piecewise constant H
- [ ] cache all numerical schemas


torchqdynamics/sesolve/propagator.py
```diff
@@ -8,11 +8,9 @@ class SEPropagator(Solver):
     def __init__(self, *args):
         super().__init__(*args)
 
-        self.H = self.H[:, None, ...]  # (b_H, 1, n, n)
-
     def run(self):
         y, t1 = self.y0, 0.0
         for t2 in tqdm(self.t_save, disable=not self.options.verbose):
-            y = torch.matrix_exp(-1j * self.H * (t2 - t1)) @ y
+            y = torch.matrix_exp(-1j * self.H(t2) * (t2 - t1)) @ y
```
gautierronan, line 14: A priori the `propogator` method is always time independent. We will have to check this somewhere I guess, but it can be a separate PR.

gautierronan (approved):
I think we can merge this. And then we'll iterate in subsequent PRs.

tests/odeint/test_h_caching.py
pierreguilmin, line 1: This file should be in `tests/` directly.

pierreguilmin (changes requested):
Minor modifications, then OK to merge 🙌

gautierronan (approved):

pierreguilmin (approved):
Lezgooo 🥳

PR #89, opened 2023-05-04
Handle batching for time-dependent Hamiltonians

gautierronan:
And a few other small things.


torchqdynamics/solvers/solver.py
```diff
@@ -24,9 +24,12 @@ def H_squared(self, t):
     @functools.wraps(func)
     def wrapper(instance, t, *args, **kwargs):
         if func.__name__ not in instance._cache or (
-            t != instance._cache[func.__name__][0] and instance._H_changed(t)
+            (t, args, kwargs) != instance._cache[func.__name__][0]
```
abocquet, line 27: There is a topic here, because args and kwargs can be different in value but have `old_args == args` if args contain pointer to objects that have been modified. IMO this is very misleading can entail non trivial bugs.
To solve it, two solutions : 
- perform a deep comparison of the arguments (cf all the machinery in https://github.com/pierreguilmin/torchqdynamics/blob/abocquet/time-dependent-hamiltonian-2/torchqdynamics/decorators.py#L50)
- put the dust under the rug and only support 't'. This sounds reasonable as I don't see a usecase for calling twice a H dependant function with the same time but different arguments, but maybe I'm missing something

abocquet (commented):
Great work 🚀 thanks for doing this
I see you won't to merge into `[solver-structure-v2](https://github.com/pierreguilmin/torchqdynamics/tree/solver-structure-v2)`, why not `main` direclty

torchqdynamics/solvers/solver.py
pierreguilmin: I agree that this modification may lead to very subtle bugs. I would favour the second cheap proposed solution (only care about `t`) as discussed at the time.

pierreguilmin (changes requested):
I think we need to move some logic in `tensor_types.py` because it's better abstracted there. Ideally when we have a `TDOperatorLike` object somewhere in the code we should just:
- use a function to convert it to a `TDOperator` object (`H = to_td_op(H)`)
- always use high-level functions on this `TDOperator` object (e.g. get the value at some time with `H(t)`, batch with `H_batched = td_op_batch(H, 1)`, get the size with `b_H = td_op_size(H, 0)`, etc...)

torchqdynamics/utils/tensor_types.py
```diff
@@ -24,7 +24,7 @@
 OperatorLike = Union[TensorLike, Qobj]
 
 # type for objects convertible to a `TDOperator` using `time_dependent_to_tensor`
-TDOperatorLike = Union[OperatorLike, Callable[[float], OperatorLike]]
+TDOperatorLike = Union[OperatorLike, Callable[[float], Tensor]]
```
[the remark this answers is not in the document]
gautierronan: > 



> I think we should support the previous typing. It's very handy e.g. for people defining a time-dependent Hamiltonian using QuTiP operators.

I don't think we should allow this. The problem is that, if people pass time-dependent qutip objects, we will have to convert them at every time step. It can significantly slow down the code, especially for device conversion, but even type or dtype conversion is not a good habit.

Hence I would be in favor of not allowing it.

torchqdynamics/solvers/solver.py
gautierronan: Ok that's fine with me. I will add a warning in the docstring then.

torchqdynamics/utils/tensor_types.py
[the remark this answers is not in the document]
pierreguilmin: > It can significantly slow down the code

I disagree with the _significantly_, I think it's a minor time overhead compared to our gigantic matrix multiplications. I **do agree** that we should raise a warning and have the doc clearly say that it's better to have proper torch tensors for time-dependent Hamiltonians, but for convenience to test things quickly and all, it's quite nice to support qutip objects or numpy arrays. But I'm fine if you really think we should strongly forbid it.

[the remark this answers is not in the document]
gautierronan: No a CPU/GPU transfer of a big matrix is much worse than a single matrix multiplication on GPU. So definitly it can be a significant slowdown. 

However, there are some qutip types that we could allow, such as the `list[list[qobj, callable[[float], complex]]` type. But maybe this should be another PR.

Here is the list of all time-dependent formats supported by qutip: https://qutip.org/docs/latest/guide/dynamics/dynamics-time.html. We can decide on which ones we want to keep or not.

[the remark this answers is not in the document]
pierreguilmin: Ok my bad then, let's change to forbid non-genuine torch tensors.

gautierronan:
I think what you're describing is essentially a proper class for `TDOperator` (which I renamed to `TDTensor` since it only allows tensors). So we would convert `H` to this class in the beginning, and then the class would feature some proper methods like `__call__(self, t)` or `size(self, i)` and such.

If you agree with this, I would suggest doing it in another PR.

torchqdynamics/utils/tensor_types.py
```diff
@@ -67,6 +66,57 @@ def to_tensor(
...
+        x = to_tensor(x, dtype=dtype, device=device, is_complex=True)
+    elif isinstance(x, callable):
+        # check number of arguments and compute at initial time
+        try:
+            x0 = x(0.0)
+        except TypeError:
+            raise TypeError(
+                'Time-dependent operators in the `callable` format should only accept a'
+                ' single argument for time `t`.'
+            )
```
pierreguilmin, line 89:
```suggestion
        except TypeError as e:
            raise TypeError(
                'Time-dependent operators in the `callable` format should only accept a'
                ' single argument for time `t`.'
            ) from e
```
Nitpicking, this is called exception chaining and is a best practice.

torchqdynamics/utils/tensor_types.py
```diff
@@ -67,6 +66,57 @@ def to_tensor(
...
+                f' operator is currently of dtype {x0.dtype} instead of {dtype}.'
+            )
+        elif x0.device != device:
+            raise TypeError(
+                'Time-dependent operators in the `callable` format should always'
+                ' return a `torch.Tensor` with the same dtype and device as provided'
+                ' to the solver. This avoids type conversion or device transfer at'
+                ' every time step that would slow down the solver. The provided'
+                f' operator is currently on device {x0.device} instead of {device}.'
+            )
```
pierreguilmin, line 115: Let's factor this
```suggestion
        prefix = 'Time-dependent operators in the `callable` format should always'
                ' return a `torch.Tensor` with the same dtype and device as provided'
                ' to the solver. This avoids type conversion or device transfer at'
                ' every time step that would slow down the solver.'
                
        
        if not isinstance(x0, Tensor):
            raise TypeError(f'{prefix} The provided operator is currently of type {type(x0)} instead of {Tensor}.')
        elif x0.dtype != dtype:
            raise TypeError(f'{prefix} The provided operator is currently of dtype {x0.dtype} instead of {dtype}.')
        elif x0.device != device:
            raise TypeError(f'{prefix} The provided operator is currently on device {x0.device} instead of {device}.')
```

torchqdynamics/utils/tensor_types.py
```diff
@@ -67,6 +66,57 @@ def to_tensor(
         )
 
 
+def to_tdtensor(
+    x: TDOperatorLike,
+    dtype: torch.dtype | None = None,
+    device: torch.device | None = None,
+    is_complex: bool = False,
+):
```
pierreguilmin, line 74:
```suggestion
) -> TDTensor:
```

pierreguilmin (changes requested):
Minor comments.

pierreguilmin:
> I think what you're describing is essentially a proper class for TDOperator (which I renamed to TDTensor since it only allows tensors). So we would convert H to this class in the beginning, and then the class would feature some proper methods like __call__(self, t) or size(self, i) and such.
> 
> If you agree with this, I would suggest doing it in another PR.

I am a bit scared of performance issues and future troubles with `torch.compile` when using custom classes for tensors, I think it's fine to have a function-based object because we only use two functions, we don't need a proper subclassing.

gautierronan:
> I am a bit scared of performance issues and future troubles with `torch.compile` when using custom classes for tensors, I think it's fine to have a function-based object because we only use two or three functions, we don't need a proper subclassing.

Well such a sub-class could also handle the call to 
```python
H_batched = lambda t: H(t)[:, None, ...]  # (b_H, 1, n, n)
```
and the
```python
    @depends_on_H
    def H(self, t: float) -> Tensor:
        if isinstance(self._H, Tensor):
            return self._H
        elif callable(self._H):
            return self._H(t)
        else:
            raise TypeError('Piecewise constant Hamiltonian not supported yet.')

    def _H_changed(self, _t: float) -> bool:
        if isinstance(self._H, Tensor):
            return False
        elif callable(self._H):
            return True
        else:
            raise TypeError('Piecewise constant Hamiltonian not supported yet.')
```
that is currently within `Solver`.

I think we're doing exactly the same thing, whether it is in a separate class or not, it doesn't matter. It's just more clean. What you're scared of is sub-classing `Tensor`, which is not what we would be doing.

torchqdynamics/utils/tensor_types.py
gautierronan: Ok that's nice. Didn't know about this one.

gautierronan:
> Ok, let's try to subclass it in another PR then 👌

I still made an implementation like you were suggesting in the end (see latest commit). But you'll see that there's a bunch of `tdtensor` functions, which really screams for a proper class.

pierreguilmin (approved):
> But you'll see that there's a bunch of tdtensor functions, which really screams for a proper class.

Yes I 100% agree now 🤓

PR #91, opened 2023-05-05
Fix tqdm warning for adaptive solvers

pierreguilmin:


torchqdynamics/solvers/ode/adaptive_solver.py
```diff
@@ -67,9 +67,7 @@ def run_autograd(self):
                     t, y, ft = t + dt, y_new, ft_new
 
                     # update the progress bar
-                    with warnings.catch_warnings():  # ignore tqdm precision overflow
-                        warnings.simplefilter('ignore', TqdmWarning)
-                        pbar.update(dt.item())
```
gautierronan, line 72: Are you sure this one is not issuing warnings anymore? At the time it was for me.

gautierronan (approved):

torchqdynamics/solvers/ode/adaptive_solver.py
pierreguilmin: It does not for the current tests set. Maybe we need both, let's merge this and see if we need this one too 🤷‍♂️

PR #93, opened 2023-05-05
Move gradient algorithm parametrisation to solver options

pierreguilmin:
The gradient algorithm is now a solver option (which makes sense because the supported ways to compute the gradients depend on the solver). This also allow to move to the solver options the gradient algorithm-specific arguments (like `parameters` when the gradient algorithm is the adjoint).

Joint work with @gautierronan.


gautierronan (approved):
All good for me !

abocquet (approved):
LGTM

pierreguilmin:
⚡

PR #94, opened 2023-05-05
No autograd in base solver

pierreguilmin:
Non-differentiable solvers are now supported.

### DIFF
--- a/torchqdynamics/options.py
+++ b/torchqdynamics/options.py
@@ -14,7 +14,7 @@ __all__ = [
 
 
 class Options:
-    GRADIENT_ALG = ['autograd']
+    GRADIENT_ALG = []
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


pierreguilmin:
Note that there is a minor design flaw with the current code: one cannot implement a solver which only supports gradient by the adjoint and not autograd. I don't see any use case for this so... osef 😋

abocquet:
Is there a case for this? All solvers are differentiable at least through autograd no?

pierreguilmin:
No indeed, but if you want to quickly test a solver (say plug in some numpy or qutip code or some new torch diagonalization magic which does not support autograd for complex, or analytic gradients with GRAPE) this is very useful 😋

gautierronan (approved):
Gucci

PR #97, opened 2023-05-05
Add TDTensor class

gautierronan:
As discussed in https://github.com/pierreguilmin/torchqdynamics/pull/89, this adds a class to handle time-dependent tensors.

Not super happy about the way caching is handled, and the duality with `depends_on_H` so I'm happy to hear suggestions. I essentially tried to minimize the number of "wrapping" functions around calls to the Hamiltonian function.


tests/test_h_caching.py
pierreguilmin, line 1: We should move this to `tests/utils/test_td_tensor.py`.

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+from abc import ABC, abstractmethod
+from typing import get_args
+
+import torch
+from torch import Tensor
+
+from .tensor_types import OperatorLike, TDOperatorLike, cdtype, rdtype, to_tensor
+
+
+def to_tdtensor(
```
pierreguilmin, line 12:
```suggestion
def to_td_tensor(
```

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+        dtype = cdtype(dtype)
+
+    if isinstance(x, get_args(OperatorLike)):
+        x = to_tensor(x, dtype=dtype, device=device, is_complex=is_complex)
+        return ConstantTDTensor(x, dtype=dtype, device=device)
+    elif callable(x):
+        if dtype is None and not is_complex:
+            dtype = rdtype(dtype)
+        if device is None:
+            device = torch.device('cpu')
```
pierreguilmin, line 29:
```suggestion
```
I don't think we need this?

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+        return CallableTDTensor(x, dtype=dtype, device=device)
+
+
+class TDTensor(ABC):
+    @abstractmethod
+    def __call__(self, t: float) -> Tensor:
+        """Evaluate a `TDTensor` at a given time"""
```
pierreguilmin, line 36:
```suggestion
        """Evaluate at a given time."""
```

pierreguilmin: And remove ``` of a `TDTensor` ``` in all subsequent docstrings.

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+        pass
+
+
+class ConstantTDTensor(TDTensor):
+    def __init__(
+        self, tensor: Tensor, *, dtype: torch.dtype, device: torch.device | None
+    ):
+        self._tensor = tensor
+        self.dtype = dtype or tensor.dtype
+        self.device = device or tensor.device
```
pierreguilmin, line 66:
```suggestion
    def __init__(self, tensor: Tensor):
        self._tensor = tensor
```
The `dtype` and `device` we need is the one of `tensor`, the conversion already happened with the call to `to_tensor` within `to_tdtensor`.

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+        return CallableTDTensor(
+            self._callable,
+            dtype=self.dtype,
+            device=self.device,
+            shape=new_shape,
+            check_input=False,
+        )
+
+    def has_changed(self, t: float) -> bool:
+        return True
```
pierreguilmin, line 130:
```suggestion
        return t != self._last_t
```

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+        device: torch.device,
+        shape: torch.Size | None = None,
+        check_input: bool = True,
+    ):
+        self._callable = f
+        self.dtype = dtype
+        self.device = device
+        if check_input:
+            tensor = self._check_callable()
+        self._shape = shape or tensor.shape
```
pierreguilmin, line 101:
```suggestion
        if check_input:
            tensor = self._check_callable()
            self._shape = tensor.shape
        else:
            self._shape = shape
```

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,165 @@
...
+        return False
+
+
+class CallableTDTensor(TDTensor):
+    def __init__(
+        self,
+        f: callable,
+        *,
+        dtype: torch.dtype,
+        device: torch.device,
```
pierreguilmin, line 92:
```suggestion
        expected_dtype: torch.dtype,
        expected_device: torch.device,
```
maybe, to differentiate from `dtype` and `device` that are usually used to convert the tensor?

pierreguilmin (changes requested):
Very nice 👏 @abocquet does this fit what you had in mind for PWC Hamiltonian?

torchqdynamics/utils/td_tensor.py
pierreguilmin: Sorry we do now that I read the whole code, but we should catch the default device rather than `'cpu'`. And the first condition can be simplified to `if dtype is None`.

pierreguilmin:
Can you also fix the `__init__` of `MERouchon` calling `self.H(0.0)` which should be unnecessary (at worst we should call it only once in `CallableTDTensor`.

torchqdynamics/utils/td_tensor.py
gautierronan: No, `True` was the right thing to do. But the function name was misleading however. I changed it, and also updated the test function to make sure we're doing the intended evaluations when needed.

torchqdynamics/utils/td_tensor.py
gautierronan: In the end I moved the `check_callable` function outside of the class. I think it's better this way, it makes the class lighter which should be favorable for performance.

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,172 @@
...
+        return ConstantTDTensor(x)
+    elif callable(x):
+        # get default dtype and device
+        if dtype is None:
+            if is_complex:
+                dtype = cdtype(dtype)
+            else:
+                dtype = rdtype(dtype)
+        if device is None:
+            device = get_default_device()
```
abocquet, line 32: With `device = None` the device will be automatically set to the default device
```suggestion
```

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def get_default_device() -> torch.device:
+    """Get the default device."""
+    return torch.ones(1).device
+
+
+def check_callable(
+    f: callable,
+    *,
```
abocquet, line 47: What is the purpose of `*` here? I understand what it does in python, but not what it brings

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def requires_updates(self, t: float) -> bool:
+        return False
+
+
+class CallableTDTensor(TDTensor):
+    def __init__(
+        self,
+        f: callable,
+        *,
```
abocquet, line 139: Same question

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,172 @@
...
+            self._cached_tensor = self._callable(t).view(self._shape)
+            self._cached_t = t
+        return self._cached_tensor
+
+    def size(self, dim: int) -> int:
+        return self._shape[dim]
+
+    def dim(self) -> int:
+        return len(self._shape)
+
```
abocquet, line 162: If `shape` is not given in the constructor, this will always fail. Maybe it's better to update `self.shape` in `__call__` and then in `size` and `dim` if `shape` is not set then we use `__call__` to determine the shape

torchqdynamics/utils/td_tensor.py
```diff
@@ -0,0 +1,172 @@
...
+    def unsqueeze(self, dim: int) -> CallableTDTensor:
+        new_shape = list(self._shape)
+        new_shape.insert(dim, 1)
+        new_shape = torch.Size(new_shape)
+        return CallableTDTensor(
+            self._callable, shape=new_shape, dtype=self.dtype, device=self.device
+        )
```
abocquet, line 169: I see what you did there. Can I suggest this implementation, that may simplify things:
```suggestion
        def new_callable(t):
             res = self._callable(t)
             res = res.unsqueeze(dim)
             return res
        
        return CallableTDTensor(
            new_callable, dtype=self.dtype, device=self.device
        )
```

torchqdynamics/utils/td_tensor.py
abocquet: Could you develop? I'm not sure I follow either

abocquet:
Great work, this should make PWC painless to implement. I just have coding remarks

abocquet (changes requested):
Great work, this should make PWC painless to implement 🚀  I just have coding remarks

torchqdynamics/utils/td_tensor.py
gautierronan: Yes but the goal is to feed it to the `check_callable` function which doesn't know about that. I can call it within the function if you'd prefer however.

torchqdynamics/utils/td_tensor.py
gautierronan: True, I just like the feature so I tend to put it everywhere haha. But I guess it doesn't make sense here.

torchqdynamics/utils/td_tensor.py
gautierronan: Well the way I was thinking about the class was that `shape` was kinda mandatory. Would it be fine if I just remove the default argument of `shape` in the `__init__`, and remove the typing `None` ?

torchqdynamics/utils/td_tensor.py
gautierronan: I would argue my version is better, because yours will result in nested calls to the base `_callable` with multiple calls to `unsqueeze`, whereas mine just applies a single `.view` on the final tensor. In the end it might not change much in terms of performance, but I think it's cleaner.

torchqdynamics/utils/td_tensor.py
gautierronan: I don't think we can just replace it. Because the `t != instance._cache[func.__name__][0]` is also important for other tensors, such as `H_nh` or `M1`. If we only run that within the `has_changed` function, these will be recomputed on every call.

gautierronan: With the current implementation, there has no double checks for caching and everything is handled. However, the caching of `H` and of the other `depends_on_H` is taken care of in different places. This is the only thing that could be improved IMO, but I'm not sure how or if its needed.

gautierronan: So, just to make it clearer, because the current implementation is not trivial (it took me a while to fix it). There are multiple boolean checks all around, which are (in order):

1. `solver.py` > `depends_on_H` > `if t != instance._cache[func.__name__][0]`
2. `td_tensor.py` > `if TDTensor.requires_updates(t)`
3. `td_tensor.py` > `CallableTDTensor.__call__(t)` > `if t != self._cached_t`

- Check (1) is first called everytime a `depends_on_H` function is called. If False, it means the current tensor was already cached at the current time and we use the cached version. If True, we never computed this particular tensor for the current time, and we move on to check (2).
- Check (2) is always False for `ConstantTDTensor` since the Hamiltonian is time-independent. For `CallableTDTensor`, it is always True and we move on to check (3). [NB: for PWC Hamiltonians, I guess it will be sometimes True and sometimes False.]
- Check (3) checks whether we have already computed the Hamiltonian at the current time or not. This way, the Hamiltonian is never recomputed twice for any given time, even if multiple `depends_on_H` functions are called.

abocquet (approved):
Could you just implement the changes in [torchqdynamics/utils/td_tensor.py](https://github.com/pierreguilmin/torchqdynamics/pull/97/files/ce1bcad9bb091b5960a87f2c4283f7a1aa64607f#diff-a26cb8fffa0744b4515fff6229aa5ee7cfcb75e0b9110d3ec0b88522ca8693f3) please?

pierreguilmin:
Wouups sorry for this, I forgot it was targetting this branch. I guess we can keep it now 🤷‍♂️

PR #108, opened 2023-05-10
Doc utils

pierreguilmin:
I propose to start documenting cleanly the part of the source code that will not evolve, for now I would say it only concerns the three following files:
- `torchqdynamics.utils.utils` (addressed by this PR)
- `torchqdynamics.utils.operators`
- `torchqdynamics.utils.states`

To review this PR I suggest that you build the doc locally and check if you agree, rather than reading the modifications to `utils.py`. If you haven't done it yet, it's just two lines:
```shell
pip install -e ".[dev]"
task docserve
```
And then open http://localhost:8000/ in your web browser and check the `Reference` tab. Could you 👍 this message if it works on your local installations @abocquet and @gautierronan?

Screenshot 2023-05-10 at 17 40 05 (2)

Random infos:
- To document the API it's best practice to do as for the tests files and follow the source directory structure.
- I'm not super satisfied with the way we specify the tensor dimensions, we could write custom code to handle this better, but this is currently what's best looking IMO. I think it's nice that the dimension of tensors is part of the typing though, and not part of the description.
- Let's not edit other files with this custom format for tensor typing as it may evolve, we should only transit to it when writing the clean documentation.
- We have a bunch of todos wandering in the code, I think that's fine to have both this and issues, it's much faster to drop a todo for a small thing that's missing.


torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
+
+    Returns:
+        `True` if the last dimension `n` is `1`, `False` otherwise.
+    """
     return x.size(-1) == 1
 
 
 def ket_to_bra(x: Tensor) -> Tensor:
-    """Linear map (bra) representation of a state vector (ket).
+    r"""Linear map $\bra\psi$ representation of a state vector $\ket\psi$.
```
gautierronan, line 38:
```suggestion
    r"""Returns the adjoint state vector $\bra\psi$ representation of a state vector $\ket\psi$.
```

The linear map terminology is not well-known. I would call them either "bras" or "Adjoint state vectors".

gautierronan, line 38:
```suggestion
    r"""Returns the adjoint state vector $\bra\psi$ representation of a state vector $\ket\psi$.
```

The linear map terminology is not well-known. I would call them either "bras" or "Adjoint state vectors" throughout the documentation.

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
 
     Args:
-        x: Tensor of dimension `(..., d, 1)`.
+        x ( Tensor (..., d, 1) ): State vector.
```
gautierronan, line 41:
```suggestion
        x ( Tensor (..., n, 1) ): State vector.
```

gautierronan, line 41:
```suggestion
        x ( Tensor (..., n, 1) ): State vector.
```
Throughout the documentation as well.

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
 
     Returns:
-        Tensor of dimension `(..., 1, d)`.
+        ( Tensor (..., 1, d) ): Corresponding linear map.
```
gautierronan, line 44: Same comment, I don't think this deserves a Returns section. See e.g. https://pytorch.org/docs/stable/generated/torch.adjoint.html?highlight=adjoint#torch.adjoint. Just having the typing at the top in the function definition `torch.adjoint(Tensor) → Tensor` is enough IMO. Otherwise it's too heavy and harder to understand.

Arguably, even the Args section is not useful here if an Example section is provided, but that's a bit more extreme.

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
     """
     return x.adjoint()
 
 
 def ket_to_dm(x: Tensor) -> Tensor:
-    """Density matrix formed by the outer product of a state vector (ket).
+    r"""Density matrix $\ket\psi\bra\psi$ formed by the outer product of a state vector
```
gautierronan, line 50:
```suggestion
    r"""Returns the density matrix $\ket\psi\bra\psi$ associated to a state vector
```

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
     $$
 
     Warning:
-        This definition is different from QuTiP `fidelity()` function which
-        uses the square root fidelity $F' = \sqrt{F}`.
+        This definition is different from `qutip.fidelity()` which uses the square root
+        fidelity $F_\text{qutip} = \sqrt{F}$.
 
     Note:
         This fidelity is also sometimes called the *Uhlmann state fidelity*.
```
gautierronan, line 91: Is this useful? It feels like you're trying to teach us a lesson :D

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
     Returns:
-        Complex-valued overlap (`torch.complex64` or `torch.complex128`).
+        ( Tensor (...) ): Complex-valued overlap.
     """
     return (ket_to_bra(x) @ y).squeeze(-1).sum(-1)
 
 
 def ket_fidelity(x: Tensor, y: Tensor) -> Tensor:
-    r"""Return the fidelity between two state vectors.
+    r"""Fidelity between two state vectors.
```
gautierronan, line 77:
```suggestion
    r"""Fidelity of two state vectors.
```

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,106 @@
...
     Returns:
-        Tensor of size `(...)` holding the real-valued fidelity.
+        ( Tensor (...) ): Real-valued fidelity.
     """
     return ket_overlap(x, y).abs().pow(2).real
 
 
 def dm_fidelity(x: Tensor, y: Tensor) -> Tensor:
-    r"""Return the fidelity between two density matrices.
+    r"""Fidelity between two density matrices.
```
gautierronan, line 104:
```suggestion
    r"""Fidelity of two density matrices.
```

torchqdynamics/utils/utils.py
```diff
@@ -173,70 +184,116 @@ def dissipator(L: Tensor, rho: Tensor) -> Tensor:
...
     operators.
 
+    Examples:
+        >>> rho = tq.tensprod(
+        ...     tq.coherent_dm(20, 2.0),
+        ...     tq.fock_dm(2, 0),
+        ...     tq.fock_dm(5, 1)
+        ... )
+        >>> rho.shape
+        torch.Size([200, 200])
```
gautierronan, line 222: Can you give a similar example for kets first?

gautierronan (changes requested):
Great PR, thanks for taking the time to do this ! I know it's very tedious.

I have not reviewed in detail yet, but here's a first batch of comments.

Overall, I think you went a bit too heavy on the documentation of certain functions that are simple. It's difficult to find a good middle ground between under- and over-documenting, but for now it's a bit on the over side IMO.

Maybe the website style is not helping as well. There are too many boxes and too little text compared to the amount of space for now.

torchqdynamics/utils/utils.py
pierreguilmin: Do you propose to use `n` instead of `d` for all the Hilbert space dimensions?

torchqdynamics/utils/utils.py
pierreguilmin: QuTiP and Wikipédia use between, do you think it's uncorrect english?

torchqdynamics/utils/utils.py
pierreguilmin: Yes I agree, it's just a little annoying I don't know where to place the tensors shape 🤔

pierreguilmin:
Thanks a lot for the review 🙏 

> Overall, I think you went a bit too heavy on the documentation of certain functions that are simple. It's difficult to find a good middle ground between under- and over-documenting, but for now it's a bit on the over side IMO.

100% agree, I forced myself to be exhaustive with returns everywhere and tensor dimensions, but I don't think it's necessary indeed.

> Maybe the website style is not helping as well. There are too many boxes and too little text compared to the amount of space for now.

We can customize this later to our needs, I agree it's a bit heavy for now. I changed the format to the `list` as discussed, and also moved the typing to the function signature.

Waiting for your replies to the various discussions before pushing the changes.

torchqdynamics/utils/utils.py
gautierronan: Yes I agree adjoint state vector is not good terminology. But when I read "linear map" I was like wtf is this haha, and had to look it up to understand you meant a bra. So I would simply stick with "Bra" which is universal. So in this case:
```
Returns the bra representation $\bra\psi$ of a state vector $\ket\psi$.
```

torchqdynamics/utils/utils.py
gautierronan: Yes, I would argue it's better. It's closer to numpy and pytorch notations which uses "n" or "N". See e.g. https://numpy.org/doc/stable/reference/generated/numpy.ndarray.put.html#numpy.ndarray.put and https://pytorch.org/docs/stable/generated/torch.chain_matmul.html.

torchqdynamics/utils/utils.py
gautierronan: No it's because I looked at wikipedia and I think it uses "of" in https://en.wikipedia.org/wiki/Fidelity_of_quantum_states. It's "overlap between" but "fidelity of". I also think it sounds better. But whatever you prefer.

pierreguilmin:
Here's a revised version with custom formatting for the typing. The custom formatting allows specifying the tensor size in the typing part of the documentation. I think it's better, because it's really a typing info, not a description of what is the argument. It's also very easy for the developer to spot the parameters tensor size, they are not hidden in the description of the parameter.

Concerning the modification 

```diff
- r"""Density matrix $\ket\psi\bra\psi$ associated to a state vector $\ket\psi$.
+ r"""Returns the density matrix $\ket\psi\bra\psi$ associated to a state vector $\ket\psi$.
```
I think we should not do that, otherwise we need to add `Returns` at the beginning of the docstring of every function that returns something, which is a bit useless.

Comments welcome! 😊

pierreguilmin:
Screenshot 2023-05-11 at 14 50 22 (2)

torchqdynamics/utils/utils.py
```diff
@@ -22,95 +22,93 @@
 ]
 
 
-def is_ket(x: Tensor) -> Tensor:
-    """Check if a tensor is in state vector format."""
+def is_ket(x: Tensor) -> bool:
+    """Returns True if a tensor is in the format of a state vector, i.e. if its last
+    dimension is 1."""
```
gautierronan, line 27: So in the end should this match the other functions and also add a Parameters and Returns section? At least the Parameters one I think is important, even if its just to write "the input tensor". Right now I feel like it's too short haha. Sorry I'm annoying >.<

torchqdynamics/utils/utils.py
```diff
@@ -173,70 +171,126 @@ def dissipator(L: Tensor, rho: Tensor) -> Tensor:
...
-                                             + \sum_{i=1}^n \mathcal{D}[L_i](\rho).
+        \mathcal{L}(\rho) = -i[H,\rho] + \sum_{k=1}^N \mathcal{D}[L_k](\rho),
     $$
 
+    where $H$ is the system Hamiltonian, $\{L_k\}$ is a set of $N$ jump operators
+    (arbitrary operators) and $\mathcal{D}[L]$ is the Lindblad dissipation superoperator
+    (see [dissipator()][torchqdynamics.dissipator]).
+
+    Note:
+        This superoperator is also sometimes called the *Liouvillian*.
```
gautierronan, line 185:
```suggestion
        This superoperator is also sometimes called *Liouvillian*.
```

torchqdynamics/utils/utils.py
```diff
@@ -173,70 +171,126 @@ def dissipator(L: Tensor, rho: Tensor) -> Tensor:
...
+        [issue #69](https://github.com/pierreguilmin/torchqdynamics/issues/69)).
+
     Note:
-        This function is the equivalent of `qutip.tensor`.
-    """
-    return reduce(torch.kron, x)
+        This function is the equivalent of `qutip.tensor()`.
 
+    Args:
+        *args (n_k, 1) or (n_k, n_k): Sequence of state vectors, density matrices or
```
gautierronan, line 237:
```suggestion
        *args (n, 1) or (n, n): Sequence of state vectors, density matrices or
```
I know you wanted to match the above notations, but it makes it harder to read IMO.

gautierronan (changes requested):
> Concerning the modifications like
> 
> ```diff
> - r"""Density matrix $\ket\psi\bra\psi$ associated to a state vector $\ket\psi$.
> + r"""Returns the density matrix $\ket\psi\bra\psi$ associated to a state vector $\ket\psi$.
> ```
> 
> I think we should not do that, otherwise we need to add `Returns` at the beginning of the docstring of every function that returns something, which is a bit useless.

I think I prefer it the other way around. It doesn't have to be "Returns", but I think they all should start with a verb. If you look at the list of pytorch functions in https://pytorch.org/docs/stable/torch.html, they all start with a verb (often Returns of Creates, but not always). Same for numpy.

Also, would it be possible to add some kind of separation between all functions in the doc? Like just a thin line that spans the screen size or something? Or underline each function name with a touch of color somewhere to make it clear we're in a new function? It will give structure to the webpage, and make it easier to follow. Another possibility is to make function name titles in bold or in color (or both), see e.g. https://numpy.org/doc/stable/reference/arrays.ndarray.html.

torchqdynamics/utils/utils.py
pierreguilmin: I agree it's a bit harder to read, but I think it's crucial to keep it, to differ from a function which would require same shape tensors (as I would expect when reading the proposed suggestion).

torchqdynamics/utils/utils.py
pierreguilmin: They are synonyms and used rather equally by different communities, I just thought if someone searches `Liouvillian` in the search bar he should end up here...

pierreguilmin:
New function name style, I went a bit off track, tell me what you think 😅
Screenshot 2023-05-11 at 18 13 53 (2)
Screenshot 2023-05-11 at 18 13 56 (2)

pierreguilmin:
> No offense but I really don't like it haha. Sorry 😅

No worries haha, it was just a test. I like your suggestion, but there are a few things to fix (it doesn't look very well in dark mode). I propose to discuss this in another PR because it's a bit unrelated, there we can suggest and try different styles. I reverted my styling commit. If you're OK with the content of this PR I propose to merge it. 😊 

Ho and what do you think about renaming state vector to ket everywhere? I have the feeling state vector is a little unclear maybe.

gautierronan:
> > No offense but I really don't like it haha. Sorry 😅
> 
> No worries haha, it was just a test. I like your suggestion, but there are a few things to fix (it doesn't look very well in dark mode and the color is ot in the original palette). I propose to discuss this in another PR because it's a bit unrelated, there we can suggest and try different styles. I reverted my styling commit. If you're OK with the content of this PR I propose to merge it. 😊
> 
> Ho and what do you think about renaming state vector to ket everywhere? I have the feeling state vector is a little unclear maybe.

Yes I'm ok with renaming to ket. It's probably better indeed. Ok for merge

gautierronan (dismissed):

gautierronan (approved):

PR #111, opened 2023-05-12
Fix typing confusion between float and 1-element tensors and remove arguments count check for `CallableTDTensor`

gautierronan:
Adds two small fixes:

1. About the way of counting arguments in the callable (so that other TypeErrors are caught normally)
2. About supplying only tensor-like time values throughout the code, for consistency


pierreguilmin (changes requested):
> About the way of counting arguments in the callable (so that other TypeErrors are caught normally)

Weew we should not use inspect if we can avoid to, actually I would suggest completely removing the check of the callable tensor arguments. Argument checking is a bit of a nightmare and hard to do correctly, Python philosophy is to write clear doc and not enforce arguments type in the code.

> About supplying only tensor-like time values throughout the code, for consistency

Is this really needed? (does it fix a bug?) I feel it makes the code a little unclear (and potentially negligibly less efficient, bc a built-in `float` vs a `Tensor` is not the same memory wise).

gautierronan:
> Weew we should not use inspect if we can avoid to, actually I would suggest completely removing the check of the callable tensor arguments. Argument checking is a bit of a nightmare and hard to do correctly, Python philosophy is to write clear doc and not enforce arguments type in the code.

Ok that's fine with me ! I removed it.

> Is this really needed? (does it fix a bug?) I feel it makes the code a little unclear (and potentially negligibly less efficient, bc a built-in `float` vs a `Tensor` is not the same memory wise).

Yes I think it's really needed. I run into a problem when i was calling `torch.round(t / T)` in my own optimization, because this torch function only takes torch tensors as input. Since the time is called back to a user-defined function, I think it's important to keep a consistent type throughout the codebase. Furthermore, I think there could be cases where there are time-dependent gradients, so it should be a Tensor.

pierreguilmin:
> Yes I think it's really needed. I run into a problem when i was calling torch.round(t / T) in my own optimization, because this torch function only takes torch tensors as input. Since the time is called back to a user-defined function, I think it's important to keep a consistent type throughout the codebase.

But then we need to modify the type of `dt` and `t` everywhere no? I think it's better to convert the time to tensor only where needed (I expect one place or two in the code) rather than carrying a one-element tensor everywhere. Moreover the typing for the callable Hamiltonian would now have to be `Callable[[Tensor], Tensor]` instead of the current `Callable[[float], Tensor]]`, which I find very confusing...

> Furthermore, I think there could be cases where there are time-dependent gradients, so it should be a Tensor.

I didn't get this 🤔

dynamiqs/solvers/ode/adaptive_integrator.py
```diff
@@ -84,7 +85,7 @@ def init_tstep(self, f0: Tensor, y0: Tensor, t0: float) -> float:
         return min(100 * h0, h1)
```
pierreguilmin, line 85: We could maybe just call `.item()` once here?

dynamiqs/solvers/ode/adaptive_integrator.py
```diff
@@ -154,7 +155,7 @@ def build_tableau(self, target):
         return alpha, beta, csol5, csol5 - csol4
 
     def step(
-        self, f0: Tensor, y0: Tensor, t0: float, dt: float
+        self, f0: Tensor, y0: Tensor, t0: Tensor, dt: float
```
pierreguilmin, line 158: We should maybe add a comment saying that `t0` is a one-element tensor. Another option is to convert `t_save` to a numpy array to have proper floats. I quite like this option, because it makes clear that no gradient is computed w.r.t. time variable. It would also avoid the `dt = ts.item() - t` in `adaptive_solver.py`. And finally going through the codebase this would fix the last typing mistake we need, in `.solver.py` for the `def next_tsave(self) -> float:` function that currently returns a tensor.

pierreguilmin (changes requested):
Thanks a lot, I think it's much clearer!

pierreguilmin:
Alright, I think this fixes the type confusion we had. Now the big question is wether we want to keep `t_save` as a `torch.Tensor` or not. I don't really care to be honest, I would say both is fine, now that the typing is corrected 😊

PR #114, opened 2023-05-13
Add logos and accent colors to doc

gautierronan:


docs/stylesheets/extra.css
```diff
@@ -0,0 +1,8 @@
+:root {
+    --md-primary-fg-color:        #ED7764;
+    --md-primary-fg-color--light: #edc7c2;
+    --md-primary-fg-color--dark:  #903527;
+    --md-accent-fg-color:        #6caca4;
+    --md-accent-fg-color--light: #c5e3df;
+    --md-accent-fg-color--dark:  #2e5f5a;
+}
```
pierreguilmin, line 8:
```suggestion
:root {
    --md-primary-fg-color:        #ed7764;
    --md-primary-fg-color--light: #edc7c2;
    --md-primary-fg-color--dark:  #903527;
    --md-accent-fg-color:         #6caca4;
    --md-accent-fg-color--light:  #c5e3df;
    --md-accent-fg-color--dark:   #2e5f5a;
}
```

pierreguilmin (changes requested):
Nice!! I just find a bit weird that the accent color is so different from the primary color, I think I like what https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#accent-color does with an accent color a bit darker, it feels more uniform and in the brand that having a completely different color.

abocquet (approved):

docs/stylesheets/extra.css
gautierronan: > Nice!! I just find a bit weird that the accent color is so different from the primary color, I think I like what https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#accent-color does with an accent color a bit darker, it feels more uniform and in the brand that having a completely different color.

I'm fine with any other color. I couldn't find where it was being used in the doc actually. So I just put one for now, but if you have other recommandations I can change !

pierreguilmin (approved):
Thanks a lot for this 🙏

PR #116, opened 2023-05-13
Uniformize short dev doc to track tensor shapes in solvers

pierreguilmin:
I think following the tensor shapes is really hard if we don't force ourselves to inline comment, although it's a bit annoying to do. What do you guys think?

I didn't do the `backward_augmented` methods of the mesolve Rouchon solvers. @gautierronan if you have the motivation to add it 😉


abocquet (approved):
This is a great idea, thanks!

gautierronan (approved):

PR #120, opened 2023-05-15
Base `MESolver` class

pierreguilmin:
- fix caching for `MERouchon1` adjoint
- use caching in `MEEuler` and `MEAdaptive`
- factor common code


tests/mesolve/me_solver_tester.py
```diff
@@ -82,7 +82,7 @@ def _test_adjoint(
         *,
         num_t_save: int,
         rtol: float = 1e-5,
-        atol: float = 1e-3,
+        atol: float = 1e-2,
```
gautierronan, line 85: Did you have to change the test tolerence for it to pass? Or is it for something else?

gautierronan (commented):
Nice PR :)

tests/mesolve/me_solver_tester.py
pierreguilmin: Yes, sorry I should have made this more explicit in the PR description. I'm guessing we were just on the edge btw both tolerances because adding `assert torch.allclose(M0_adj, self.M0_adj(t))` to compare the computed `M0_adj` and the cached one they were the same.

gautierronan (approved):

PR #121, opened 2023-05-15
Rename `dt` to `delta_t` for `Propagator`

pierreguilmin:


pierreguilmin:
@abocquet this solver is actually a good use case to implement caching on more arguments than `t`. Here we would enjoy caching the propagator `torch.matrix_exp(-1j * self.H(t) * delta_t)` on `t` _and_ `delta_t`. Note that currently the `delta_t` are quite unreliably different, e.g.:
```python
0.1
0.1
0.10000001
0.099999994
0.099999994
0.100000024
0.099999964
0.100000024
0.099999964
0.100000024
```
which would have to be fixed to enjoy from the caching.

abocquet (approved):

gautierronan (approved):

PR #126, opened 2023-05-23
Fix adjoint gradients from exp_ops

gautierronan:
Fixes https://github.com/pierreguilmin/dynamiqs/issues/117

In the end it's just a few lines of code, but it took me the whole day to fix because it's conceptually quite different. Anyway it works as intended now.

### DIFF
--- a/dynamiqs/solvers/ode/fixed_solver.py
+++ b/dynamiqs/solvers/ode/fixed_solver.py
@@ -135,7 +135,7 @@ class AdjointFixedAutograd(torch.autograd.Function):
         ctx.save_for_backward(solver.y_save)
 
         # returning `y_save` is required for custom backward functions
-        return solver.y_save
+        return solver.y_save, solver.exp_save
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


dynamiqs/solvers/ode/fixed_solver.py
```diff
@@ -135,7 +135,7 @@ def forward(
         ctx.save_for_backward(solver.y_save)
 
         # returning `y_save` is required for custom backward functions
-        return solver.y_save
+        return solver.y_save, solver.exp_save
```
pierreguilmin, line 138: We need to fix the return type too.

pierreguilmin (dismissed):
Thank you!! I can't give a proper review because this part of the code is still a bit obscure to me (but it's on my todo do understand it better, read the related articles and review it 😋). So I trust you 😉 It'd be nice to add tests as well in another PR, to make sure we don't break this feature in later improvements.

pierreguilmin (approved):

PR #129, opened 2023-05-25
Unify caching and fix adjoint tests

gautierronan:
This PR unifies the way caching is handled by grouping everything inside a generalized `TDTensor` class. This also fixes the adjoint tests that were broken, see https://github.com/pierreguilmin/dynamiqs/issues/128.

This is not super well polished yet, I mostly wanted to see if you guys agreed with the way this is written. If you do I will add back the tests for caching. If not, I'm completely open to other proposals.


gautierronan:
Here's a second version because I was not happy with the initial one. In the end I used a simple `lru_cache` decorator like @abocquet was proposing initially for the Kraus operators, and for functions that only depend on time, I use the generalized `TDTensor` class. For the issue with caching grad or no_grad tensors, I defined a `clear_cache` function which is called once at the end of every `run_nograd`.

dynamiqs/solvers/solver.py
```diff
@@ -128,6 +101,16 @@ def _save_exp_ops(self, y: Tensor):
         if len(self.exp_ops) > 0:
             self.exp_save[..., self.save_counter] = bexpect(self.exp_ops, y)
 
+    def clear_cache(self):
```
pierreguilmin, line 104: This is quite a dangerous way to do what you want, let's keep this for now, I'll propose something cleaner soon.

dynamiqs/solvers/solver.py
```diff
@@ -128,6 +101,16 @@ def _save_exp_ops(self, y: Tensor):
...
+        """Clear the cache of all `TDTensor` attributes."""
+        for attr_str in dir(self):
+            attr = getattr(self, attr_str)
+            if isinstance(attr, TDTensor):
+                attr.clear_cache()
+
+            if hasattr(attr, 'cache_clear'):
+                attr.cache_clear()
```
pierreguilmin, line 112: Why adding this?

dynamiqs/mesolve/rouchon.py
```diff
@@ -20,26 +21,69 @@ def __init__(self, *args, **kwargs):
...
         # -> (b_H, 1, n, n)
         return self.I - 1j * dt * self.H_nh(t)
 
-    @depends_on_H
-    def M0_adj(self, t: float, dt: float) -> Tensor:
+    @lru_cache(maxsize=1)
+    def M1s(self, t: float, dt: float) -> Tensor:
         # -> (b_H, 1, n, n)
-        return self.I + 1j * dt * self.Hdag_nh(t)
+        return sqrt(dt) * self.jump_ops
```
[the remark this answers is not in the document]
pierreguilmin: Actually the Rouchon solver is a fixed time step scheme so `dt` is always the same, we don't need to cache on dt anywhere in this file.

pierreguilmin (changes requested):
I like the idea!!

pierreguilmin:
Replaced by #133.

PR #130, opened 2023-05-26
Add caching for SEPropagator

gautierronan:

PR #133, opened 2023-06-09
Caching tensors

### LINKED ISSUE #128: Caching checks for other arguments and for `requires_grad`
A rehaul of the caching is required to make the adjoint tests pass cleanly. Right now it's bugged. When caching, we need to:

- check for other arguments being changed (typically `dt`)
- check whether `requires_grad` is True or False in the current scope (we should update the cache if this has changed), otherwise the given Tensor might be disconnected from the operation graph.

Once this if fixed, the adjoint tests should pass with `atol = 1e-4` instead of the current `atol = 1e-2`.

pierreguilmin:
Here’s my proposal for the caching, built upon previous work from @abocquet and the PRs #129 and #130 from @gautierronan. This PR replaces them. Also fixes #128.

**TL;DR** Look at the [propagator.py](https://github.com/dynamiqs/dynamiqs/blob/fix-caching/dynamiqs/sesolve/propagator.py) file of this branch to understand how the caching is handled. Then you can look at the [MERouchon1](https://github.com/dynamiqs/dynamiqs/blob/fix-caching/dynamiqs/mesolve/rouchon.py#L23) `__init__` and `forward` methods to see how handy it is to defined cached tensors.

🎯 **Goal**
- We want to cache some tensors that depend on some `float` parameters (like `t` or `delta_t`) and/or other cached tensors (for example the tensor `M0 = I - 1j * dt * H_nh` depends on the tensor `H_nh` which is also cached).
- We have to use `lru_cache` or similar tooling to handle the caching cleanly for multithreading.
- We want to have a fast and simple way to write a cached tensor.

🔪 **Sharp bits** (probably to read after a first look at the proposed solution)
- We can use `lru_cache` to cache methods taking tensors as argument, because `lru_cache` checks cache hit using objects `__hash__` (and not `__eq__`), and the hash of a PyTorch tensor is just the hash of its memory adress. This means `lru_cache` never checks the equality of two tensors, it just checks wether they are exactly the same object.
- `lru_cache` from the Python standard library is not instance-specific, this means that the cache is shared by all instances. Thus I propose to rather use `lru_cache` from the `methodtools` library, which serves exactly the same purpose, but handles the cache separately for each class instance.
- When we compute the gradient using the adjoint, we first use `torch.no_grad` to run the forward pass without storing the graph of operations. As a result, for a constant Hamiltonian for example, the computed tensors are detached from the graph (their `require_grad` attribute is set to `False`). As a consequence, if we later use these cached tensors outside the `torch.no_grad` block, they will result in cancelling gradients (they are considered as constants by PyTorch). So it's crucial to clear the cache right after any `torch.no_grad` block. This is done by having a custom caching decorator, which includes the current PyTorch grad mode status in the caching to make sure tensors are re-computed when it switches.
- The `TDTensor` class _does not_ handle the caching of `H`, this could be done with a simple call to `cache`. What it handles is encapsulating the dim, size, unsqueeze calls for a constant, piecewise constant or callable tensor. Note that we don't need the `requires_update` anymore: if the `H` tensor is recomputed, every object depending on it will be recomputed thanks to how `lru_cache` works.

I am quite happy with the solution, I think it makes the caching very explicit: it's easy to understand what depends on what.


pierreguilmin:
Special thanks to @clementblaudeau who helped me think this through. 🙏🏼

pierreguilmin:
I replaced 6f84d9ee93de8b1fd7e795264ef57cab61fd951e by a76efd5fdc5d19794bafdd91d2313f067d68c80b to prevent the issue of cached tensors detached from the graph not recomputed. The caching function now also includes PyTorch grad mode status, forcing to recompute these tensors if it changes.

dynamiqs/mesolve/me_solver.py
```diff
@@ -1,28 +1,23 @@
...
-        # -> (b_H, 1, n, n)
-        return self.H(t) - 0.5j * self.sum_no_jump
+        # define cached operator
+        # non-hermitian Hamiltonian
+        self.H_nh = cache(lambda H: H - 0.5j * self.sum_no_jump)  # (b_H, 1, n, n)
 
     def lindbladian(self, t: float, rho: Tensor) -> Tensor:
-        H_nh_rho = self.H_nh(t) @ rho  # (b_H, b_rho, n, n)
+        H = self.H(t)
+        H_nh_rho = self.H_nh(H) @ rho  # (b_H, b_rho, n, n)
```
gautierronan, line 21: Maybe group these two lines together? I don't know how important it is to not define new objects, but here it's free

dynamiqs/mesolve/rouchon.py
```diff
@@ -20,26 +19,18 @@ def __init__(self, *args, **kwargs):
...
 
-    @depends_on_H
-    def M0(self, t: float, dt: float) -> Tensor:
-        # build time-dependent Kraus operators
-        # -> (b_H, 1, n, n)
-        return self.I - 1j * dt * self.H_nh(t)
+        # define cached operators
+        self.M0 = cache(lambda H_nh: self.I - 1j * self.dt * H_nh)  # (b_H, 1, n, n)
+        self.M0dag = cache(lambda M0: M0.adjoint())  # (b_H, 1, n, n)
+        self.M0inv = cache(lambda H_nh: self.I + 1j * self.dt * H_nh)  # (b_H, 1, n, n)
```
gautierronan, line 30: `M0inv` is not good notation here (you're not defining the inverse). Maybe `M0rev` instead? as in "reverse"

dynamiqs/mesolve/me_solver.py
gautierronan: If it doesn't change anything, keep it as you did since it's more readable.

dynamiqs/mesolve/rouchon.py
```diff
@@ -53,26 +44,55 @@ def forward(self, t: float, dt: float, rho: Tensor) -> Tensor:
         """
         # rho: (b_H, b_rho, n, n) -> (b_H, b_rho, n, n)
 
+        # compute cached operators
+        H = self.H(t)  # (b_H, 1, n, n)
+        H_nh = self.H_nh(H)  # (b_H, 1, n, n)
+        M0 = self.M0(H_nh)  # (b_H, 1, n, n)
```
gautierronan, line 50: Same comment as before about grouping lines.

gautierronan (changes requested):
Great PR, great work ! Just a few minor comments.

Have you checked that `cache` and `TDTensor` play well together? I'm worried that the object might not be cached properly and recomputed at every step even though it might not have changed. This could be a separate test in a subsequent PR, but it doesn't have to be I think. Just checking it once is enough.

dynamiqs/mesolve/me_solver.py
pierreguilmin: Yes I thought it'd be more readable by splitting. Also for someone who wants to read along and understand how the caching works nothing is hidden: `H` is a new tensor only if `self.H(t)` returns something new, and the subsequent tensors are only re-computed if `H` is a new tensor.

pierreguilmin:
Thanks for the review!

> Have you checked that cache and TDTensor play well together?

Yes! We need to write some proper tests but I tested locally.

dynamiqs/mesolve/rouchon.py
pierreguilmin: Also for someone who wants to read along and understand how the caching works nothing is hidden: `H` is a new tensor only if `self.H(t)` returns something new, and the subsequent tensors are only re-computed if `H` is a new tensor.

gautierronan: Yes let's keep it like you wrote already. I don't think there should be any performance/memory gain by one-lining this type of code in Python. There could be however if it was something like
```
H = A
H += B
H *= c
```
because operations would be fused (I think, not 100% sure).

gautierronan:
> Thanks for the review!
> 
> > Have you checked that cache and TDTensor play well together?
> 
> Yes! We need to write some proper tests but I tested locally.

Ok nice. I'd argue there's no need to test that since there's not much chance for it to break. But if you want to write a test that's good with me.

gautierronan:
Also, you need to add `methodtools` to the dependencies I think.

gautierronan (approved):
LGTM. Nice work !

pierreguilmin:
I think GitHub now messes up in the "Files changed" tab, if you look at the branch file the https://github.com/dynamiqs/dynamiqs/blob/fix-caching/pyproject.toml is correct.

PR #138, opened 2023-06-12
Factor adjoint tests

pierreguilmin:


pierreguilmin:
@gautierronan it seems that the gradient on `exp_ops` computed with adjoint for the Rouchon method is much worse than the previous test was suggesting, when we compare it with the autograd gradient.

gautierronan:
> @gautierronan it seems that the gradient on `exp_ops` computed with the adjoint for the Rouchon method is much worse than the previous test was suggesting, when we compare it with the autograd gradient.

I didn't catch that. What do you mean?

pierreguilmin:
See line 27 of `tests/mesolve/test_rouchon.py`, I had to lower the `atol` a lot to have the test pass.

pierreguilmin:
Sorry, I should probably make an explicite separated commit every time I modify test tolerances.

pierreguilmin:
Yes!

gautierronan:
> Yes!

Just looking at the code, could it be because you're running `torch.autograd.grad` twice in a row without recomputing the graph? Not sure but maybe you need to `.reset()` and rerun `mesolve` again?

gautierronan:
> > Yes!
> 
> Just looking at the code, could it be because you're running `torch.autograd.grad` twice in a row without recomputing the graph? Not sure but maybe you need to `.reset()` and rerun `mesolve` again?

If this is indeed the issue, note that you can do the same by calling `torch.autograd.grad` just once, but with its first argument as `(rho_loss, exp_loss)` and it will return the gradients as a tuple `(rho_grad, exp_grad)` as well. I'm not sure `retain_graph` works with the adjoint since we're not really building the graph, but rather computing the gradients through a custom backward method.

pierreguilmin:
This is a little better, I guess the issue is still here, but we summed the gradients of `rho` and `exp` (`torch.autograd.grad` does not return a tuple, it sums the outputs gradients for each parameter) and the gradient on `rho` has a much better accuracy than the one on `exp` for some reason.

pierreguilmin:
Thanks for helping me with the debugging btw 😊

gautierronan:
> This is a little better, I guess the issue is still here, but we summed the gradients of `rho` and `exp` (`torch.autograd.grad` does not return a tuple, it sums the outputs gradients for each parameter) and the adjoint gradient on `rho` has a much better accuracy than the one on `exp` for some reason.

My bad you're right. It returns a tuple if you do that on the parameters argument. I think it'd be nice to run separate tests then, rather than the sum of the two gradients.

pierreguilmin:
I did not split the test (because I prefer that we do not solve the evolution twice while it's unnecessary), I just added specific rtol/atol for rho and exp, because as you said it just depends on the loss function.

I also added logging, it doesn't print if the test succeeds, but if it fails it does, e.g. for Rouchon2:
```shell
WARNING  root:me_solver_tester.py:107 grad_rho     = (tensor(0.0611), tensor(-0.0306), tensor(0.0611))
WARNING  root:me_solver_tester.py:108 grad_rho_adj = (tensor(0.0612), tensor(-0.0305), tensor(0.0610))
WARNING  root:me_solver_tester.py:109 grad_exp     = (tensor(0.2046), tensor(-0.7573), tensor(6.1726))
WARNING  root:me_solver_tester.py:110 grad_exp_adj = (tensor(0.2076), tensor(-0.7579), tensor(6.1708))
```

tests/mesolve/me_solver_tester.py
```diff
@@ -82,80 +75,47 @@ def _test_adjoint(
         system: OpenSystem,
         *,
         num_t_save: int,
-        rtol: float = 1e-5,
-        atol: float = 3e-3,
+        rtol_rho: float = 1e-5,
+        rtol_exp: float = 1e-5,
+        atol_rho: float = 3e-3,
+        atol_exp: float = 1e-2,
```
gautierronan, line 81: I think we're defaulting these in the wrong way. What we want is to test whether the gradients are close but in relative value rather than in absolute value. So, I think we should raise the `rtol` quite a lot, and maybe lower the `atol`. This would also help standardize the values between the `exp` check and `rho` check and between Rouchon1 and Rouchon2.

Typically, I think 5% relative error on the gradients is fine, so `rtol=5e-2`, while we should maybe only tolerate ~0.001-0.003 absolute error (for very small gradients close to 0, e.g. when the gradient descent is converging to a global minima), so `atol = 3e-3`. This way all tests pass with the same tolerences. 

Then, you can lower these for Rouchon2 if you want (e.g. `rtol=1e-2` and `atol=1e-3`), but I don't think that's required.

tests/mesolve/test_rouchon.py
```diff
@@ -24,11 +24,7 @@ def test_rho_save(self):
 
     def test_adjoint(self):
         options = dq.options.Rouchon1(dt=1e-3)
-        self._test_adjoint(options, grad_leaky_cavity_8, num_t_save=11)
-
-    def test_adjoint_expops(self):
-        options = dq.options.Rouchon1(dt=1e-3)
-        self._test_adjoint_expops(options, grad_leaky_cavity_8, num_t_save=11)
+        self._test_adjoint(options, grad_leaky_cavity_8, num_t_save=11, atol_exp=2e-1)
```
gautierronan, line 27: See comment above.

gautierronan (changes requested):

tests/mesolve/test_rouchon.py
gautierronan: (So, in essence, it's strange to raise the atol to 0.2 for all tests here, when it's only needed for the one gradient that's very large and which features "only" a 2% error)

gautierronan:
> 100% agree, what I did is dumb.

Well you just copied my previous mistake, so it's on me :D

PR #141, opened 2023-06-12
Add `smesolve`

pierreguilmin:
Here's a first attempt at writing `smesolve`, with a basic Euler scheme (that works!), and only for diffusive-type measurement. Thanks to the arbitrary batching of the state and the reorganisation of the solvers, this was actually quite easy. I had to slighlty modify the saving logic of the `Solver` base class (see first commit).

Left to do (in other PRs):
- [ ] Rouchon method
- [ ] run trajectories in //
- [ ] testing
- [ ] jump SME (later)

Two things we need to discuss together:
1. Currently the signal value we get at each time step `dt` is integrated to give a value for each entry of `t_save`. Maybe it's better that the user specifies a separate `tbin` parameter (which divides `t_save[-1]`) over which we integrate the signal regularly, and keep `t_save` only for the state?
2. Moreover, the returned signal has one element less than `t_save`, because technically we don't have the signal value for the first time. This is a bit annoying for post-processing. We could _fake_ a first measurement by computing the signal on `rho_0`?

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -97,7 +97,7 @@ def mesolve(
     options = options or Dopri5()
 
     # define the solver
-    args = (H_batched, rho0_batched, t_save, exp_ops, options)
+    args = (H_batched, rho0_batched, t_save, t_save, exp_ops, options)
     if isinstance(options, Rouchon1):
         solver = MERouchon1(*args, jump_ops=jump_ops)
     elif isinstance(options, Rouchon1_5):
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -4,7 +4,7 @@ import torch
 from torch import Tensor
 
 from ..options import Dopri5, Euler, Options, Rouchon1, Rouchon1_5, Rouchon2
-from ..utils.solver_utils import check_time_array
+from ..utils.solver_utils import check_time_tensor
 from ..utils.tensor_formatter import TensorFormatter
 from ..utils.tensor_types import OperatorLike, TDOperatorLike, TensorLike
 from .adaptive import MEDormandPrince5
@@ -93,7 +93,7 @@ def mesolve(
 
     # convert t_save to a tensor
     t_save = torch.as_tensor(t_save, dtype=options.rdtype, device=options.device)
-    check_time_array(t_save, 't_save')
+    check_time_tensor(t_save, arg_name='t_save')
 
     # default options
     options = options or Dopri5()
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -4,6 +4,7 @@ import torch
 from torch import Tensor
 
 from ..options import Dopri5, Euler, Options, Rouchon1, Rouchon1_5, Rouchon2
+from ..utils.solver_utils import check_time_array
 from ..utils.tensor_formatter import TensorFormatter
 from ..utils.tensor_types import OperatorLike, TDOperatorLike, TensorLike
 from .adaptive import MEDormandPrince5
@@ -92,12 +93,14 @@ def mesolve(
 
     # convert t_save to a tensor
     t_save = torch.as_tensor(t_save, dtype=options.rdtype, device=options.device)
+    check_time_array(t_save, 't_save')
 
     # default options
     options = options or Dopri5()
 
     # define the solver
-    args = (H_batched, rho0_batched, t_save, t_save, exp_ops, options)
+    t_save_all = t_save
+    args = (H_batched, rho0_batched, t_save_all, t_save, exp_ops, options)
     if isinstance(options, Rouchon1):
         solver = MERouchon1(*args, jump_ops=jump_ops)
     elif isinstance(options, Rouchon1_5):
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -99,8 +99,7 @@ def mesolve(
     options = options or Dopri5()
 
     # define the solver
-    t_save_all = t_save
-    args = (H_batched, rho0_batched, t_save_all, t_save, exp_ops, options)
+    args = (H_batched, rho0_batched, t_save, exp_ops, options)
     if isinstance(options, Rouchon1):
         solver = MERouchon1(*args, jump_ops=jump_ops)
     elif isinstance(options, Rouchon1_5):
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -4,7 +4,8 @@ import torch
 
 from ..options import Dopri5, Euler, Options, Rouchon1, Rouchon1_5, Rouchon2
 from ..solvers.result import Result
-from ..solvers.utils.tensor_formatter import TensorFormatter, check_time_tensor
+from ..solvers.utils.tensor_formatter import TensorFormatter
+from ..solvers.utils.utils import check_time_tensor
 from ..utils.tensor_types import OperatorLike, TDOperatorLike, TensorLike
 from .adaptive import MEDormandPrince5
 from .euler import MEEuler
--- a/dynamiqs/smesolve/euler.py
+++ b/dynamiqs/smesolve/euler.py
@@ -1,0 +1,20 @@
+from torch import Tensor
+
+from ..solvers.ode.fixed_solver import FixedSolver
+from .sme_solver import SMESolver
+
+
+class SMEEuler(SMESolver, FixedSolver):
+    def forward(self, t: float, rho: Tensor) -> Tensor:
+        # rho: (b_H, b_rho, ntrajs, n, n) -> (b_H, b_rho, ntrajs, n, n)
+
+        # sample Wiener process
+        dw = self.sample_wiener(self.dt)
+
+        # update state
+        drho = self.dt * self.lindbladian(t, rho) + self.diff_backaction(rho, dw)
+
+        # update measured signal
+        self.update_meas(rho, dw)
+
+        return rho + drho
--- a/dynamiqs/smesolve/euler.py
+++ b/dynamiqs/smesolve/euler.py
@@ -12,9 +12,9 @@ class SMEEuler(SMESolver, FixedSolver):
         dw = self.sample_wiener(self.dt)
 
         # update state
-        drho = self.dt * self.lindbladian(t, rho) + self.diff_backaction(rho, dw)
+        drho = self.dt * self.lindbladian(t, rho) + self.diff_backaction(dw, rho)
 
         # update measured signal
-        self.update_meas(rho, dw)
+        self.update_meas(dw, rho)
 
         return rho + drho
--- a/dynamiqs/smesolve/sme_solver.py
+++ b/dynamiqs/smesolve/sme_solver.py
@@ -1,0 +1,110 @@
+from __future__ import annotations
+
+from math import sqrt
+
+import torch
+from torch import Tensor
+
+from ..mesolve.me_solver import MESolver
+from ..options import Options
+from ..utils.solver_utils import cache
+from ..utils.td_tensor import TDTensor
+from ..utils.utils import trace
+
+
+class SMESolver(MESolver):
+    def __init__(
+        self,
+        H: TDTensor,
+        y0: Tensor,
+        exp_ops: Tensor,
+        options: Options,
+        *,
+        jump_ops: Tensor,
+        meas_ops: Tensor,
+        etas: Tensor,
+        generator: torch.Generator,
+        t_save_y: Tensor,
+        t_save_meas: Tensor,
+    ):
+        t_save = torch.cat((t_save_y, t_save_meas)).unique().sort()[0]
+        super().__init__(H, y0, t_save, t_save_y, exp_ops, options, jump_ops=jump_ops)
+
+        self.meas_ops = meas_ops  # (len(meas_ops), n, n)
+        self.t_save_meas = t_save_meas  # (len(t_save_meas))
+        self.etas = etas  # (len(meas_ops))
+        self.generator = generator
+
+        self.save_meas_counter = 0
+
+        # initialize additional save tensors
+        batch_sizes = self.y0.shape[:-2]
+
+        self.meas_shape = (*batch_sizes, len(self.meas_ops))
+
+        # meas_save: (..., len(meas_ops), len(t_save_meas))
+        if len(t_save_meas) > 0:
+            self.meas_save = torch.zeros(
+                *self.meas_shape,
+                len(self.t_save_meas) - 1,
+                dtype=self.dtype_real,
+                device=self.device,
+            )
+        else:
+            self.meas_save = None
+
+        # tensor to hold the sum of measurement results on a time bin
+        self.bin_meas = torch.zeros(self.meas_shape)  # (..., len(etas))
+
+    def _save(self, t: float, y: Tensor):
+        if t in self.t_save_y:
+            super()._save(t, y)
+        if t in self.t_save_meas:
+            self.save_meas()
+
+    def save_meas(self):
+        if self.save_meas_counter != 0:
+            t_bin = (
+                self.t_save_meas[self.save_meas_counter]
+                - self.t_save_meas[self.save_meas_counter - 1]
+            )
+            self.meas_save[..., self.save_meas_counter - 1] = self.bin_meas / t_bin
+            self.bin_meas = torch.zeros_like(self.bin_meas)
+        self.save_meas_counter += 1
+
+    def sample_wiener(self, dt: float) -> Tensor:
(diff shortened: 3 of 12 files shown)


pierreguilmin:
Let's roll!

dynamiqs/smesolve/sme_solver.py
```diff
@@ -0,0 +1,110 @@
...
+        # L @ rho + rho @ Ldag
+        Lp_rho = self.Lp(rho)  # (..., b, n, n)
+
+        # L @ rho + rho @ Ldag - Tr(L @ rho + rho @  Ldag) rho
+        tmp = Lp_rho - torch.einsum('...bii,...jk->...bjk', Lp_rho, rho)
+        # tmp: (..., b, n, n)
+
+        # sum sqrt(eta) * dw * [L @ rho + rho @ Ldag - Tr(L @ rho + rho @ Ldag) rho]
+        fact = (self.etas.sqrt() * dw).to(self.dtype)
+        return torch.einsum('...b,...bij->...ij', fact, tmp)
```
gautierronan, line 106: No need to use einsum for this type of simple operation. It's probably more costly due to the checks it runs. Also I don't think conversion to complex is useful as long as the precision is the same.
```suggestion
        fact = self.etas.sqrt() * dw
        return (fact[..., None, None] * tmp).sum(-3)
```

gautierronan: Also I don't like the naming `fact`, I'm reading "factorial". Just use the full name instead, so `factor` or `prefactor` even.

dynamiqs/smesolve/sme_solver.py
```diff
@@ -0,0 +1,110 @@
...
+        # L @ rho + rho @ Ldag
+        L_rho = torch.einsum('bij,...jk->...bik', self.meas_ops, rho)
+        return L_rho + L_rho.mH
+
+    def diff_backaction(self, rho: Tensor, dw: Tensor) -> Tensor:
+        # Compute the measurement backaction term of the diffusive SME.
+        # $$ \sum_k \sqrt{\eta_k} \mathcal{M}[L_k](\rho) dW_k $$
+        # where
+        # $$ \mathcal{M}[L](\rho) = L \rho + \rho L^\dag -
+        # \mathrm{Tr}\left[(L + L^\dag) \rho\right] \rho $$
```
gautierronan, line 93: Why not use a docstring?

dynamiqs/smesolve/euler.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SMEEuler(SMESolver, FixedSolver):
+    def forward(self, t: float, rho: Tensor) -> Tensor:
+        # rho: (b_H, b_rho, ntrajs, n, n) -> (b_H, b_rho, ntrajs, n, n)
+
+        # sample Wiener process
+        dw = self.sample_wiener(self.dt)
+
+        # update state
+        drho = self.dt * self.lindbladian(t, rho) + self.diff_backaction(rho, dw)
```
gautierronan, line 15: For consistency I would invert the two arguments in `self.diff_backaction` and put `rho` last.

dynamiqs/smesolve/sme_solver.py
```diff
@@ -0,0 +1,110 @@
...
+        # L @ rho + rho @ Ldag - Tr(L @ rho + rho @  Ldag) rho
+        tmp = Lp_rho - torch.einsum('...bii,...jk->...bjk', Lp_rho, rho)
+        # tmp: (..., b, n, n)
+
+        # sum sqrt(eta) * dw * [L @ rho + rho @ Ldag - Tr(L @ rho + rho @ Ldag) rho]
+        fact = (self.etas.sqrt() * dw).to(self.dtype)
+        return torch.einsum('...b,...bij->...ij', fact, tmp)
+
+    def update_meas(self, rho: Tensor, dw: Tensor):
+        tr = trace(self.Lp(rho)).real
```
gautierronan, line 109: You're computing this guy twice per time step here. Even if it's cached, the trace is not. Could we compute it only once? I think it would be clearer.

gautierronan: Also, `tr` is not a good variable name. Maybe `exp_val` or `meas_val` instead? Since it's the expectation value of (L+Ldag).

dynamiqs/smesolve/smesolve.py
```diff
@@ -1,23 +1,110 @@
...
+    # H: (b_H?, n, n), rho0: (b_rho0?, n, n) -> (y_save, exp_save, meas_save) with
+    #    - y_save: (b_H?, b_rho0?, ntrajs, len(t_save), n, n)
+    #    - exp_save: (b_H?, b_rho0?, ntrajs, len(exp_ops), len(t_save))
+    #    - meas_save: (b_H?, b_rho0?, ntrajs, len(meas_ops), len(t_meas) - 1)
+
+    # check jump_ops
+    if isinstance(jump_ops, list) and len(jump_ops) == 0:
+        raise ValueError(
+            'Argument `jump_ops` must be a non-empty list of tensors. Otherwise,'
+            ' consider using `sesolve`.'
```
gautierronan, line 41:
```suggestion
            ' consider using `ssesolve`.'
```

dynamiqs/smesolve/smesolve.py
```diff
@@ -1,23 +1,110 @@
...
+    jump_ops = formatter.format(jump_ops)  # (len(jump_ops), n, n)
+
+    # convert t_save to a tensor
+    t_save = torch.as_tensor(t_save, dtype=dtype_complex_to_real(dtype), device=device)
+
+    # convert etas to a tensor and check
+    etas = torch.as_tensor(etas, dtype=dtype_complex_to_real(dtype), device=device)
+    if len(etas) != len(jump_ops):
+        raise ValueError('Argument `etas` must have the same length as `jump_ops`.')
+    if torch.all(etas == 0.0):
```
gautierronan, line 61: Can you also check that they are all between 0 and 1?

dynamiqs/smesolve/smesolve.py
```diff
@@ -1,23 +1,110 @@
...
+        device=device,
     )
+
+    # define random number generator from seed
+    generator = torch.Generator(device=device)
+    generator.seed() if seed is None else generator.manual_seed(seed)
+
+    # default options
+    if options is None:
+        raise ValueError()
```
gautierronan, line 84: Can you add a message to the error? As in "`smesolve` does not yet feature a default solver. Please specify one through the `options` argument."

dynamiqs/smesolve/smesolve.py
```diff
@@ -1,23 +1,110 @@
...
+        raise ValueError(
+            'Argument `jump_ops` must be a non-empty list of tensors. Otherwise,'
+            ' consider using `sesolve`.'
+        )
+
+    # format and batch all tensors
+    formatter = TensorFormatter(dtype, device)
+    H_batched, rho0_batched = formatter.format_H_and_state(H, rho0, state_to_dm=True)
+    H_batched = H_batched.unsqueeze(2)
+    rho0_batched = rho0_batched.unsqueeze(2).repeat(1, 1, ntrajs, 1, 1)
```
gautierronan, line 48: Should we write a specific `TensorFormatter` subclass here to avoid the double unsqueeze within this function?

dynamiqs/smesolve/sme_solver.py
```diff
@@ -0,0 +1,110 @@
...
+        y0: Tensor,
+        exp_ops: Tensor,
+        options: Options,
+        *,
+        jump_ops: Tensor,
+        meas_ops: Tensor,
+        etas: Tensor,
+        generator: torch.Generator,
+        t_save_y: Tensor,
+        t_save_meas: Tensor,
```
gautierronan, line 28: I don't really understand how you chose keyword vs non-keyword arguments here. Since its an internal function, I maybe would not bother with the `*` and use only keyword arguments when defining the class instance

gautierronan (changes requested):
Thanks for the nice PR. Here's a first batch of comments. I didn't review any of the code that was related to saving since I figure you'll make changes according to what we discussed about the signal etc.

dynamiqs/smesolve/smesolve.py
pierreguilmin: Meeeh, we could  but it's a bit overkill 🤷‍♂️

dynamiqs/smesolve/sme_solver.py
pierreguilmin: The non-keyword arguments are the one of the base class.

dynamiqs/smesolve/sme_solver.py
pierreguilmin: Because I just jot down the formula, I thought we might want to use inline comments when we didn't take time to write a proper dosctring, and reserve the docstrings for well-written documentation (even for private functions), WDYT?

pierreguilmin:
Sorry for the force push 🙈

dynamiqs/smesolve/sme_solver.py
gautierronan: For me docstring are for function-wide documentation, whereas inline comments are for specifics within the function. But I'm fine with any convention. Also we can just keep our own styles for anything not user-facing. I think that's completely fine the way you did it, I was just wondering.

dynamiqs/smesolve/smesolve.py
```diff
@@ -1,23 +1,105 @@
...
+            ' consider using `ssesolve`.'
+        )
+
+    # format and batch all tensors
+    formatter = TensorFormatter(options.dtype, options.device)
+    H_batched, rho0_batched = formatter.format_H_and_state(H, rho0, state_to_dm=True)
+    H_batched = H_batched.unsqueeze(2)
+    rho0_batched = rho0_batched.unsqueeze(2).repeat(1, 1, ntrajs, 1, 1)
+    # H_batched: (b_H, 1, 1, n, n)
+    # rho0_batched: (b_H, b_rho0, ntrajs, n, n)
```
gautierronan, line 43: Having comments in the middle kinda breaks the flow when reading (it looks like 2 different blocks of code). You can put these at the top below `# format and batch all tensors` (line 37)

dynamiqs/smesolve/smesolve.py
```diff
@@ -1,23 +1,105 @@
...
+        )
+
+    # define the solver
+    args = (H_batched, rho0_batched, exp_ops, options)
+    kwargs = dict(
+        jump_ops=jump_ops,
+        meas_ops=meas_ops,
+        etas=etas,
+        generator=generator,
+        t_save_y=t_save,
```
gautierronan, line 88: A general comment regarding the PR (I'm putting it here but it's PR-wide). 

I don't quite understand why you introduced `t_save_y` on top of the `t_save` we had before. It seems you initialize it to `t_save` everywhere anyway (in `smesolve`, in `mesolve` and in `sesolve`). Plus it complexifies the way you handle saving, and not only for `smesolve` but for all Solver child classes.

If you need a solution specificaly for `smesolve`, I think we should implement a solution that does not alter the other Solvers (unless we'll need it as well at some point, but it doesn't look like it for now).

gautierronan (changes requested):
Some comments about the saving. I'm kinda bothered by this introduction of `t_save_y` which I don't see the need of. Otherwise the PR looks good ! :D

dynamiqs/smesolve/smesolve.py
gautierronan: So in short, if there's a way to simplify this whole saving workflow, I think it'd be nice and might avoid us problems later on.

pierreguilmin: `t_save` defined by the `Solver` base class is "the times when smthg has to be saved", and then we have `t_save_y` to save the state and exp_ops, and `t_save_meas` to save the measurement record (for example for the SME, `t_save = torch.cat(t_save_y, t_save_meas)`). This is necessary to
- specify correct tensor dimensions when we initialize y_save and exp_save
- differentiate when we save the state vs when we save the measurement record

Ideally, we would have defined a base solver class with a `t_save` that doesn't know what to save, just when something has to be saved. Then we would inherit to define the logic to save states `t_save_y`, and inherit again to define the logic to save measurement record `t_save_meas`. The confusion comes because we directly implement the saving of the state and exp_ops in the base solver class. If you prefer I can write this base class that just knows about `t_save` and knows nothing about the problem, but I thought it's a bit overkill.

pierreguilmin: If you have a better idea I'd be happy to try! But I think the best solution is to make the `Solver` base class a bit more generic concerning saving.

gautierronan: Ok thanks for the clarification it's clearer ! I didn't understand that you meant `t_save` for any save time, and I thought `t_save_y` was just the times at which we save the state (and not the exp_ops). 

Maybe what I would propose is the following:

- keep the overall implementation you propose
- rename `t_save_y` to `t_save` as we had before. Currently, `t_save` is a different object when passed by users and within the code. Also, the `_y` was misleading IMO.
- rename `t_save` to `t_allsave` (or similar) to account for all save times (i.e. the concatenation of `t_save` and `t_meas`).
- in the arguments of the `Solver` base class, pass only `t_save`, and add a method `init_t_allsave` (or similar) that computes the `t_allsave` object from `t_save` and `t_meas`. There will be one such method in the base class, and one in the `SMESolver` class.
- in the definition of `SMESolver._save`, avoid if possible the `if t in self.t_save_y` lines by using a counter and some pre-processing. It's nice not to search through the object, and we should know in advance when to save y and when to save the measurements. It's a bit faster and a bit more robust.

What do you think?

pierreguilmin: I tried something, it's quite complicated but I don't know how we could simplify it 😅 Any suggestions welcome!

dynamiqs/mesolve/mesolve.py
```diff
@@ -92,6 +93,7 @@ def mesolve(
 
     # convert t_save to a tensor
     t_save = torch.as_tensor(t_save, dtype=options.rdtype, device=options.device)
+    check_time_array(t_save, 't_save')
```
gautierronan, line 96:
```suggestion
    check_time_array(t_save, error_str='t_save')
```
So it reads without having to check the function. Can be `error_str`, `arg_name`, `raise_str`, `str` or whatever you prefer.
Also applies to other calls.

dynamiqs/solvers/ode/fixed_solver.py
```diff
@@ -27,30 +27,30 @@ def run_autograd(self):
...
         """
-        # assert that `t_save` values are multiples of `dt`
+        # assert that `t_stop` values are multiples of `dt`
         if not torch.allclose(
-            torch.round(self.t_save / self.dt), self.t_save / self.dt
+            torch.round(self.t_stop / self.dt), self.t_stop / self.dt
         ):
             raise ValueError(
-                'For fixed time step solvers, every value of `t_save` must be a'
+                'For fixed time step solvers, every value of `t_stop` must be a'
```
gautierronan, line 35: It's a bit annoying to raise an error with an argument that the user does not know about. I think we should keep `t_save` here, and ideally, also check separately for `t_meas` and raise a similar error. Or otherwise we just say `t_save` and `t_meas` in the same error message, specifying that `t_meas` is only for the SME.

gautierronan: I'm being a bit picky here, but I think this is an error message that many users will see at least once.

gautierronan (changes requested):
Thank you very much for the PR and for the many back and forth iterations ! I really like it this way. Great job !

pierreguilmin:
I merged main and adapted the measure class
```text
==== Result ====
Options    : Euler
Start      : 2023-07-03 18:27:10
End        : 2023-07-03 18:28:33
Total time : 83.73 s
states     : Tensor (3, 4, 1001, 11, 8, 8) | 64.52 Mb
expvals    : Tensor (3, 4, 1001, 2, 11) | 2.02 Mb
measvals   : Tensor (3, 4, 1001, 2, 100) | 9.16 Mb
```

pierreguilmin:
Wdyt about renaming the aliases `expvals` to `exp` and `measvals` to `meas`?

gautierronan:
> Wdyt about renaming the aliases `expvals` to `exp` and `measvals` to `meas`?

Yes I don't really like `measvals` either.

How about something like that (in the singular) ?
```python
======== Result ========
Method      : Euler
Start       : 2023-07-03 18:27:10
End         : 2023-07-03 18:28:33
Total time  : 83.73 s
state       : Tensor (3, 4, 1001, 11, 8, 8) | 64.52 Mb
expect      : Tensor (3, 4, 1001, 2, 11) | 2.02 Mb
measurement : Tensor (3, 4, 1001, 2, 100) | 9.16 Mb
```
I think it's qutip conventions. But whatever you prefer !

pierreguilmin:
I went for states, expects and measurements (the singular sounds a bit weird to me in this case 🤔). Anyway, we can take some time to rehaul all the user-facing namings some day to look for improvements.

gautierronan (approved):

PR #146, opened 2023-06-20
Performance and notation improvements for the adaptive solver

gautierronan:
Follows notations and workflow of scipy more closely (see [scipy/integrate/_ivp/rk.py](https://github.com/scipy/scipy/blob/c1ed5ece8ffbf05356a22a8106affcd11bd3aee0/scipy/integrate/_ivp/rk.py#L102)).

Also, moves from `torch.einsum` to `torch.tensordot` for performance. With this simple benchmark, I saw a systematic improvement (sometimes quite a lot for smaller matrices) both on cpu and gpu:
```python
import torch
import timeit

M = 10
N = 1000
device = 'cpu'

c = torch.randn(M, dtype=torch.complex64).to(device)
k = torch.randn(M, N, N, dtype=torch.complex64).to(device)

print(torch.allclose(torch.einsum('b,b...', c, k), torch.tensordot(c, k, dims=([0], [0]))))
%timeit torch.einsum('b,b...', c, k)
%timeit torch.tensordot(c, k, dims=([0], [0]))
```

### DIFF
--- a/dynamiqs/options.py
+++ b/dynamiqs/options.py
@@ -99,7 +99,7 @@ class ODEAdaptiveStep(Options):
         atol: float = 1e-8,
         rtol: float = 1e-6,
         max_steps: int = 100_000,
-        factor: float = 0.9,
+        safety_factor: float = 0.9,
         min_factor: float = 0.2,
         max_factor: float = 5.0,
         **kwargs,
@@ -108,7 +108,7 @@ class ODEAdaptiveStep(Options):
         self.atol = atol
         self.rtol = rtol
         self.max_steps = max_steps
-        self.factor = factor
+        self.safety_factor = safety_factor
         self.min_factor = min_factor
         self.max_factor = max_factor
 
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -220,7 +220,7 @@ class DormandPrince5(AdaptiveSolver):
         alpha, beta, csol, cerr = self.tableau
 
         # compute iterated Runge-Kutta values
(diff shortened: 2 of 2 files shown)


pierreguilmin (approved):

pierreguilmin:
I guess there are a bunch of other places where we could use `torch.tensordot` rather than `torch.einsum`?

pierreguilmin:
Actually can't we use `torch.matmul`?

pierreguilmin:
For information, the benchmark yields exactly the same result for `einsum` and `tensordot` on my machine.

pierreguilmin:
Wow, interestingly chatGPT proposes
```
c.view(-1, 1, 1) * k
```
which is four times faster on my machine 🤯

pierreguilmin:
I'm a little confused by the code, don't we want to do element-wise multiplication rather than einsum? Something like `c.view(-1, 1, 1) * k`?

gautierronan:
So yes we may be able to use tensordot in some places. I think the point is that tensordot is straight to the point, whereas einsum tries to optimize the tensor contraction in a smart way. However it cannot do miracles when there are no miracles to be made, in which case it's probably better to use tensordot. We should maybe benchmark our `kraus_map` function in particular.

But there are tons of different tensor contraction functions in torch (tensordot, einsum, matmul, bmm, ...), and it seems like they're all advantageous in their own regime of sizes.

In this case, it seems `tensordot` is the fastest for me, but very close to einsum. Your other proposal with the view is much slower however. The proper call is `(c.view(-1, 1, 1) * k).sum(0)`. Any way, it's not really a big performance gain because we're mostly limited by the call to `lindbladian` and not by these small tensor contractions.

PR #147, opened 2023-06-20
Check propagator Hamiltonian type

gautierronan:
Fixes https://github.com/dynamiqs/dynamiqs/issues/88

### DIFF
--- a/dynamiqs/solvers/propagator.py
+++ b/dynamiqs/solvers/propagator.py
@@ -3,10 +3,18 @@ from abc import abstractmethod
 from torch import Tensor
 
 from ..utils.progress_bar import tqdm
+from ..utils.td_tensor import ConstantTDTensor
 from .solver import AutogradSolver
 
 
 class Propagator(AutogradSolver):
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


dynamiqs/solvers/propagator.py
```diff
@@ -3,10 +3,18 @@
...
 from .solver import AutogradSolver
 
 
 class Propagator(AutogradSolver):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # check that Hamiltonian is time-independent
+        if not isinstance(self.H, ConstantTDTensor):
+            raise TypeError('Propagator solvers require time-independent Hamiltonians.')
```
pierreguilmin, line 16:
```suggestion
            raise TypeError('Propagator solvers require a time-independent Hamiltonian.')
```

pierreguilmin: (maybe you used the `s` for the batching, but I think we should ignore the batching when referring to the Hamiltonian or state in written sentences, things are clearer this way)

pierreguilmin (dismissed):
Thanks!

PR #150, opened 2023-06-22
Separate public and solver-related utilities in separate folders

gautierronan:


dynamiqs/solvers/utils/solver_utils.py
pierreguilmin, line 1: We could rename this file `utils.py` (as we do for the `utils.py` file of the `dynamiqs/utils/` folder.

dynamiqs/solvers/utils/progress_bar.py
pierreguilmin, line 1: I guess we could merge this file with `solver_utils.py` directly as it just contains two lines.

pierreguilmin (changes requested):
Nice!

pierreguilmin (approved):

PR #152, opened 2023-06-22
Add table that summarizes all functions in documentation API

gautierronan:
(with clickable links that bring to the appropriate function definition)

Screenshot 2023-06-23 at 13 51 37


docs/templates/python/material/module.html
pierreguilmin, line 1: When copy-pasting templates that you then modify, it'd be nice to have a commit where you just copy-paste the template, and then a subsequent commit where we see what you modify. It's a bit hard to track what is changed here.

dynamiqs/utils/_wigner.py
pierreguilmin, line 1: I don't think we should perform this move, it's better structured like it was done, and then we can set the API the way we prefer.

pierreguilmin (changes requested):
IMO this is not how we should document the API. When documenting an API the table of content should follow closely the package structure, as was done before. It's a part of the documentation that is meant for developers, not users. I'm guessing you're copying what's done in PyTorch doc, but it's a very different package with tons of functions (thus they have to structure the API somehow otherwise it's unreadable), but it's not common to do it like this, especially for small packages.

I like the table however, I think it would be better placed in a getting started page "Utils function" which gathers all utils functions we have for a quick lookup.

docs/templates/python/material/module.html
```diff
@@ -0,0 +1,67 @@
...
+      {% endwith %}
+
+      {% with obj = module %}
+        {% set root = False %}
+        {% set heading_level = heading_level + 1 %}
+        {% if config.table %}
+          {% include "table.html" with context %}
+        {% else %}
+          {% include "children.html" with context %}
+        {% endif %}
```
gautierronan, line 62: NB: Here is the modification compared to the copy-pasted file.

docs/templates/python/material/module.html
gautierronan: Sorry ! I knew you'd say that but I didn't know how to backtrack through my commits ^^. I essentially just modified lines 58 > 62.

dynamiqs/utils/_wigner.py
gautierronan: Yes I completely agree. But how can we make the functions appear on the same page with mkdocs? It seems kinda silly to have a separate page just for a single function `dq.wigner` no?

gautierronan:
> IMO this is not how we should document the API. When documenting an API the table of content should follow closely the package structure, as was done before. It's a part of the documentation that is meant for developers, not users. I'm guessing you're copying what's done in PyTorch doc, but it's a very different package with tons of functions (thus they have to structure the API somehow otherwise it's unreadable), but it's not common to do it like this, especially for small packages.
I like the table however, I think it would be better placed in a getting started page "Utils function" which gathers all utils functions we have for a quick lookup.

Yes I really like having a table that summarizes all available functions (really missing in the qutip doc IMO). I don't really care where it is, but it should be somewhere. Let's just keep the table implementation in this PR, and think later about how we structure the doc.

dynamiqs/utils/_wigner.py
pierreguilmin: No, the API is not a documentation for the user, it's a documentation for the developers, so it should mimic the package structure.

mkdocs.yml
```diff
@@ -80,7 +81,11 @@ nav:
...
+  - Python API:
       - reference/index.md
-      - dynamiqs.utils:
-          - utils: reference/utils/utils.md
+      - Utility functions:
+          - Initializing states: reference/utils/states.md
+          - Initializing operators: reference/utils/operators.md
+          - Operating on states: reference/utils/utils.md
+          - Converting tensors: reference/utils/tensor_types.md
+          - Wigner distributions: reference/utils/wigner.md
```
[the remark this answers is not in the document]
pierreguilmin: To be clear, we will make a user-facing documentation with all the titles you propose, but it's not what's typically done for the API documentation.

pierreguilmin (approved):

PR #153, opened 2023-06-23
Document all remaining utility functions

gautierronan:


dynamiqs/utils/operators.py
```diff
@@ -24,44 +24,87 @@
...
+    """
     return torch.tensor([[0.0, 0.0], [1.0, 0.0]], dtype=dtype, device=device)
 
 
 @complex_tensor
 def eye(*dims: int, dtype=None, device=None) -> Tensor:
-    """Identity operator."""
+    """Returns the identity operator.
+
+    Args:
```
abocquet, line 94: Don't you want to document `dims`?

abocquet (approved):

dynamiqs/utils/operators.py
gautierronan: Good catch ! Simply forgot about it :D

abocquet (approved):

dynamiqs/utils/wigner.py
```diff
@@ -19,16 +19,18 @@ def wigner(
...
     method: Literal['clenshaw', 'fft'] = 'clenshaw',
 ) -> tuple[Tensor, Tensor, Tensor]:
-    """Compute the wigner distribution of a state vector or density matrix.
+    """Compute the wigner distribution of a ket or density matrix.
 
     Args:
-        state: State vector or density matrix.
-        x_max: Maximum value of x for which to compute the wigner distribution.
-        p_max: Maximum value of p for which to compute the wigner distribution.
+        state (..., n, 1) or (..., n, n): State vector or density matrix.
```
pierreguilmin, line 25:
```suggestion
        state (..., n, 1) or (..., n, n): Ket or density matrix.
```

dynamiqs/utils/wigner.py
```diff
@@ -19,16 +19,18 @@ def wigner(
...
+        x_max (float): Maximum value of x for which to compute the wigner distribution.
+            Defaults to `2 * pi`.
```
pierreguilmin, line 27: This feels very weird to me, the wigner values for each pixel range between -2/pi and 2/pi, but why do we default the Wigner border to such values? It should be a coherent state amplitude IMO, e.g. `6.0`.

dynamiqs/utils/wigner.py
```diff
@@ -19,16 +19,18 @@ def wigner(
...
+        p_max (float): Maximum value of p for which to compute the wigner distribution.
             If the wigner distribution is computed using the `fft` method, `p_max` is
-            ignored, and given by `2 * pi / x_max` instead.
-        n_pixels: Number of pixels in each direction.
-        method: Method used to compute the wigner distribution.
+            ignored, and given by `2 * pi / x_max` instead. Defaults to `2 * pi`.
+        n_pixels (int): Number of pixels in each direction. Defaults to 200.
```
pierreguilmin, line 31: We use `num` in our conventions:
```suggestion
        num_pixels (int): Number of pixels in each direction. Defaults to 200.
```

dynamiqs/utils/wigner.py
pierreguilmin, line 27:
```suggestion
        x_max: Maximum value of x.
```
- The default appears in the function signature, no need to say it except if it has a strange value/behaviour.
- The type appears in the function signature, we just use it for tensors to specify their dimension.

dynamiqs/utils/wigner.py
```diff
@@ -19,16 +19,18 @@ def wigner(
...
+        method (str): Method used to compute the wigner distribution. Available
+            methods: `clenshaw` or `fft`. Defaults to `clenshaw`.
 
     Returns:
         A tuple `(xvec, pvec, w)` where
```
pierreguilmin, line 36: The list doesn't render properly, insert newline, no indentation and `-` before items.

dynamiqs/utils/wigner.py
pierreguilmin: This applies to the whole PR, we can remove the type between parentheses in the docstring if it' no different from the function signature, and we can remove the indication of default values.

dynamiqs/utils/states.py
pierreguilmin, line 1: We can remove all examples from this file, they don't really add much.

dynamiqs/utils/operators.py
pierreguilmin, line 1: I think we can remove the examples here too.

pierreguilmin: It'd be nice to add LaTeX formulas for all these operators (but we can do it in another PR).

dynamiqs/utils/operators.py
```diff
@@ -24,44 +24,93 @@
...
 @complex_tensor
 def sigmax(*, dtype=None, device=None) -> Tensor:
-    """Pauli $X$ operator."""
+    """Returns the Pauli $X$ operator.
+
+    Args:
+        dtype (torch.dtype): Data type of the returned tensor. Defaults to the complex
+            data type specified by `torch.get_default_dtype()`.
+        device (torch.device): Device of the returned tensor. Defaults to the globally
+            defined default device.
```
pierreguilmin, line 33:
```suggestion
        dtype: Data type of the returned tensor.
        device: Device of the returned tensor.
```
I'd suggest shortening the doc to this in order not to overload it, we can write smwh clearly in a "must read" section what all dtype and device mean and how they default, instead of repeating these two lines everywhere.

pierreguilmin (changes requested):
Nice, sorry for the confusion with the types and all, I wasn't clear at the time on what should be specified.

I'm not done with the review but there's already quite some changes so I'll stop here.

dynamiqs/utils/wigner.py
gautierronan: This is because of the `fft` method. When computing the wigner through the fft, the range in P is related to the range in X through a Fourier transform. So if you want a square range, the range in both directions should be [-2pi, 2pi]. Otherwise one direction is squeezed with respect to the other.

For the default method `clenshaw`, this is however not required. One can compute the Wigner in any desired range.

dynamiqs/utils/states.py
gautierronan: Don't really agree. It's nice for people who might not know exactly what these states are, and also showcases the various ways to call the function (e.g. `dq.fock_dm(3, 1)` and `dq.fock_dm((3, 2), (1, 0))`).

dynamiqs/utils/wigner.py
```diff
@@ -16,31 +16,32 @@ def wigner(
     state: Tensor,
     x_max: float = 6.2832,
     p_max: float = 6.2832,
-    n_pixels: int = 200,
+    num_pixels: int = 200,
     method: Literal['clenshaw', 'fft'] = 'clenshaw',
 ) -> tuple[Tensor, Tensor, Tensor]:
-    """Compute the wigner distribution of a state vector or density matrix.
+    """Compute the wigner distribution of a ket or density matrix.
```
pierreguilmin, line 22:
```suggestion
    """Compute the Wigner distribution of a ket or density matrix.
```

pierreguilmin: And also in the rest of the docstring.

dynamiqs/utils/wigner.py
```diff
@@ -16,31 +16,32 @@ def wigner(
...
-        p_max: Maximum value of p for which to compute the wigner distribution.
-            If the wigner distribution is computed using the `fft` method, `p_max` is
-            ignored, and given by `2 * pi / x_max` instead.
-        n_pixels: Number of pixels in each direction.
-        method: Method used to compute the wigner distribution.
+        state (..., n, 1) or (..., n, n): Ket or density matrix.
+        x_max: Maximum value of x.
+        p_max: Maximum value of p. Not used if the wigner distribution is
+            computer with the `fft` method, in which case `p_max` is given by
+            `2 * pi / x_max`.
```
pierreguilmin, line 29:
```suggestion
        p_max: Maximum value of p. Ignored if the wigner distribution is
            computed with the `fft` method, in which case `p_max` is given by
            `2 * pi / x_max`.
```

dynamiqs/utils/tensor_types.py
```diff
@@ -130,14 +129,16 @@ def dtype_real_to_complex(
...
 @complex_tensor
-def from_qutip(x: Qobj, *, dtype=None, device=None) -> Tensor:
+def from_qutip(
+    x: Qobj, *, dtype: torch.dtype | None = None, device: torch.device | None = None
+) -> Tensor:
     """Convert a QuTiP quantum object to a PyTorch tensor.
 
     Args:
-        x: QuTiP quantum object.
+        x: Input quantum object.
```
pierreguilmin, line 138: Missing `dtype` and `device` arguments.

dynamiqs/utils/tensor_types.py
```diff
@@ -130,14 +129,16 @@ def dtype_real_to_complex(
 
 
 @complex_tensor
-def from_qutip(x: Qobj, *, dtype=None, device=None) -> Tensor:
+def from_qutip(
+    x: Qobj, *, dtype: torch.dtype | None = None, device: torch.device | None = None
```
pierreguilmin, line 133: In the case where we use `@complex_tensor` decorator you want to use`dtype: torch.complex64 | torch.complex128 | None = None`. BUt it's quite length, this is why we removed the type in a PR a while ago for functions decorated with `@complex_tensor`. We can do as you prefer, either fix the typing everywhere, or remove the typing everywhere and just specify that it takes a complex data type in the dosctring description of `dtype`.

dynamiqs/utils/states.py
```diff
@@ -34,6 +34,15 @@ def fock(
...
                 [0.+0.j]], dtype=torch.complex128)
+
+    Args:
+        dims: Dimension of the Hilbert space of each mode.
+        states: Fock state of each mode.
+        dtype: Data type of the returned tensor.
+        device: Device of the returned tensor.
+
+    Returns:
+        torch.Tensor (n, 1): Ket of the Fock state or tensor product of Fock states.
```
pierreguilmin, line 45:
```suggestion
    (n, 1): Ket of the Fock state or tensor product of Fock states.
```

pierreguilmin: I should clarify how I think we should handle types:
- By default: in the code, we only write the type in the function signature, never in the docstring. In the doc website, the types only appear in the function signature.
- IF one of the argument or the returned object is a tensor, in the code we add the tensor dimension to the dosctring between parentheses. In the doc website it appears in the `Args` or `Returns` section next to the argument name.
- IF we want to write a custom type description because the function signature is a bit obscure (like `OperatorLike | List[OperatorLike] | None`), in the code we add the custom type description to the dosctring between parentheses. In the doc website it appears in the `Args` or `Returns` section next to the argument name.

Note that the doc is a bit bugged now, because
- The function signature is not always correct.
- The typing also appears in the `Args` section even if we didn't specify a custom type in the docstring.

What do you think of these conventions? It's fine for this PR, I'll fix the doc after we've merged.

dynamiqs/utils/operators.py
pierreguilmin, line 1: Use `Examples` instead of `Example` in this file too.

pierreguilmin (changes requested):
Nice!! 🥳 Minor modifications.

dynamiqs/utils/tensor_types.py
gautierronan: No I had to had it back because it was not appearing in the documentation actually. If you remove it and try to build to doc locally, you'll see that type hints are gone.

pierreguilmin (approved):
It's exciting to see the doc growing 🤩

PR #154, opened 2023-06-26
Fix Rouchon1_5 and implement caching

### LINKED ISSUE #78: Fix `inv_sqrtm` (used by `MERouchon1_5`)
https://github.com/pierreguilmin/torchqdynamics/blob/5d797a4ca199af16f0689fade49046a4452fba07/torchqdynamics/mesolve/rouchon.py#L107-L108

gautierronan:
Fixes #78.

### DIFF
--- a/dynamiqs/mesolve/rouchon.py
+++ b/dynamiqs/mesolve/rouchon.py
@@ -7,6 +7,7 @@ from torch import Tensor
 
 from ..solvers.ode.fixed_solver import AdjointFixedSolver
 from ..utils.solver_utils import cache, inv_sqrtm, kraus_map
+from ..utils.td_tensor import CallableTDTensor, ConstantTDTensor
 from ..utils.utils import trace
 from .me_solver import MESolver
 
@@ -61,7 +62,6 @@ class MERouchon1(MERouchon):
 
         Args:
             t: Time.
-            dt: Time step.
             rho: Density matrix of shape `(b_H, b_rho, n, n)`.
             phi: Adjoint state matrix of shape `(b_H, b_rho, n, n)`.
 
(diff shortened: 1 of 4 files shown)


dynamiqs/mesolve/rouchon.py
```diff
@@ -91,13 +91,27 @@ def backward_augmented(self, t: float, rho: Tensor, phi: Tensor):
...
 
-        Note:
-            No need for trace renormalization since the scheme is trace-preserving
-            by construction.
+        # compute the inverse square root renormalization matrix
+        # (for time-dependent Hamiltonians, exclude the Hamiltonian from normalization)
+        if isinstance(self.H, ConstantTDTensor):
+            H_nh = self.H_nh(0.0)  # (b_H, 1, n, n)
+        elif isinstance(self.H, CallableTDTensor):
+            H_nh = -0.5j * self.sum_no_jump  # (b_H, 1, n, n)
```
pierreguilmin, line 102: Where did you see this? Are you sure it's ok to exclude the Hamiltonian from the normalisation?

pierreguilmin (changes requested):
Verrryyy nice thanks a lot!!

dynamiqs/mesolve/rouchon.py
gautierronan: Yes essentially it's better than nothing. Recomputing the normalization at every time step would be extremely costly, and it's often the case that numerical divergence is caused by the Ldag @ L terms that explode as the Fock truncature gets larger.

gautierronan:
NB: We might want to merge `Rouchon1` and `Rouchon1_5` at some point because they share the exact same `forward` and `backward_augmented` implementations. The only difference is the init. Plus, they're actually both order 1 methods, so we could just add an option to not add the square root normalization, or even completely remove the current `Rouchon1` method.

dynamiqs/solvers/utils/utils.py
```diff
@@ -45,7 +45,8 @@ def inv_sqrtm(mat: Tensor) -> Tensor:
          https://github.com/pytorch/pytorch/issues/25481#issuecomment-584896176.
     """
```
pierreguilmin, line 46:
```diff
- """Compute the inverse square root of a matrix using its eigendecomposition.
+ """Compute the inverse square root of a complex Hermitian or real symmetric matrix
+    using its eigendecomposition.
```

dynamiqs/mesolve/rouchon.py
```diff
@@ -91,13 +88,35 @@ def backward_augmented(self, t: float, rho: Tensor, phi: Tensor):
...
-        Note:
-            No need for trace renormalization since the scheme is trace-preserving
-            by construction.
+        # compute the inverse square root renormalization matrix
+        # (for time-dependent Hamiltonians, exclude the Hamiltonian from normalization)
+        # H_nh, M0, M0rev, self.S, self.Srev: (b_H, 1, n, n)
+        if isinstance(self.H, ConstantTDTensor):
+            H_nh = self.H_nh(0.0)
+        elif isinstance(self.H, CallableTDTensor):
+            H_nh = -0.5j * self.sum_no_jump
```
pierreguilmin, line 100: Can we call this `H_nh_tmp` as it is not the real `H_nh`?

pierreguilmin (approved):
Minor comments.

> NB: We might want to merge Rouchon1 and Rouchon1_5 at some point because they share the exact same forward and backward_augmented implementations. The only difference is the init. Plus, they're actually both order 1 methods, so we could just add an option to not add the square root normalization, or even completely remove the current Rouchon1 method.

I'm down to merge them (here or in another PR), but we want to leep keep the current Rouchon1 method for two reasons:
- to compare when benchmarking (and understand how imptt is the normalisation)
- for sparse matrices, the renormalization destroys the sparness

pierreguilmin (approved):

PR #155, opened 2023-07-02
Add result class

pierreguilmin:
First basic proposal for a result class, here's how the printing looks:
```python
>>> result = dq.mesolve(...)
>>> print(result)
==== Result ===
Options    : Rouchon1
Start      : 2023-07-02 15:12:13
End        : 2023-07-02 15:12:13
Total time : 0.45 s
y_save     : tensor (4, 101, 64, 64) [12.62 Mb]
exp_save   : tensor (4, 2, 101) [6.31 Kb]
```

There's a lot to add, here's a quick list on the top of my head:
- [ ] Implement `to_qutip()` and `to_numpy()` returning a new object where we converted `y_save` and `exp_save` to a list of `Qobj`/numpy array or just big numpy arrays.
- [ ] Implement `save()` and `load()`: we need to save dynamiqs current version, the commit version (+ wip if needed) and a version saving scheme (we raise an error if we try to load an incompatible version). I think it would also be nice to have a hash for each result object, e.g. if someone wants to use them as index in a database.
- [ ] Implement solver specific savings (e.g. for adaptive solvers the number of iterations).
- [ ] Implement a `__str__` method for `Options` base class and child classes.

Moreover, we should probably choose other name for the api than `y_save` and `exp_save` (I think they are good internally though), we can just set aliases on the result class for the uses, what do you think about `states` and `expects`?

I'd like to rebase #141 on this once we agree on the structure.

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import torch
-from torch import Tensor
 
 from ..options import Dopri5, Euler, Options, Rouchon1, Rouchon1_5, Rouchon2
 from ..solvers.utils.tensor_formatter import TensorFormatter
+from ..solvers.result import Result
 from ..utils.tensor_types import OperatorLike, TDOperatorLike, TensorLike
 from .adaptive import MEDormandPrince5
 from .euler import MEEuler
@@ -19,7 +19,7 @@ def mesolve(
     *,
     exp_ops: OperatorLike | list[OperatorLike] | None = None,
     options: Options | None = None,
-) -> tuple[Tensor, Tensor | None]:
+) -> Result:
     """Solve the Lindblad master equation for a Hamiltonian and set of jump operators.
 
     The Hamiltonian `H` and the initial density matrix `rho0` can be batched over to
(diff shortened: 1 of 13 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -62,15 +62,7 @@ def mesolve(
...
-        A tuple `(rho_save, exp_save)` where
-            `rho_save` is a tensor with the computed density matrices at `t_save`
-                times, and of shape `(len(t_save), n, n)` or `(b_H, b_rho, len(t_save),
-                n, n)` if batched. If `save_states` is `False`, only the final density
-                matrix is returned with the same shape as the initial input.
-            `exp_save` is a tensor with the computed expectation values at `t_save`
-                times, and of shape `(len(exp_ops), len(t_save))` or `(b_H, b_rho,
-                len(exp_ops), len(t_save))` if batched. `None` if no `exp_ops` are
-                passed.
+        Simulation result.
```
gautierronan, line 65:
```suggestion
        Result of the master equation integration.
```

dynamiqs/solvers/result.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def memory_str(x: Tensor) -> str:
+    mem = memory_bytes(x)
+    if mem < 1024**2:
+        return f'{mem / 1024:.2f} Kb'
+    elif mem < 1024**3:
+        return f'{mem / 1024**2:.2f} Mb'
+
+    return f'{mem / 1024**3:.2f} Gb'
```
gautierronan, line 21:
```suggestion
    else:
        return f'{mem / 1024**3:.2f} Gb'
```
Personal preference, a bit more readable :D

dynamiqs/solvers/result.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def tensor_str(x: Tensor) -> str:
+    return f'tensor {tuple(x.size())} [{memory_str(x)}]'
```
gautierronan, line 25:
```suggestion
    return f'Tensor {tuple(x.size())} | {memory_str(x)}'
```
Just a suggestion, I like it a tiny bit more:
```python
y_save     : Tensor (4, 101, 64, 64) | 12.62 Mb 
exp_save   : Tensor (4, 2, 101) | 6.31 Kb
```

dynamiqs/solvers/result.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @property
+    def total_time(self) -> timedelta | None:
+        if self.start_datetime is None or self.end_datetime is None:
+            return None
+        return self.end_datetime - self.start_datetime
+
+    def __str__(self):
+        tmp = (
+            '==== Result ===\n'
```
gautierronan, line 60:
```suggestion
            '==== Result ====\n'
```

gautierronan (changes requested):
Very nice PR ! A really like it. Some minor comments below

gautierronan:
> Moreover, we should probably choose other name for the api than `y_save` and `exp_save` (I think they are good internally though), we can just set aliases on the result class for the uses, what do you think about `states` and `expects`?

Yes I like `states` and maybe `expvals` ?

gautierronan (approved):

PR #157, opened 2023-07-05
Separate doc page for each function + summary tables

gautierronan:
Kind of a big PR with a proposal for the Python API part of the documentation. IMO it's going in the right direction, but might need some polishing still. Tell me what you think !

Tables:

Screenshot 2023-07-05 at 03 05 37

>

Individual function pages:

Screenshot 2023-07-05 at 03 05 46


dynamiqs/utils/operators.py
pierreguilmin, line 1: IMO examples should precede the `Args` and `Returns` sections, one usually don't need to read them if the examples are well written.

dynamiqs/utils/states.py
```diff
@@ -22,6 +22,16 @@ def fock(
...
 
+    Args:
+        dims: Dimension of the Hilbert space of each mode.
+        states: Fock state of each mode.
+        dtype: Data type of the returned tensor.
+        device: Device of the returned tensor.
+
+    Returns:
+        Tensor of size _(n, 1)_ : Ket of the Fock state or tensor product of Fock
+            states.
```
pierreguilmin, line 33: I'd suggest reverting to the previous format.

pierreguilmin (changes requested):
First let me say I appreciate a lot the effort you put into making the doc looks nice and being useful to use! Concerning this PR, I'm not a fan of the one page per function for a few reasons:
- Tons of clicks to explore the various functions, while before you could just scroll through a single page to read them all.
- We tamper quite a lot with mkdocs to generate a custom file tree. This is a bit tedious to maintain and will be annoying when using other mkdocs-related packages that expect "common" mkdocs file tree and format. For example, it doesn't play well with `mkdocs docserve`, for some reason the website freezes after a few clicks. But I expect us to encounter many more issues if we go this custom by implementing our own parsing.

I agree it looks a tiny bit better visually, but I don't think it's worth the trouble. If you reaaaaally prefer it strongly we can keep the current changes, but I'd prefer to revert. Again, this part of the documentation is meant for developers not users, and I think the more closely it looks like the package (e.g. all the functions of a file are gathered in the same page) the easier it is to navigate for developers.

Finally, I think for big changes that are a little subjective for the doc we should discuss beforehand, to agree on what we commonly like. Otherwise we might loose a lot of time implementing things that we don't agree on 😋

dynamiqs/utils/operators.py
gautierronan: I like having Examples last because they don't always appear in a function documentation. Whereas Args and Returns always appear. This way, the order is always kept the same and its easier to follow from one function to the next.

dynamiqs/utils/states.py
gautierronan: Is it fine if this is in a separate PR? I did some doc updates in this one, but it's starting to be messy (sorry for that).

gautierronan:
What it now looks like. @pierreguilmin can you check the way "generate_reference.py" works and the way the doc is organized is fine with you, and then we can push? Then we can do all the fine-tuning related to documentation, section names, etc. in a separate PR. This one is becoming messy.

Screenshot 2023-07-09 at 10 26 23

pierreguilmin:
Ok, let's merge this one and iterate on it.

pierreguilmin (approved):

PR #166, opened 2023-07-13
Refactor tests

pierreguilmin:
Merge of various PRs:
- Add analytic tests for expectation values
- Add analytic tests for autograd and adjoint gradient
- Refactor tests base class


pierreguilmin:
@gautierronan I merged all PRs into this one to make the review easier.

tests/solver_tester.py
```diff
@@ -0,0 +1,111 @@
...
+            result.exp_save,
+            system.expects(t_save),
+            rtol=exp_save_rtol,
+            atol=exp_save_atol,
+        )
+
+    def test_correctness(self):
+        pass
+
+    def _test_gradient(
```
gautierronan, line 74: Can you separate this function in two subfunctions for readability? A first one `_test_gradient_y` and a second one `_test_gradient_exp`, and then wrap them both inside your current `_test_gradient`.

tests/mesolve/test_rouchon.py
```diff
@@ -1,48 +1,60 @@
...
+        options = dq.options.Rouchon1(dt=1e-3, sqrt_normalization=sqrt_normalization)
+        self._test_correctness(options, leaky_cavity_8, num_t_save=11)
+
+    @pytest.mark.parametrize('sqrt_normalization', [False, True])
+    def test_autograd(self, sqrt_normalization):
+        options = dq.options.Rouchon1(
+            dt=1e-3, sqrt_normalization=sqrt_normalization, gradient_alg='autograd'
+        )
+        self._test_gradient(
+            options, grad_leaky_cavity_8, num_t_save=11, rtol=1e-2, atol=1e-2
```
gautierronan, line 25: Could this pass with lower `rtol` or `atol`? It seems quite high right now.

tests/solver_tester.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _test_correctness(
+        self,
+        options: Options,
+        system: System,
+        *,
+        num_t_save: int,
+        y_save_norm_atol: float = 1e-2,
+        exp_save_rtol: float = 1e-2,
+        exp_save_atol: float = 1e-2,
```
gautierronan, line 54: The default are a bit high maybe? Are we sure to catch an error if there's one? Maybe it's already good enough I'm not sure.

tests/mesolve/test_rouchon.py
gautierronan: (but don't reduce `dt` it's already good enough if the test can't do better)

gautierronan (changes requested):
Great PR and great work ! One small comment about readability, and then some about tolerences. I'm always scared they're too large and we're not catching errors haha.

tests/solver_tester.py
pierreguilmin: I'm not sure it's a good idea as (1) both part of the function share common variables (`system`, `t_save` and `result`) that we'd have to pass around, and (2) we don't want these subfunctions to be called individually. I guess the comments with `===` are enough to separate both parts no?

tests/mesolve/test_rouchon.py
pierreguilmin: I lower all the rtol and atol to their min values in next commit 😉

tests/solver_tester.py
pierreguilmin: > I lower all the rtol and atol to their min values in next commit 😉

pierreguilmin:
Wew, finally 😅

pierreguilmin:
It's a little weird that the `Rouchon2` performs bad no? 🤔

gautierronan:
> It's a little weird that the `Rouchon2` performs bad no? 🤔

Yes I think there's a bug. By reducing to `dt = 1e-4` I get a worse performance than `dt = 1e-3`. I'm pretty sure it used to work so probably a bug introduced in a recent PR? Rouchon1 seems correct though because reducing `dt` by a factor of 10, I get a factor 10 reduction in error as well.

pierreguilmin:
> Yes I think there's a bug. By reducing to dt = 1e-4 I get a worse performance than dt = 1e-3. I'm pretty sure it used to work so probably a bug introduced in a recent PR? Rouchon1 seems correct though because reducing dt by a factor of 10, I get a factor 10 reduction in error as well.

Alright, let's look at this in another PR.

PR #172, opened 2023-07-16
Uniform and cleaner dtype and device handling

pierreguilmin:
I am sorry, this PR gathers much in a single commit, but it tries to fix many previous flaws at once:
- Hard to track which dtype/device was used by the solver.
- Hard to understand what was the priority of the chosen dtype/device (the default one, the one from options, the one from `y0`?).
- Hard to make sure that all tensors had the same dtype/device (actually it was bugged).
- False exception raised like "tensors are not on the same device" while one was just a "cuda" `torch.device` and the other one a "cuda" `str`.
- Some dangling `None` dtype/device were carried around everywhere, however when using `.to(dtype=None, device=None)` it defaults to _not changing_ the dtype/device, while we wanted it to use the default torch dtype/device. This is fixed by always directly converting the `None` dtype/device to a default.

A few remarks:
- This also sets the ground for mypy type checking not to scream too loud.
- I chose to remove the `complex_tensor` decorator to simplify the logic.
- With this proposal, every tensor within the solver uses **the same** dtype/device, which is stored in options.dtype/options.device. It is either specified by the user, or taken from torch default. It disregards the type of the argument tensors. We may use the dtype/device of y0 instead of torch default if options is not specified, I'm not sure whether it's better. What do you think?


dynamiqs/options.py
```diff
@@ -41,13 +39,15 @@ def __init__(
...
+            device (string or torch.device, optional): Device on which the tensors are
+                stored.
         """
         self.gradient_alg = gradient_alg
         self.save_states = save_states
         self.verbose = verbose
-        self.dtype = dtype
-        self.device = device
+        self.cdtype = to_cdtype(dtype)
+        self.rdtype = dtype_complex_to_real(self.cdtype)
```
gautierronan, line 49: Wait why do we have two different functions `to_cdtype` and `dtype_real_to_complex` again? Shouldn't we merge them?

dynamiqs/solvers/solver.py
```diff
@@ -38,6 +36,11 @@ def __init__(
         self.exp_ops = exp_ops
         self.options = options
 
+        # aliases
+        self.cdtype = self.options.cdtype
+        self.rdtype = self.options.rdtype
+        self.device = self.options.device
+
```
gautierronan, line 43: Not sure about this. Don't you think it's risky, in case one is modified and not the other?

gautierronan (changes requested):
Marvellous !

dynamiqs/solvers/solver.py
pierreguilmin: We could do better by having only a property, but honestly it's internal I think it's fine.

pierreguilmin:
I didn't rename `dtype_complex_to_real`, I think it's clear as is.

PR #175, opened 2023-07-16
Add `ArrayLike` type to replace `TensorLike` and `OperatorLike`

pierreguilmin:
On top of #173.

Fuses `TensorLike` and `OperatorLike` for simpler API. I don't really expect anyone to use a `qutip.Qobj` to define their `t_save`, so I think it's fine to not distinguish them. 😅 Also fix a bug with `to_tensor` not working for list of floats.


gautierronan (approved):
Down for this

PR #176, opened 2023-07-16
Simplify tensor batching

pierreguilmin:
On top of #175.

With the new tools we have it's easy to get rid of `TensorFormatter`, I think it's easier to read this way. Note that we can't put `batch_H` and `batch_y0` in `solvers.utils.utils` because this would create a circular import `solvers.utils.utils > solvers.utils.td_tensor > solvers.utils.utils`. This is why I created a new file `solvers.utils.batching`. I also added some import to `solvers.utils.__init__` to lighten the imports.


gautierronan (dismissed):
Beautiful

PR #180, opened 2023-07-19
Clean up getting started page

gautierronan:


gautierronan:
@pierreguilmin How can I fix the CI? It seems there are no more tests for doctest in the docs/ folder, and so it's not happy.

pierreguilmin:
> @pierreguilmin How can I fix the CI? It seems there are no more tests for doctest in the docs/ folder, and so it's not happy.

Small temporary fix hehe, it won't appear on the website.

PR #183, opened 2023-07-21
Add documentation examples to utils

pierreguilmin:
Screenshot 2023-07-21 at 23 23 09 (2)


abocquet (approved):
Great work!

gautierronan:
Thanks for writing all this @pierreguilmin ! Did not review the full content yet, but I made a proposal for an alternative examples styling in d9a826160e9a0df1a4f5ecba64949b9d454fa802, the goal being to have a single block of code for each examples section, instead of multiple in a row. It's just a proposal which I personally prefer, but if you don't like it your solution is fine with me too !

Screenshot 2023-07-23 at 10 15 35

Screenshot 2023-07-23 at 10 15 39

dynamiqs/utils/utils.py
```diff
@@ -111,6 +161,52 @@ def unit(x: Tensor) -> Tensor:
...
+        tensor(1.)
+
+        If the argument is batched, each batch element is normalized separately:
+        >>> fock0, fock1 = dq.fock(8, 0), dq.fock(8, 1)
+        >>> x = torch.stack([fock0, fock0 + fock1])  # shape: (2, 8, 1)
+        >>> dq.norm(x)
+        tensor([1.000, 1.414])
+        >>> x = dq.unit(x)                           # shape: (2, 8, 1)
+        >>> dq.norm(x)
+        tensor([1.000, 1.000])
```
gautierronan, line 208: I think this is overdocumented. It's a simple function, it should have simple documentation.

For instance for the ket part, I think this is already plenty:
```suggestion

    Examples:
        For a ket:
        >>> psi = dq.fock(4, 0) + dq.fock(4, 1)
        >>> dq.norm(psi)
        tensor(1.414)
        >>> psi = dq.unit(psi)
        >>> dq.norm(psi)
        tensor(1.000)
```

dynamiqs/utils/utils.py
```diff
@@ -205,20 +305,43 @@ def ptrace(x: Tensor, keep: int | tuple[int, ...], dims: tuple[int, ...]) -> Ten
...
+        If the argument is batched, the partial trace is computed for each batch
+        element:
+        >>> psi1 = dq.tensprod(dq.fock(3, 0), dq.fock(4, 1), dq.fock(5, 1))
+        >>> psi2 = dq.tensprod(dq.fock(3, 1), dq.fock(4, 3), dq.fock(5, 2))
+        >>> x_abc = torch.stack([psi1, psi2])     # shape: (2, 60, 1)
+        >>> x_a = dq.ptrace(x_abc, 0, (3, 4, 5))  # shape: (2, 3, 3)
+        >>> torch.all(x_a[0] == dq.ptrace(x_abc[0], 0, (3, 4, 5)))
+        tensor(True)
+        >>> torch.all(x_a[1] == dq.ptrace(x_abc[1], 0, (3, 4, 5)))
+        tensor(True)
```
gautierronan, line 343: You used `all` instead of `torch.all` in the previous function.

dynamiqs/utils/utils.py
```diff
@@ -315,6 +462,29 @@ def lindbladian(H: Tensor, Ls: Tensor, rho: Tensor) -> Tensor:
...
+        >>> H = delta * a.mH @ a
+        >>> L1 = np.sqrt(kappa_1) * a
+        >>> Lphi = np.sqrt(2 * kappa_phi) * a.mH @ a
+        >>> Ls = torch.stack([L1, Lphi])
+        >>> rho = dq.coherent_dm(4, 0.5)
+        >>> dq.lindbladian(H, Ls, rho)
+        tensor([[ 0.019+0.000j, -0.029+0.389j, -0.038+0.275j, -0.025+0.125j],
+                [-0.029-0.389j, -0.015+0.000j, -0.012+0.069j, -0.008+0.042j],
+                [-0.038-0.275j, -0.012-0.069j, -0.004+0.000j, -0.002+0.007j],
+                [-0.025-0.125j, -0.008-0.042j, -0.002-0.007j, -0.001+0.000j]])
```
gautierronan, line 483: Is it necessary to introduce delta, kappa_1 and kappa_phi here? Why not just define directly:
```suggestion
>>> a = dq.destroy(4)
>>> H = a.mH @ a
>>> Ls = torch.stack([a, a.mH @ a])
>>> rho = dq.fock_dm(4, 1)
>>> dq.linbladian(H, Ls, rho)
```
for instance?

dynamiqs/utils/utils.py
```diff
@@ -378,6 +572,31 @@ def ket_to_bra(x: Tensor) -> Tensor:
...
         _(..., 1, n)_ Bra.
+
+    Examples:
+        >>> psi = dq.fock(3, 1)  # shape: (3, 1)
+        >>> psi
+        tensor([[0.+0.j],
+                [1.+0.j],
+                [0.+0.j]])
+        >>> dq.ket_to_bra(psi)  # shape: (1, 3)
+        tensor([[0.-0.j, 1.-0.j, 0.-0.j]])
```
gautierronan, line 582:
```suggestion
>>> psi = dq.fock(3, 1)       # shape: (3, 1)
>>> bra = dq.ket_to_bra(Psi)  # shape: (1, 3)
>>> bra == psi.mH
True
```
More straight to the point IMO

gautierronan (changes requested):
Thanks for the PR. Some comments below. Overall I fear some functions are over-documented. I'd be in favor of reducing a bit and going more straight to the point about what the function is doing. This is especially true for simpler functions.

pierreguilmin:
I kept the italic styling, but removed the colors (I think they look better a little dimmed compared to the rest of the documentation). I simplified all the examples, I think you're right they are way too long, and we can something clear in getting started to explain how batching works (its all the same for all functions).

pierreguilmin:
> PS: For longer examples (more complicated functions), we can also add a <BLANKLINE> to separate the codeblock in two, but still keeping a single code block. This should improve readability.

We can't use <BLANKLINE>, it's a command specific to doctest. I added an example of how to do what you suggest in the documentation of `norm`.

> You used all instead of `torch.all` in the previous function.

Yes you're right, we should use `torch.all` in all examples.

For the suggestion for `ket_to_bra` I like showing how the tensors looks like, to really understand the shape. I think we could add your suggestion but not as an example, rather a note like proposed in the last commit.

gautierronan:
> and we can something clear in getting started to explain how batching works (its all the same for all functions).

Yes I think having a section on batching, and a section on GPU simulations in Getting Started would be the best way to go about this.

gautierronan (approved):
Thanks for the corrections I think it's nice like that ! :)

PR #188, opened 2023-08-01
Fix result formatting bug

abocquet:
Used to brake the readme example

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -75,19 +75,19 @@ def mesolve(
     # check jump_ops
     if not isinstance(jump_ops, list):
         raise TypeError(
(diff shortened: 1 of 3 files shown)


pierreguilmin (changes requested):
Nice, can you fix `sesolve` and `smesolve` too?

Also we only need it for `exp_save`, `y_save` is never `None`.

pierreguilmin (approved):
Thanks!

> I feel like we should have an option to not save all the matrices, as Qutip does. It could save a lot of memory space, especially when batching simulations. What do you think?

We do have this options (`save_states`), if `False` only the final state is returned (this is the reason why `y_save` is never `None`). Do you think we should be more granular, with something like `save_states = True | False | 'last'`?

abocquet:
No it's fine like this. I didn't have this detail in mind, thanks!

PR #189, opened 2023-08-02
Favicon fix

gautierronan:
Use a .svg instead of a .ico for better resolution. Also add a line for safari tab favicon (looked weird before).


docs/overrides/main.html
```diff
@@ -0,0 +1,28 @@
...
+    <link rel="canonical" href="{{ page.canonical_url }}">
+{% endif %}
+{% if page.previous_page %}
+    <link rel="prev" href="{{ page.previous_page.url | url }}">
+{% endif %}
+{% if page.next_page %}
+    <link rel="next" href="{{ page.next_page.url | url }}">
+{% endif %}
+<link rel="mask-icon" href="{{ config.theme.favicon | url}}" color="#ffffff">
+<link rel="icon" class="js-site-favicon" type="image/svg+xml" href="{{ config.theme.favicon | url}}">
```
gautierronan, line 26: These are the two new lines compared to `base.html` default.

pierreguilmin (approved):

PR #193, opened 2023-08-28
Fix installation instructions

pierreguilmin:
The previous installation was broken when installing with pip in non-editable mode. This fixes the issue, and switches from the default `setuptools` to `flit` to package our library (much simpler to use).


pierreguilmin:
Thanks @blaise-vignon for pointing that out.

PR #197, opened 2023-09-05
Fix incorrect `torch.squeeze` usage

pierreguilmin:
Passing a tuple to `torch.squeeze` is only valid in PyTorch >= 2.0.


gautierronan (approved):
Thanks for the fix

PR #201, opened 2023-09-08
Rename internal variables

pierreguilmin:


pierreguilmin:
We could also name them `Ls` because it's many jump operators (for example we named `M1s` in Rouchon method), not too sure.

dynamiqs/mesolve/me_solver.py
```diff
@@ -7,14 +7,12 @@
 class MESolver(Solver):
     def __init__(self, *args, jump_ops: Tensor):
         super().__init__(*args)
-        self.jump_ops = jump_ops[None, ...]  # (1, len(jump_ops), n, n)
-        self.sum_no_jump = (
-            (self.jump_ops.mH @ self.jump_ops).squeeze(0).sum(dim=0)
-        )  # (n, n)
+        self.L = jump_ops[None, ...]  # (1, len(L), n, n)
+        self.LdagL = (self.L.mH @ self.L).squeeze(0).sum(dim=0)  # (n, n)
```
gautierronan, line 11:
```suggestion
        self.sum_LdagL = (self.L.mH @ self.L).squeeze(0).sum(dim=0)  # (n, n)
```
(or the other way around, `self.LdagL_sum`, or even `self.LdagLsum`)

gautierronan (changes requested):

gautierronan (approved):

PR #203, opened 2023-09-08
Add Rouchon solver of order 1 for SME

pierreguilmin:
Here's a little cheatsheet to help checking the solver, it's actually very similar to Rouchon of order 1 for `mesolve`.

Screenshot 2023-09-08 at 16 41 48

Note that purely dissipative processes (jump operators not monitored, named $T$ in the code) have $\eta_k=0$, and we can thus compute terms that depend on $\eta_k$ only for monitored jump operators (named $V$ in the code).

### DIFF
--- a/dynamiqs/smesolve/sme_solver.py
+++ b/dynamiqs/smesolve/sme_solver.py
@@ -110,7 +110,9 @@ class SMESolver(MESolver):
         prefactor = self.etas.sqrt() * dw
         return (prefactor[..., None, None] * tmp).sum(-3)
 
-    def update_meas(self, dw: Tensor, rho: Tensor):
-        Lp_rho = self.Lp(rho)  # (..., b, n, n)
-        exp_val = self.exp_val(Lp_rho)  # (..., b)
-        self.bin_meas += self.etas.sqrt() * exp_val * self.dt + dw
+    def update_meas(self, dw: Tensor, rho: Tensor) -> Tensor:
+        Lp_rho = self.Lp(rho)  # (..., len(V), n, n)
+        exp_val = self.exp_val(Lp_rho)  # (..., len(V))
+        dy = self.etas.sqrt() * exp_val * self.dt + dw  # (..., len(V))
+        self.bin_meas += dy
+        return dy
--- a/dynamiqs/smesolve/sme_solver.py
+++ b/dynamiqs/smesolve/sme_solver.py
@@ -15,17 +15,20 @@ class SMESolver(MESolver):
         self,
         *args,
         jump_ops: Tensor,
-        meas_ops: Tensor,
         etas: Tensor,
         generator: torch.Generator,
         t_meas: Tensor,
     ):
-        self.V = meas_ops  # (len(V), n, n)
-        self.etas = etas  # (len(V))
-        self.generator = generator
         self.t_meas = t_meas  # (len(t_meas))
-
         super().__init__(*args, jump_ops=jump_ops)
+
+        # split jump operators between purely dissipative (eta = 0) and
+        # monitored (eta != 0)
+        mask = etas == 0.0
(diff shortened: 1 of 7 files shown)


dynamiqs/smesolve/sme_solver.py
```diff
@@ -15,24 +15,27 @@ def __init__(
...
-        self.generator = generator
         self.t_meas = t_meas  # (len(t_meas))
-
         super().__init__(*args, jump_ops=jump_ops)
 
+        # split jump operators between purely dissipative (eta = 0) and
+        # monitored (eta != 0)
+        mask = etas == 0.0
+        self.T = jump_ops[mask]  # (len(T), n, n) purely dissipative
+        self.V = jump_ops[~mask]  # (len(V), n, n) monitored
```
gautierronan, line 29: Are these notations from somewhere? IMO they suggest kinetic (T) and potential (V) energy terms. Could we use something else? Maybe `self.J` (jump) or `self.C` (collapse) and `self.M` (measured/monitored)?

dynamiqs/smesolve/sme_solver.py
```diff
@@ -15,24 +15,27 @@ def __init__(
...
         etas: Tensor,
         generator: torch.Generator,
         t_meas: Tensor,
     ):
-        self.meas_ops = meas_ops  # (len(meas_ops), n, n)
-        self.etas = etas  # (len(meas_ops))
-        self.generator = generator
         self.t_meas = t_meas  # (len(t_meas))
-
         super().__init__(*args, jump_ops=jump_ops)
```
gautierronan, line 23: Can you rename `jump_ops` to `L` like you did in mesolve (I think?)?

dynamiqs/smesolve/sme_solver.py
```diff
@@ -79,37 +82,40 @@ def sample_wiener(self, dt: float) -> Tensor:
...
     @cache
-    def Lp(self, rho: Tensor) -> Tensor:
-        # rho: (b_H, b_rho, ntrajs, n, n) -> (b_H, b_rho, ntrajs, n, n)
-        # L @ rho + rho @ Ldag
-        L_rho = torch.einsum('bij,...jk->...bik', self.meas_ops, rho)
-        return L_rho + L_rho.mH
+    def Vp(self, rho: Tensor) -> Tensor:
+        # rho: (b_H, b_rho, ntrajs, n, n) -> (b_H, b_rho, ntrajs, len(V), n, n)
+        # V @ rho + rho @ Vdag
+        V_rho = torch.einsum('bij,...jk->...bik', self.V, rho)
```
gautierronan, line 88: Probably faster to use something else than einsum when possible. Here I'm guessing a simple `@` is best.

gautierronan (changes requested):
Nice. Did not check the formulas, so may need a double check

dynamiqs/smesolve/sme_solver.py
pierreguilmin: `SMESolver` inherits from `MESolver` so it's done automatically.

dynamiqs/smesolve/sme_solver.py
pierreguilmin: `@` doesn't work (`self.V` has shape `(len(V), n, n)` and `rho` has shape `(..., n, n)`).

dynamiqs/smesolve/sme_solver.py
pierreguilmin: `self.M` conflicts with `Rouchon` solver `self.M0/self.M1s` and with the measurement back-action superoperator `\mathcal{M}`. 😕 I picked `Lc` (for collapse) and `Lm` (for monitored), tell me if that works for you.

gautierronan: Perfect !

gautierronan (approved):
Great job !

PR #204, opened 2023-09-08
Dictionary API for solver options

pierreguilmin:
This is a proposal to simplify the options API.

Before:
```python
options = dq.options.Rouchon1(dt=1e-3, gradient_alg='autograd', verbose=False)
result = dq.mesolve(H, jump_ops, rho0, t_save, options=options)
```
After
```python
options = dict(dt=1e-3, verbose=False)
result = dq.mesolve(H, jump_ops, rho0, t_save, solver='rouchon1', gradient='autograd', options=options)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ t_save = np.linspace(0, 1.0, 101)
 # torch.set_default_device('gpu')
 
 # simulation
-options = dq.options.Dopri5(gradient_alg='autograd', verbose=False)
+options = dict(gradient_alg='autograd', verbose=False)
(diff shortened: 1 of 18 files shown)


gautierronan (changes requested):
Great PR. I would simply add `gradient_alg` also to the base parameters of `mesolve`. Because both `solver` and `gradient_alg` determine which subroutine of `mesolve` will be called. Whereas all the other arguments are common to all subroutines.

gautierronan (approved):
Perfect. Love it ! @abocquet FYI we thought the current API was too complicated and wanted to simplify to ease adoption of the library. That's what we converged towards.

PR #207, opened 2023-09-08
Align some of utils API with qutip

gautierronan:
Aligns some of our utilities API with existing qutip API to avoid confusion. For some functions I just couldn't however... For instance, I find `qutip.ket2dm` absolutely 🤮


pierreguilmin (changes requested):
Thanks for the PR! I really hate and would prefer to revert
- `qeye`, there's nothing quantum about it, it's just the identity. I think we should really keep `eye` (like numpy and pytorch).
- `num`, it's a shortcut for an already short name `number` so I don't see the point, and we always call it the *number operator*.

I actually remember building these two by hands because I couldn't find them in qutip at first, before discovering they had weird names.

pierreguilmin (approved):

PR #214, opened 2023-09-11
Add utilities for quantum optimal control

pierreguilmin:
This PR proposes a structure to separate:
- the definition of the parameters with respect to which we compute the gradient (see the `rand_complex` function)
- the definition of the pulse used in the Hamiltonian, which is a function that takes some time `t` and returns the pulse value at that time (see the `pwc_pulse` fuction)

Example:
```python
>>> nseeds = 10
>>> nbins = 101
>>> theta = dq.rand_complex((nseeds, nbins), rmax=2.0, seed=42)  # parameters w.r.t which we compute the gradient
>>> pulse = dq.pwc_pulse(0.0, 1.0, theta)  # returns a function that defines the pulse value at each time
>>> pulse(0.5)
tensor([ 0.844-0.905j, -1.593-0.798j])
```

It's slightly overkill for now, but will be very useful:
- to define pulses in different bases (e.g. Fourier, the parameters will be the Fourier coefficients, but the pulse will always be a function that takes a time and returns the pulse value)
- to define filtering (we can now do `gaussian_filter(pulse, ...)`)
- to use directly `pulse(t)` in the code, e.g. `H = lambda t: pulse(t).conj() * a + pulse(t) * a.mH` rather than handling annoying indices
- to make clear the distinction between these two objects, in the end the pulse can be quite a complex construction on top of the parameters over which we optimize, it shouldn't be the same object

For `rand_complex`, the `size` parameter would typically be `(nseeds, nbins)` where `nseeds` is the number of random initialisation to run in //, and `nbins` the number of time bins.

Note that this PR is not 100% finished, I'd like to refine the doc of `pwc_pulse`.


pierreguilmin:
Another option would be to have `dq.pwc_pulse` taking all the arguments of the current `rand_complex` and `rand_pwc`, and returning both the pulse function and the parameters. But I think it's better to decompose a bit, it's more clear IMO.

dynamiqs/utils/optimal_control.py
```diff
@@ -0,0 +1,131 @@
...
+        resulting complex numbers are uniformly distributed in the complex plane.
+
+    Args:
+        size: Size of the returned tensor.
+        rmax: Maximum magnitude.
+        requires_grad: Whether gradients need to be computed with respect to the
+            returned tensor.
+        seed: Seed for the random number generator.
+        dtype: Complex data type of the returned tensor.
+        device: Device of the returned tensor.
```
gautierronan, line 51: Missing types in paranthesis, like in the rest of the doc.

dynamiqs/utils/optimal_control.py
```diff
@@ -0,0 +1,131 @@
...
+    # generate random phase with values in [0, 2pi]
+    theta = 2 * torch.pi * rand()
+    x = r * torch.exp(1j * theta)
+
+    x.requires_grad = requires_grad
+
+    return x
+
+
+def pwc_pulse(t_start: float, t_end: float, x: Tensor) -> callable[[float], Tensor]:
```
gautierronan, line 89: How about non evenly spaced PWC pulses? I would add this possibility in the same function. It can be done through an optional parameter `times` that takes the list of intermediate times, and is of same length as `x`. Alternatively, we can directly replace `t_start` and `t_end` with a `np.linspace`-like array of times. Or allow the choice between the two. 

So maybe:
```python
def pwc_pulse(times: Tuple[float, float] | Tensor, amplitudes: Tensor) -> callable[[float], Tensor]
```

gautierronan, line 89: Maybe rename `x` for a more suggestive name? `amplitudes` maybe?

dynamiqs/utils/optimal_control.py
```diff
@@ -0,0 +1,131 @@
...
+    """Returns a piecewise-constant (PWC) pulse.
```
gautierronan, line 90: Not super clear that `pulse` actually means a callable here. Would be nice to make it clear, either in the single-line or later.

gautierronan (changes requested):
A few comments below :)

dynamiqs/utils/optimal_control.py
pierreguilmin: I added the typing to `size` but not the others, we only type complex types in this part of the doc, the rest is in the signature.

dynamiqs/utils/optimal_control.py
pierreguilmin: I agree we need this feature, but I think we should make a different function because it's a 1% vs 99% use-case, so I'd rather keep the API and doc of `pwc_pulse` as simple as possible.

pierreguilmin: `amplitude` refers to the magnitude of the complex value, I picked `values`.

gautierronan (approved):

PR #215, opened 2023-09-11
Automatically generated figures in the doc

pierreguilmin:
WIP PR to use doctest to automatically generate figures that sit in our doc.

Left to do:
- [ ] transparent background / darkmode compatible figures
- [ ] hide line with comment `# hide` from the resulting doc
- [ ] run doctest in the doc building CI

Screenshot 2023-09-11 at 09 51 15 (2)


pierreguilmin:
Will be replaced soon, thanks for the good starting point!

PR #216, opened 2023-09-11
Round time difference in propagator solvers to prevent useless cache reloading

### LINKED ISSUE #209: Propagator solvers are cached on slighlty changing `delta_t`
Both the `sesolve` and `mesolve` propagator solvers are cached on the time step `delta_t` to take, which should be constant for linearly spaced `t_save`. Thus, the propagator should be computed only once. However, due to numerical imprecisions, the `delta_t` changes slightly even when `t_save` is linearly spaced, resulting in frequent recomputations of the same quantity.

pierreguilmin:
Fixes #209.

```python
import dynamiqs as dq
import numpy as np

n = 10
a = dq.destroy(n)
H = a.mH @ a
jump_ops = [np.sqrt(0.1) * a]
rho0 = dq.coherent(n, 1.0)
t_save = np.linspace(0.0, 1e-3, 11, np.float32)
res = dq.mesolve(H, jump_ops, rho0, t_save, solver='propagator')
```
Output when adding a `print` statement to `Propagator` and another `print` statement in the cached method `mesolve.propagator` to ensure it runs only once:
```
Propagator: 0.00009999999747378751635551452637 --> 0.00010000000000000000479217360239
MEPropagator: compute propagator
Propagator: 0.00009999999747378751635551452637 --> 0.00010000000000000000479217360239
Propagator: 0.00010000001930166035890579223633 --> 0.00010000000000000000479217360239
Propagator: 0.00009999997564591467380523681641 --> 0.00010000000000000000479217360239
Propagator: 0.00010000003385357558727264404297 --> 0.00010000000000000000479217360239
Propagator: 0.00010000000474974513053894042969 --> 0.00010000000000000000479217360239
Propagator: 0.00009999994654208421707153320312 --> 0.00010000000000000000479217360239
Propagator: 0.00010000000474974513053894042969 --> 0.00010000000000000000479217360239
Propagator: 0.00010000000474974513053894042969 --> 0.00010000000000000000479217360239
Propagator: 0.00010000006295740604400634765625 --> 0.00010000000000000000479217360239
```

### DIFF
--- a/dynamiqs/solvers/propagator.py
+++ b/dynamiqs/solvers/propagator.py
@@ -19,8 +19,8 @@ def round_truncate(x: np.float32 | np.float64) -> np.float32 | np.float64:
     # removes any numerical error that may have accumulated in the 5 least significant
     # bits of the mantissa.
     leading = abs(int(np.log2(x)))
-    kept = leading + 18
-    return (x * 2**kept).round() / 2**kept
+    keep = leading + 18
+    return (x * 2**keep).round() / 2**keep
 
 
 class Propagator(AutogradSolver):
--- a/dynamiqs/solvers/propagator.py
+++ b/dynamiqs/solvers/propagator.py
@@ -1,10 +1,22 @@
+from __future__ import annotations
+
 from abc import abstractmethod
 
+import numpy as np
 from torch import Tensor
 
 from .solver import AutogradSolver
 from .utils.td_tensor import ConstantTDTensor
 from .utils.utils import tqdm
+
+
+def round_truncate(x: np.float32 | np.float64) -> np.float32 | np.float64:
+    # round a strictly positive-valued float to remove numerical errors, and enable
+    # comparing floats for equality
+
+    # round to 5 additional decimal places beyond the highest digit of the number
+    decimals = abs(int(np.log10(x))) + 5
+    return (x * 10**decimals).round() / 10**decimals
 
 
 class Propagator(AutogradSolver):
@@ -21,7 +33,10 @@ class Propagator(AutogradSolver):
     def run_autograd(self):
         y, t1 = self.y0, 0.0
         for t2 in tqdm(self.t_stop.cpu().numpy(), disable=not self.options.verbose):
-            y = self.forward(t1, t2 - t1, y)
+            if t2 != 0.0:
+                # round time difference to avoid numerical errors when comparing floats
+                delta_t = round_truncate(t2 - t1)
+                y = self.forward(t1, delta_t, y)
             self.save(y)
             t1 = t2
 
--- a/dynamiqs/solvers/propagator.py
+++ b/dynamiqs/solvers/propagator.py
@@ -14,9 +14,13 @@ def round_truncate(x: np.float32 | np.float64) -> np.float32 | np.float64:
     # round a strictly positive-valued float to remove numerical errors, and enable
     # comparing floats for equality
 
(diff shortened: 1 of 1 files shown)


dynamiqs/solvers/propagator.py
```diff
@@ -1,12 +1,24 @@
...
 from .utils.td_tensor import ConstantTDTensor
 from .utils.utils import tqdm
 
 
+def round_truncate(x: np.float32 | np.float64) -> np.float32 | np.float64:
+    # round a strictly positive-valued float to remove numerical errors, and enable
+    # comparing floats for equality
+
+    # round to 5 additional decimal places beyond the highest digit of the number
+    decimals = abs(int(np.log10(x))) + 5
```
gautierronan, line 18: I think it would make more sense to do this in binary (also easier to adapt between float16/float32/float64). So `decimals = abs(int(np.log2(x))) + 16` maybe? (there are 23 bits allocated to the fraction in a float32, and 2**16 = 65536, `20` could be good also)

gautierronan (changes requested):

dynamiqs/solvers/propagator.py
pierreguilmin: Why not, but for some reason the truncation is less satisfactory:

For decimals:
```
0.00009999999747378751635551452637 -> 0.00010000000000000000479217360239
0.00009999999747378751635551452637 -> 0.00010000000000000000479217360239
0.00010000001930166035890579223633 -> 0.00010000000000000000479217360239
0.00009999997564591467380523681641 -> 0.00010000000000000000479217360239
0.00010000003385357558727264404297 -> 0.00010000000000000000479217360239
0.00010000000474974513053894042969 -> 0.00010000000000000000479217360239
0.00009999994654208421707153320312 -> 0.00010000000000000000479217360239
0.00010000000474974513053894042969 -> 0.00010000000000000000479217360239
0.00010000000474974513053894042969 -> 0.00010000000000000000479217360239
0.00010000006295740604400634765625 -> 0.00010000000000000000479217360239
```

For binary:
```text
0.00009999999747378751635551452637 -> 0.00009999983012676239013671875000
0.00009999999747378751635551452637 -> 0.00009999983012676239013671875000
0.00010000001930166035890579223633 -> 0.00009999983012676239013671875000
0.00009999997564591467380523681641 -> 0.00009999983012676239013671875000
0.00010000003385357558727264404297 -> 0.00009999983012676239013671875000
0.00010000000474974513053894042969 -> 0.00009999983012676239013671875000
0.00009999994654208421707153320312 -> 0.00009999983012676239013671875000
0.00010000000474974513053894042969 -> 0.00009999983012676239013671875000
0.00010000000474974513053894042969 -> 0.00009999983012676239013671875000
0.00010000006295740604400634765625 -> 0.00009999983012676239013671875000
```

I'm not sure it makes a huge difference, I also made a distinction between different types of float but then thought that it's a completely extreme use case to need such high precision on the `delta_t`.

gautierronan: As you wish, I prefer the binary version, but it's not super important in any case. The two rounded numbers you're showing are probably the same in float32 precision, and it's kinda random whether you're closer or not to the wanted `delta_t`.

pierreguilmin: Alright, I tried to make the comment as clear as possible.

gautierronan (approved):

PR #218, opened 2023-09-12
Add extra args to callable Hamiltonians

gautierronan:
Small PR to add Hamiltonians of the form `H(t: float, *args) -> Tensor` instead of `H(t: float) -> Tensor`.

- [ ] Test it works with an example.
- [ ] Write a test function ?

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -38,8 +38,8 @@ def mesolve(
     the graph of operations is not stored to improve performance.
 
     For time-dependent problems, the Hamiltonian `H` can be passed as a function with
-    signature `H(t: float) -> Tensor`. Piecewise constant Hamiltonians can also be
-    passed as... TODO Complete with full Hamiltonian format
+    signature `H(t: float, *H_args) -> Tensor`. Piecewise constant Hamiltonians can also
+    be passed as... TODO Complete with full Hamiltonian format
 
     Available solvers:
       - `Dopri5`: Dormand-Prince of order 5. Default solver.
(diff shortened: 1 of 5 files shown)


pierreguilmin:
I don't see the use case for this 🤔 If you know the `*args` when you define `options`, then you can easily define a `H` function compatible with `dynamiqs`:
```python
H = lambda t: H_with_args(t, *args)
```

gautierronan:
> I don't see the use case for this 🤔 If you know the `*args` when you define `options`, then you can easily define a `H` function compatible with `dynamiqs`:
> 
> ```python
> H = lambda t: H_with_args(t, *args)
> ```

Don't know enough about the good coding practices, but isn't it considered a bad habit to do what you describe? Doesn't this make `*args` a hidden/global variable or something like that?

For instance [`scipy.optimize.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html) does what I implemented in this PR, and so does QuTiP.

pierreguilmin:
> Don't know enough about the good coding practices, but isn't it considered a bad habit to do what you describe? Doesn't this make `*args` a hidden/global variable or something like that?

Not that I know of, it's actually better IMO. But maybe @abocquet has a different opinion on this? I think scipy and qutip do this for convenience (so you don't have to define a lambda function to match their API), but I find it rather confusing to mix my personal function arguments with the minimizer/solver options 🤷‍♂️

gautierronan:
> > Don't know enough about the good coding practices, but isn't it considered a bad habit to do what you describe? Doesn't this make `*args` a hidden/global variable or something like that?
> 
> Not that I know of, it's actually better IMO. But maybe @abocquet has a different opinion on this? I think scipy and qutip do this for convenience (so you don't have to define a lambda function to match their API), but I find it rather confusing to mix my personal function arguments with the minimizer/solver options 🤷‍♂️

Ok I'm fine with either solution. This PR is definitly short enough that I don't care whether it's pushed or not :D

pierreguilmin:
Closing this as discussed, don't hesitate to reopen if you disagree.

PR #224, opened 2023-09-16
Rename `t_save` to `tsave`

gautierronan:
This a proposal, to be discussed and approved or rejected.

I would like to rename `t_save` to `tsave`, as well as `t_meas` to `tmeas` (internal and external), `y_save` to `ysave` and `t_stop` to `tstop` (internal only).

The advantage I see to this:
 - Cleaner IMO (I don't like underscores all around the place when they are not needed)
 - More natural to me (maybe to most users?)
 - Follows qutip API more closely (qutip uses `tlist`)
 - This goes along the other recent API simplifications (`options = dict()`, `gradient_alg` -> `gradient`, etc.)

The obvious disadvantage is that it mixes underscore variables (`tsave`) with non-underscore variables (e.g. `exp_ops` or `jump_ops`). So does qutip, and I personally think it's fine.


pierreguilmin (approved):
100% agree!

PR #225, opened 2023-09-16
Refactor solvers

gautierronan:
This PR makes some refactoring of the solvers code in preparation for the adjoint state method with adaptive solvers.

In particular:
 - Create a new file that contains the `AdjointAutograd` class
 - For all solvers (fixed, adaptive, adjoint, propagator...), use the same logic of having a double nested loop. The outer loop runs over all `t_save` values. The optional inner loop runs the integration between `t_save[i]` and `t_save[i+1]`. For `Propagator`, there is only the outer loop.
 - Remove the previously defined `t_stop` in the saving logic, because it seemed redundent with `t_save` in the way it was previously written. But maybe I'm missing something, in which case I can revert the change.


gautierronan:
Closing in favor of #266

PR #226, opened 2023-09-17
Add adjoint state method for adaptive time step solvers

gautierronan:
This is first working version of the adjoint state method for adaptive time step solvers. It's not perfect but it does the job.

Some things I'd like to do, probably in subsequent PRs:

- [ ]  Merge the classes `AdjointFixedAutograd` and `AdjointAdaptiveAutograd` because they are very similar. There's a lot of common code in these two.
- [ ]  Compute the gradient in a more precise manner. Right now I'm just evaluating the gradient integral once every `dt` with the value of `y` and `a` computed. However, we also know intermediate values of `y` and `a` since they're used for the Runge Kutta integration. So we could improve precision on the gradient, and release the `rtol/atol` in the tests.
- [ ] Either implement the adjoint method for `sesolve`, or fix the class inheritence. Right now `SEDormandPrince` inherits from `DormandPrince` which itself inherits from `AdjointAdaptiveSolver`. However, it does not support the adjoint method. I don't expect anyone to ever use the adjoint method with `sesolve`, so may not be worth implementing it. Will depend on how simple this is to do.


gautierronan:
Closing in favor of #266

PR #227, opened 2023-09-18
Minor doc fixes

pierreguilmin:


docs/python_api/solvers.md
```diff
@@ -1,7 +1,7 @@
 # Solvers
 
-**dynamiqs** provides end-to-end differentiable solvers for three quantum differential equations.
+**dynamiqs** implements differentiable solvers for three types of quantum systems:
```
gautierronan, line 3: I preferred the previous formulation, it's more straightforward. I like the rest however.

gautierronan (changes requested):
Thanks for the corrections. Below one formulation I'm not so sure about, the rest is good.

docs/python_api/solvers.md
pierreguilmin: I think `quantum differential equations` is not a correct term, we're working with normal differential equations that describe quantum systems, I think the contraction might be confusing. And the `end-to-end` is obvious if not precised IMO, and a little confusing (what solvers would not be end-to-end?), but we can add it back if you prefer.

gautierronan: I agree the `end-to-end` is superfluous. Concerning the terminology `quantum differential equations`, I think it's fine and quite self-explanatory, although I agree it's not extremely rigorous. 

But what I mostly meant is that your formulation implies that the different equations act on different quantum systems. It's not the case. All three equations can act on the same system, but we take different stuff into account. So they're rather differential equations that describe quantum systems in three different ways, of increasing complexity.

So maybe:
```
**dynamiqs** provides differentiable solvers for three differential equations that describe quantum systems.
```

pierreguilmin: I agree, I like the reformulation, please check the chat-gpt rewritten version 😉 

> **dynamiqs** provides differentiable solvers for three types of differential equations describing quantum systems:

gautierronan (approved):

PR #233, opened 2023-09-30
Add support for PWC tensor

pierreguilmin:
This PR implements support for PWC Hamiltonian.

Notably it allows to use the propagator solver for optimal control problems with PWC drives, which leads to a huge speedup (e.g. optimisation of Fock 2 preparation with cavity + transmon goes from ~3 minutes to ~20 seconds on my CPU). I think it’s important to have a good enough PWC Hamiltonian `sesolve` solver, because this is the first step of many optimal control proof-of-concept, it's a quick way to know wether something is doable or not.

The propagator solver is by far our best solver for sesolve (for constant and PWC Hamiltonian), we're roughly on par with qutip performance-wise for n ≤ 300 on CPU. There are many possible improvements to be made to the propagator solver, namely support for sparse matrices, and use of e.g. Krylov method to compute directly the application of the matrix exponential to the state. Here's an example mini benchmark:
```python
n = 128
a = dq.destroy(n)
K = 0.1
H = a.mH @ a - K/2 * a.mH @ a.mH @ a @ a
psi0 = dq.coherent(n, 3)
tsave = np.linspace(0, 1.0, 11)

# ...

# === sesolve timing results
# dynamiqs (dopri5, CPU):     125.061 ± 10.26 ms
# dynamiqs (propagator, CPU): 3.141 ± 0.51 ms
# qutip:                      3.439 ± 0.17 ms
```

<details>
  <summary>Mini benchmark code</summary>

```python
import dynamiqs as dq
import qutip as qt
import numpy as np
from time import time

def run(f):
    niters = 10
    res = []
    for _ in range(niters):
        start = time()
        f()
        end = time()
        res.append((end - start) * 1000)
    res = np.array(res)
    print(f'{res.mean():.3f} ± {res.std():.2f} ms')

n = 128
a = dq.destroy(n)
K = 0.1
H = a.mH @ a - K/2 * a.mH @ a.mH @ a @ a
psi0 = dq.coherent(n, 3)
tsave = np.linspace(0, 1.0, 11)

# dynamiqs, dopri5
run(lambda: dq.sesolve(H, psi0, tsave, options=dict(verbose=False)))

# dynamiqs, propagator
run(lambda: dq.sesolve(H, psi0, tsave, solver='propagator', options=dict(verbose=False)))

# qutip, propagator
H = dq.to_qutip(H)
psi0 = dq.to_qutip(psi0)
run(lambda: qt.sesolve(H, psi0, tsave))
```

</details>

I propose to modify the doc later, once this is a bit more used and we have converged on the PWC format. For now the chosen format is tuple of static and PWC parts `(H0, (H1, times, values), ...)` where `times` has shape `n+1` and `values` has shape `n`. It’s a bit tricky to use lists because we also allow them for batched tensors.

### DIFF
--- a/dynamiqs/mesolve/propagator.py
+++ b/dynamiqs/mesolve/propagator.py
@@ -10,17 +10,20 @@ from .me_solver import MESolver
 class MEPropagator(MESolver, Propagator):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.lindbladian = slindbladian(self.H, self.L)  # (b_H, 1, n^2, n^2)
+        # self.lindbladian: (b_H, 1, n^2, n^2)
+        self.lindbladian = cache(lambda H: slindbladian(H, self.L))
         self.y0 = operator_to_vector(self.y0)  # (b_H, b_rho, n^2, 1)
 
     @cache
-    def propagator(self, delta_t: float) -> Tensor:
+    def propagator(self, lindbladian: Tensor, delta_t: float) -> Tensor:
         # -> (b_H, 1, n^2, n^2)
-        return torch.matrix_exp(self.lindbladian * delta_t)
+        return torch.matrix_exp(lindbladian * delta_t)
 
     def forward(self, t: float, delta_t: float, rho_vec: Tensor) -> Tensor:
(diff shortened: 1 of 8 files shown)


pierreguilmin:
Will be replaced by soon-to-be PRs.

PR #236, opened 2023-10-02
Add generic fidelity method

abocquet:
Add a `fidelity` method that dispatches to `dm_fidelity` or `ket_fidelity`

### DIFF
--- a/dynamiqs/utils/utils.py
+++ b/dynamiqs/utils/utils.py
@@ -629,17 +629,14 @@ def _dm_fidelity(x: Tensor, y: Tensor) -> Tensor:
 def _ket_dm_fidelity(x: Tensor, y: Tensor) -> Tensor:
     r"""Returns the fidelity between a pure state and a density matrix
 
-    The fidelity of a pure state $\rho$ and $\sigma$ is defined by:
+    The fidelity of a pure state $\rho$ and $\psi$ is defined by:
 
     $$
         F(\rho, \sigma) = \tr{\sqrt{\sqrt{\rho}\sigma\sqrt{\rho}}}^2.
     $$
-
-    Assuming $\rho = \ket{\psi}\bra{\psi}$, $\rho^2 = \rho$ thus $\rho = \sqrt{\rho}$
-    $$
-        F(\rho, \sigma) = \tr{\sqrt{\sqrt{\rho}\sigma\sqrt{\rho}}}^2
-        = \tr{\sqrt{\rho\sigma\rho}}^2
-        = \tr{\sqrt{\ket{\psi}\braket{}^2
+    where $\sigma = \ket{\psi}\bra{\psi}$  and simplifiable to
+    $
+        F(\rho, \sigma) = \tr{\bra{\psi}\rho\ket{\psi}}^2.
     $$
 
     Args:
--- a/dynamiqs/utils/utils.py
+++ b/dynamiqs/utils/utils.py
@@ -600,7 +600,6 @@ def fidelity(x: Tensor, y: Tensor) -> Tensor:
         >>> dq.fidelity(fock0, fock01)
         tensor(0.500)
     """
-
     if isket(x) and isket(y):
         return _ket_fidelity(x, y)
     elif isket(x):
--- a/dynamiqs/utils/utils.py
+++ b/dynamiqs/utils/utils.py
@@ -553,7 +553,7 @@ def fidelity(x: Tensor, y: Tensor) -> Tensor:
         >>> fock0 = dq.fock(3, 0)
         >>> dq.fidelity(fock0, fock0)
         tensor(1.)
-        >>> fock01 = 0.5 * (ket_to_dm(dq.fock(3, 1)) + ket_to_dm(dq.fock(3, 0)))
+        >>> fock01 = 0.5 * (dq.ket_to_dm(dq.fock(3, 1)) + dq.ket_to_dm(dq.fock(3, 0)))
         >>> dq.fidelity(fock01, fock01)
         tensor(1.000)
         >>> dq.fidelity(fock0, fock01)
--- a/dynamiqs/utils/utils.py
+++ b/dynamiqs/utils/utils.py
@@ -577,7 +577,7 @@ def ket_overlap(x: Tensor, y: Tensor) -> Tensor:
 
 
 def fidelity(x: Tensor, y: Tensor) -> Tensor:
-    r"""Returns the fidelity of two states, kets or density matrix.
+    r"""Returns the fidelity of two states, kets or density matrices.
 
     Warning:
         This definition is different from `qutip.fidelity()` which uses the square root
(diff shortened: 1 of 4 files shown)


gautierronan (changes requested):
Nice ! I think @pierreguilmin said it all, nothing else to add :)

pierreguilmin:
What would also be nice (but can be done in another PR) is to support having one argument batched but not the other. This is typically the case when you have a single target state $psi_target$ of shape $(n, 1)$ but multiple optimisation seeds resulting in many final states $psi_final$ of shape $(nseeds, n, 1)$. Currently one needs to do
```python
psi_target = psi_target[None, ...].repeat(nseeds, 1, 1)
```

dynamiqs/utils/utils.py
```diff
@@ -550,26 +553,24 @@ def fidelity(x: Tensor, y: Tensor) -> Tensor:
         >>> fock0 = dq.fock(3, 0)
         >>> dq.fidelity(fock0, fock0)
         tensor(1.)
-        >>> fock01 = 0.5 * (ket_to_dm(fock(3, 1)) + ket_to_dm(fock(3, 0)))
-        >>> dq.fidelity(fock1, fock1)
-        tensor(1)
+        >>> fock01 = 0.5 * (ket_to_dm(dq.fock(3, 1)) + ket_to_dm(dq.fock(3, 0)))
```
gautierronan, line 556:
```suggestion
        >>> fock01 = 0.5 * (dq.ket_to_dm(dq.fock(3, 1)) + dq.ket_to_dm(dq.fock(3, 0)))
```

dynamiqs/utils/utils.py
```diff
@@ -539,9 +538,13 @@ def ket_overlap(x: Tensor, y: Tensor) -> Tensor:
 def fidelity(x: Tensor, y: Tensor) -> Tensor:
     r"""Returns the fidelity of two states, kets or density matrix.
```
gautierronan, line 540: Can you add the fidelity formulas for ket/ket and dm/dm cases? Can be inline math so it's not too heavy to look at.

gautierronan (changes requested):
Nice! Just two small comments regarding the documentation :)

dynamiqs/utils/utils.py
```diff
@@ -536,7 +535,42 @@ def ket_overlap(x: Tensor, y: Tensor) -> Tensor:
...
+        return _ket_fidelity(x, y)
+    elif isket(x):
+        return _ket_dm_fidelity(x, y)
+    elif isket(y):
+        return _ket_dm_fidelity(y, x)
+    else:
+        return _dm_fidelity(x, y)
+
+
+def _ket_fidelity(x: Tensor, y: Tensor) -> Tensor:
```
pierreguilmin, line 573: I think it's costly and useless to maintain clean docstrings for internal functions (developers are knowledgeable, they don't need as much info), let's remove them and shorten the internal doc to:

For `_ket_fidelity`:
```python
# returns the fidelity of two kets: |<x, y>|^2
# x: (..., n, 1), y: (..., n, 1) -> (...)
```

For `_dm_fidelity`:
```python
# returns the fidelity of two density matrices: Tr[sqrt(sqrt(x) @ y @ sqrt(x))]^2
# x: (..., n, n), y: (..., n, n) -> (...)
```

For `_ket_dm_fidelity`:
```python
# returns the fidelity of a ket `x` and a density matrix `y`: <x|y|x>
# x: (..., n, 1), y: (..., n, n) -> (...)
```

dynamiqs/utils/utils.py
```diff
@@ -620,6 +626,30 @@ def dm_fidelity(x: Tensor, y: Tensor) -> Tensor:
...
+
+    Args:
+        x _(..., n, n)_: A density matrix.
+        y _(..., n, n)_: A ket.
+
+    Returns:
+        _(...)_ Real-valued fidelity.
+
+    """
+    return trace(ket_to_bra(x) @ y @ x).real
```
pierreguilmin, line 650:
```suggestion
    return (ket_to_bra(x) @ y @ x).squeeze(-1).squeeze(-1).real
```
Does the same, but is more clear on what we want to do on the tensor IMO (the resulting shape of `ket_to_bra(x) @ y @ x` is `(..., 1, 1)` which we want to squeeze. I'm scared the trace is a bit confusing becasue it appears in the definition of fidelities, but we don't actually compute the trace of any matrix here, it's just a hidden way to squeeze the result.

dynamiqs/utils/utils.py
```diff
@@ -536,7 +535,42 @@ def ket_overlap(x: Tensor, y: Tensor) -> Tensor:
...
+        y _(..., n, 1) or (..., n, n)_: Second ket or density matrix.
+
+    Returns:
+        _(...)_ Real-valued fidelity.
+
+    Examples:
+        >>> fock0 = dq.fock(3, 0)
+        >>> dq.fidelity(fock0, fock0)
+        tensor(1.)
+        >>> fock01 = 0.5 * (dq.ket_to_dm(dq.fock(3, 1)) + dq.ket_to_dm(dq.fock(3, 0)))
```
pierreguilmin, line 556:
```suggestion
        >>> fock01 = 0.5 * (dq.fock_dm(3, 1) + dq.fock_dm(3, 0))
```

dynamiqs/utils/utils.py
```diff
@@ -536,7 +535,42 @@ def ket_overlap(x: Tensor, y: Tensor) -> Tensor:
     return (ket_to_bra(x) @ y).squeeze(-1).sum(-1)
 
 
-def ket_fidelity(x: Tensor, y: Tensor) -> Tensor:
+def fidelity(x: Tensor, y: Tensor) -> Tensor:
+    r"""Returns the fidelity of two states, kets or density matrix.
```
pierreguilmin, line 539:
```suggestion
    r"""Returns the fidelity of two states, kets or density matrices.
```

dynamiqs/utils/utils.py
gautierronan: You can also `.squeeze(-2, -1)` in a single call.

Also, if we take the real part, we might need a check that `y` is hermitian, and to return a warning to say we take the real part if it isn't (in case people want to call that for operators).

pierreguilmin: > You can also .squeeze(-2, -1) in a single call.

Only Pytorch 2.0 unfortunately.

abocquet: I think I disagree. You wrote that `_ket_dm_fidelity` computes `<x|y|x>` which is by definition `Tr(y |xXx|)` so it makes perfect sense to use `trace` here

gautierronan: If you actually compute `<x|y|x>` then I agree we should not take the trace. Even if it's the same, since `Tr[<x|y|x>] = <x|y|x>`. And you're not computing `Tr[y |x><x|]` here which would equivalent but slower.

pierreguilmin (approved):
I added the formulas to the docstring, LGTM, thanks a lot 🥳

PR #239, opened 2023-10-02
Add first plots

pierreguilmin:
Here's a first set of methods dumped from code I had. I know it's already a lot, all comments welcome!

As discussed together, they would all deserve proper documentation, but maybe we can a little bit to iterate on these plotting functions before writing clean doc. I also need to finish #215 before.

You can try with:

```python
import dynamiqs as dq
import numpy as np
dq.mplstyle()
```

**`plot_pwc_pulse`**
```python
n = 20
theta = dq.rand_complex(n, seed=42)
tsave = np.linspace(0, 1.0, n + 1)
dq.plot_pwc_pulse(tsave, theta)
```
4d7dbf06-55e7-4660-9540-49b28899b752

**`plot_fock`**
```python
psi = dq.coherent(12, 2)
dq.plot_fock(psi)
```
6680341d-f026-4364-99fd-ec6eeb925ccf

```python
psi = dq.coherent(32, 2)
dq.plot_fock(psi, xticksall=False)
```
f9c17e4a-684a-4d83-826f-32e87f953fab

```python
psi = dq.coherent(64, 4)
dq.plot_fock(psi, xticksall=False, ymax=None)
```
da1fcee1-0be2-48a9-bb09-a2d79f219f05

**`plot_fock_evolution`**
```python
dq.plot_fock_evolution([dq.fock(16, i) for i in range(6)])
```
9535580b-1c49-46f4-8a51-adf79a484525

```python
dq.plot_fock_evolution([dq.coherent_dm(16, i // 2) for i in range(6)], logscale=True, logvmin=1e-5)
```
8a5b7876-dac3-48f4-97a3-b276b958b2c3


abocquet (dismissed):
Awesome! 

Nit : The difference between `plot_fock` and `plot_focks` is tiny. Maybe `plot_focks` could be renamed to something like `plock_fock_evolution` or `plock_fock_in_time`. It would be easier orally also

I'm not a huge fan of the inner ticks but that seems to be the convention in some theoretical papers so I guess it's as good as the outer ticks

dynamiqs/plots/utils.py
```diff
@@ -0,0 +1,147 @@
...
+            'axes.grid': False,
+            'axes.titlesize': 'larger',
+            'axes.labelsize': 'larger',
+            'grid.color': 'gray',
+            'grid.linestyle': '--',
+            'grid.alpha': 0.3,
+            'scatter.marker': 'x',
+            'lines.linewidth': 1.0,
+        }
+    )
```
gautierronan, line 96:
```suggestion
    plt.rcParams.update(
        {
        'figure.facecolor': 'white',
        'axes.facecolor': 'white',
        'savefig.facecolor': 'white',
        'font.size': 12,
        'figure.dpi': 150,
        'xtick.direction': 'in',
        'ytick.direction': 'in',
        'xtick.major.size': 4.0,
        'xtick.minor.size': 2.5,
        'xtick.major.width': 1.0,
        'xtick.labelsize': 10,
        'xtick.minor.visible': True,
        'ytick.major.size': 4.0,
        'ytick.minor.size': 2.5,
        'ytick.major.width': 1.0,
        'ytick.labelsize': 10,
        'ytick.minor.visible': True,
        'axes.grid': False,
        'axes.titlesize': 12,
        'axes.labelsize': 12,
        'axes.linewidth': 1.0,
        'axes.prop_cycle': cycler('color', ['#1398c4', '#ff6b6b', '#2ec4b6', '#ffc463', '#869bac', '#966f94', '#c0675c', '#20817d']),
        'grid.color': 'gray',
        'grid.linestyle': '--',
        'grid.alpha': 0.3,
        'scatter.marker': 'x',
        'lines.linewidth': 1.5,
        'legend.frameon': False,
        'legend.fontsize': 10,
        'legend.borderpad': 0.5,
        'legend.labelspacing': 0.08,
        'legend.handlelength': 1.2,
        'legend.handletextpad': 0.4,
        'legend.columnspacing': 0.5,
        'figure.figsize': [4.0, 2.5],
        }
    )
```
(taken from my own mplstyle)

gautierronan:
```python
from cycler import cycler
import numpy as np

plt.rcParams.update({})

# sample data
n = 16
x = np.linspace(0, 10, 30)
y = np.zeros((30, n))
for i in range(n):
    y[:, i] = 0.3 * i + np.sin(x)

# sample plot
fig, ax = plt.subplots(1, 1)
ax.plot(x, y, lw=2)
ax.set_xlabel('Time')
ax.set_ylabel('Sine')
ax.set_title('Sample plot')
plt.tight_layout()
plt.show()
```
image
image

dynamiqs/plots/misc.py
```diff
@@ -0,0 +1,116 @@
...
+    ax: Axes = None,
+    ycenter: bool = True,
+    real_color: str = '#0C5DA5',
+    imag_color: str = '#845B97',
+):
+    times = to_numpy(times)
+    values = to_numpy(values)
+
+    # format times and values, for example:
+    # times  = [0, 1, 2, 3] -> [0, 1, 1, 2, 2, 3, 3, 4]
```
gautierronan, line 31:
```suggestion
    # times  = [0, 1, 2, 3] -> [0, 1, 1, 2, 2, 3]
```

dynamiqs/plots/misc.py
```diff
@@ -0,0 +1,116 @@
...
+    values = values.repeat(2)
+    zeros = np.zeros_like(times)
+
+    # real part
+    ax.plot(times, values.real, alpha=0.6, lw=2.0, label='real', color=real_color)
+    ax.fill_between(times, zeros, values.real, alpha=0.2, color=real_color)
+
+    # imaginary part
+    ax.plot(times, values.imag, alpha=0.6, lw=2.0, label='imag', color=imag_color)
+    ax.fill_between(times, zeros, values.imag, alpha=0.2, color=imag_color)
```
gautierronan, line 43: I find its a bit strange to do it like this. Matplotlib provides a function `plt.stairs` that achieves the same thing natively. You then have to provide an array `times` of shape (n+1) with all the edges of the PWC pulse. But then there's no arbitrary choice of beginning or end times.

dynamiqs/plots/misc.py
```diff
@@ -0,0 +1,116 @@
...
+    # plot
+    ax.bar(x, y, alpha=alpha)
+    ax.set(xlim=(0 - 0.5, n - 0.5), ylim=(0, 1 + 0.05))
+
+    # set x ticks
+    fock_ticks(ax.xaxis, n, all=xticksall)
+
+
+@optax
+def plot_focks(
```
gautierronan, line 79: Agreed with @abocquet that we might want a different name for this one. I like `plot_fock_evolution`.

gautierronan (changes requested):
Very nice ! Some comments below :)

dynamiqs/plots/utils.py
gautierronan: Feel free to modify it whichever way you like in case you want to adopt this styling. If you don't, that's fine with me as well.

dynamiqs/plots/misc.py
```diff
@@ -0,0 +1,116 @@
...
+
+__all__ = [
+    'plot_pwc_pulse',
+    'plot_fock',
+    'plot_focks',
+]
+
+
+@optax
+def plot_pwc_pulse(
```
gautierronan, line 18: Maybe add a `set_xlim` on this plot as well so there are no blank spaces on right/left sides.

dynamiqs/plots/misc.py
pierreguilmin: > But then there's no arbitrary choice of beginning or end times.

There is no arbitrary choice here, the array of times must be of shape `(n+1)` as in `plt.stairs`, it does the exact same thing. The reason I am using a custom method is because it's hard to separately set a transparent fill color but a darker edge color with `plt.stairs`. 🤷‍♂️

pierreguilmin: Actually there is a difference, but we can easily adapt the current code, do we want the pulse stard and end to be at value 0.0? (compare the beginning and end of the pulse)
4fc93866-1928-4ad6-acec-2bfc7482c9a8
184a1f9c-c102-49fb-a4ee-f9117c3cc8c7

I would say yes (contrary to my original proposition), wdyt?

dynamiqs/plots/misc.py
```diff
@@ -0,0 +1,116 @@
...
+):
+    """Plot the photon number population."""
+    state = to_numpy(state)
+    n = state.shape[0]
+    x = range(n)
+    y = populations(state)
+
+    # plot
+    ax.bar(x, y, alpha=alpha)
+    ax.set(xlim=(0 - 0.5, n - 0.5), ylim=(0, 1 + 0.05))
```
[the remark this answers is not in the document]
pierreguilmin: I added an argument `ymax: float | None = 1.0` (default to 1.0 but can be unset using `None`). I don't understand what you propose for the `xlim`, the `0.5` is to make sure that the bar is right on the tick. I think it doesn't make sense to scale the xaxis with the `n` values (for example for a Fock space of dimension 64 you would see ticks for 65 and 66, very confusing IMO).

dynamiqs/plots/misc.py
pierreguilmin: It looks a little weird to me, maybe it's just because I'm not used to it.
9f07ea9d-4a5f-4b56-a699-080ed70443fe

If you prefer it like that I'll change.

dynamiqs/plots/utils.py
pierreguilmin: Thanks a lot for sharing!

I took everything except:
```python
'figure.dpi': 150   # -> kept 72 for now
'xtick.major.size': 4.0  # -> 4.5
'xtick.labelsize': 10  # -> 12
'ytick.major.size': 4.0  # -> 4.5
'ytick.labelsize': 10  # -> 12
'legend.fontsize': 10  # -> 12
'legend.labelspacing': 0.08  # -> removed
'legend.handlelength': 1.2  # -> removed
'legend.handletextpad': 0.4  # -> removed
'legend.columnspacing': 0.5  # -> removed
'lines.linewidth': 1.5,  # -> 2.0
'figure.figsize': [4.0, 2.5]  # -> removed
```

Here are the two styles:
2fb88817-9eed-4a4c-8323-3a860b29aaf2

dc073e9a-36be-4507-aa90-cf1b67a0f3fc

I think the first plot looks super nice, but is better suited for a research paper. I like having large detailed plots when working. I think we should propose both styles actually, and let people choose. For the colors I did a little mix-and-match so my current plot look nice, but I think we need to iterate too 😅 It might be better to write 2-3 `.mplstyle` file. I propose to merge this as is, and iterate on the styling later.

dynamiqs/plots/misc.py
pierreguilmin: I changed to remove blank spaces in latest commit.

dynamiqs/plots/utils.py
pierreguilmin: I think the best is to iterate on the styling once we have a representative set of figures that we plot often (fock, wigner, pulse, etc...) to make sure the choices we make are well suited to all basic figures.

gautierronan: Agreed. All good !

dynamiqs/plots/misc.py
gautierronan: Don't know which is which, but the second one is probably best !

dynamiqs/plots/misc.py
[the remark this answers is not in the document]
gautierronan: I meant that if `N_max = 1000` for instance, then `\pm 0.5` is not much space on each side. But do what looks best, I didn't try.

dynamiqs/plots/misc.py
gautierronan: I think I prefer it like that yes, but it's a matter of preference.

dynamiqs/plots/misc.py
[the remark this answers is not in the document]
gautierronan: I meant that if `N_max = 1000` for instance, then `\pm 0.5` is not much space on each side. But do what looks best, I didn't try.

gautierronan (approved):

Issue #237, opened 2023-10-02
Implement custom `sqrtm` for `dm_fidelity`


gautierronan:
See https://arxiv.org/pdf/2201.08663.pdf and https://github.com/KingJamesSong/FastDifferentiableMatSqrt/blob/main/torch_utils.py for the implementation.

This or a Newton method should be more stable and faster than svd/eigh-based implementations.

pierreguilmin:
We closed #243 because the fidelity is ~1%, which is too low.

pierreguilmin:
See https://linear.app/dynamiqs/issue/DYN-87/implement-custom-sqrtm-for-dm-fidelity.

PR #243, opened 2023-10-04
Implement differentiable inv matrix square root

abocquet:
Use a new version of matrix inverse square root, based on https://arxiv.org/abs/2201.08663

It enables auto differentiation and is faster.

Rapid benchmark on a RTX4090
image

Solves #237


gautierronan (approved):
Very nice ! No need for a proper test, but did you check that it computes the right sqrtm in forward and in backward? If so, I think we can merge.

abocquet:
There is a test in Rouchon that ensures the forward and backward pass work correclty. I tried to mess with the code and it crashed so it must be OK

pierreguilmin (changes requested):
I think there is a bug
```python
>>> import torch
>>> from dynamiqs.solvers.utils.utils import inv_sqrtm
>>> n = 4
>>> x = torch.rand((n, n))
>>> xm12 = inv_sqrtm(x)
>>> torch.allclose(x @ xm12 @ xm12, torch.eye(n), rtol=1e-3, atol=1e-3)
False
>>> x @ xm12 @ xm12
tensor([[[ 0.7475,  1.8410, -0.9934, -1.5469],
         [ 0.2476, -0.2559,  0.6473,  0.9988],
         [-0.3816,  1.1147,  0.6527, -1.0096],
         [ 0.1111, -0.7517,  0.3074,  1.7687]]])
```

We should write unit tests for this logic, it's a bit too technical to trust it.

pierreguilmin:
Should we close this @abocquet?

PR #252, opened 2023-10-29
Add null operator

pierreguilmin:
Very handy for e.g. null Hamiltonian, the alternative today to create the null operator are
```python
 H = 0 * dq.destroy(n)
```
or
```python
H = torch.zeros(n, n, dtype=torch.complex64)
```
which are both a little unsatisfactory.


gautierronan (approved):
Very nice, super useful

PR #253, opened 2023-10-29
Change the API of PWC to accept arbitrary times

pierreguilmin:
This is more general, aligns with #239 `plot_pwc_pulse()` function, and is more common (see e.g. `plt.stairs()` in `matplotlib`).

I had to move `check_time_tensor()` to the common `_utils` file.

### DIFF
--- a/dynamiqs/utils/optimal_control.py
+++ b/dynamiqs/utils/optimal_control.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import torch
 from torch import Tensor
 
+from .._utils import check_time_tensor
 from .tensor_types import dtype_complex_to_real, get_cdtype, to_device
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


dynamiqs/utils/optimal_control.py
```diff
@@ -87,58 +88,51 @@ def rand_complex(
...
     Notes:
         You can use [rand_complex()][dynamiqs.rand_complex] to generate a tensor
         filled with random complex numbers for the parameter `values`.
 
     Args:
-        t_start: Start time of the pulse.
-        t_end: End time of the pulse.
-        values _(..., n)_: Pulse complex values where `n` is the number of different
-            values between `t_start` and `t_end`.
+        Times _(n+1)_: Time intervals.
```
gautierronan, line 107:
```suggestion
        times _(n+1)_: Time points between which the pulse takes constant values.
```

gautierronan (changes requested):
Minor comment, otherwise looks good.

PR #254, opened 2023-10-29
Simpler `to_qutip()` API

pierreguilmin:
More straightforward way to define the system dimensions, compatible with `ptrace()`.


gautierronan (dismissed):
LGTM

PR #259, opened 2023-10-29
Improve error message on non-aligned dt

gautierronan:
```python
ValueError: For fixed time step solvers, every value of `tsave` must be a multiple of the time step `dt`, but `dt = 1.00000e-02` and `tsave[1] = 1.11111e-01`.
```


pierreguilmin (changes requested):
We can't do the proposed change, it breaks the `tstop` abstraction, it's very dirty. But this whole `tsave`, `tmeas` business is quite annoying, as a first simplification I propose to directly integrate  `tmeas` in the base `Solver` class (and remove all the weirdo conditional logic and masks of `tstop`). It will just be an unused field for `sesolve` and `mesolve`. See #263. Then you can very easily implement your PR.

gautierronan:
Replaced by #273

PR #261, opened 2023-10-30
Replace options dict API by class API

pierreguilmin:
Here is the new class API.

I quickly fixed the doc but it's just temporary. In particular I removed solver-specific and gradient-specific options that **we have to put back** soon.


gautierronan (approved):
LGTM ! Very nice :)

PR #263, opened 2023-10-31
Simplify solver time logic

pierreguilmin:
- directly integrate  `tmeas` in the base `Solver` class, it will just be an unused field for `sesolve` and `mesolve`
- simplify solver saving using `iteraxis`
- create result object after running the solver

### DIFF
--- a/dynamiqs/solvers/solver.py
+++ b/dynamiqs/solvers/solver.py
@@ -44,7 +44,7 @@ class Solver(ABC):
         self.device = self.options.device
 
         # initialize time logic
-        self.tstop = torch.cat((self.tsave, self.tmeas)).unique().sort()[0]
+        self.tstop = torch.cat((self.tsave, self.tmeas)).unique()
         self.tsave_mask = torch.isin(self.tstop, self.tsave)
         self.tmeas_mask = torch.isin(self.tstop, self.tmeas)
         self.tstop_counter = 0
--- a/dynamiqs/solvers/solver.py
+++ b/dynamiqs/solvers/solver.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from math import inf
 from time import time
 
 import torch
@@ -46,6 +45,8 @@ class Solver(ABC):
 
         # initialize time logic
         self.tstop = torch.cat((self.tsave, self.tmeas)).unique().sort()[0]
+        self.tsave_mask = torch.isin(self.tstop, self.tsave)
+        self.tmeas_mask = torch.isin(self.tstop, self.tsave)
         self.tstop_counter = 0
         self.tsave_counter = 0
         self.tmeas_counter = 0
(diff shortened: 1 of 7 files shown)


dynamiqs/solvers/solver.py
```diff
@@ -89,29 +93,42 @@ def _run(self):
...
     def next_tstop(self) -> float:
         return self.tstop[self.tstop_counter].item()
 
     def save(self, y: Tensor):
-        self._save(y)
-        self.tstop_counter += 1
-
-    def _save(self, y: Tensor):
-        if self.tsave_mask[self.tstop_counter]:
+        if self.next_tstop() == self.next_tsave():
```
gautierronan, line 110: I'm scared of this line and the equivalent one for `tmeas`, with numerical precision and the such in case `tsave` or `tmeas` are modified by anything in the code. I would propose instead to have something like:
```python
def merge_unique_sorted_arrays(X, Y):
    # Combine the arrays and find the unique elements (which are also sorted)
    Z = torch.unique(torch.cat((X, Y)))

    # Check which elements of Z are in X and Y to create the indicator arrays
    from_X = torch.isin(Z, X)
    from_Y = torch.isin(Z, Y)

    return Z, from_X, from_Y

# Test
X = torch.tensor([1, 3, 5, 7])
Y = torch.tensor([2, 3, 6, 8, 9])
Z, from_X, from_Y = merge_unique_sorted_arrays(X, Y)
print("Z:", Z.numpy())
print("From X:", from_X.numpy())
print("From Y:", from_Y.numpy())
# Z: [1 2 3 5 6 7 8 9]
# From X: [ True False  True  True False  True False False]
# From Y: [False  True  True False  True False  True  True]
```
which is called when defining `tstop`. And then one can just check the arrays `from_tsave` and `from_tmeas` to increment `self.tsave_counter` and `self.tmeas_counter`.

gautierronan (changes requested):
Looks good. I have one comment regarding robustness of the checks. It should be easy to modify if you agree.

dynamiqs/solvers/solver.py
pierreguilmin: I agree, this is what was done before, don't know why I changed this part.

dynamiqs/solvers/solver.py
```diff
@@ -41,45 +43,48 @@ def __init__(
         self.rdtype = self.options.rdtype
         self.device = self.options.device
 
-        # initialize saving logic
-        self._init_time_logic()
+        # initialize time logic
+        self.tstop = torch.cat((self.tsave, self.tmeas)).unique().sort()[0]
```
gautierronan, line 47:
```suggestion
        self.tstop = torch.cat((self.tsave, self.tmeas)).unique()
```
FYI you don't need this if `tsave` and `tmeas` are already sorted. But I forget if we already check this, and "au pire ça mange pas de pain"

gautierronan (approved):
LGTM

dynamiqs/solvers/solver.py
pierreguilmin: No that's a good comment, we do check if they are sorted in `sesolve`, `mesolve`.

pierreguilmin: But `unique` sorts everything no matter what.

gautierronan (approved):

PR #264, opened 2023-10-31
New API

abocquet:
Simplifies the API even further, based on @pierreguilmin's [great work](https://github.com/dynamiqs/dynamiqs/pull/261) 
Now to create a new solver, just 

```
dq.mesolve(
    H, jump_ops, rho0, tsave,
    solver=dq.solver.Rouchon(dt=1e-3).adjoint(params=[p1, p2]),
    options=dict(save_states=False, dtype=torch.complex128),
)
```


pierreguilmin:
AFAIK we're not planning to move fwd with this API for now.

PR #271, opened 2023-10-31
Add CD gate

pierreguilmin:


pierreguilmin:
@gautierronan a few questions on this one before the review:
- A batch displacement would be very useful (e.g. for multiple seeds for the SNAP and ECD protocols displacement gate), but it's maybe niche so should we keep a separate method `bdisplace` in this file or merge it with `displace` in `dynamiqs.utils.utils`? Note also that it doesn't work with arbitrary batch size like the rest of the library, it requires a 1D tensor of displacement...
- If you see a simpler way to implement the gate (for batched `alpha`) I'm all ~ears~ eyes.

pierreguilmin:
> It's indeed a bit niche I think. I would say we keep it private for now, and if ever someone asks for it, or it becomes really useful, it'll be simple to make it public.

Agreed, I'm a bit lazy to make it public for now.

>It seems torch.linalg.matrix_exp works with a batch dimension. Maybe you can then write it in a single line without the torch.stack? Actually this may also solve the first question. In this case making it public is less of a debate.

Right, I fixed this. 👍 

> Also, I think the ECD gate is defined as D(beta / 2) |g><g| + D(-beta / 2) |e><e| and not D(beta / 2) |g><e| + D(-beta / 2) |g><e|. See the paper by Campagne-Ibarcq, 2020 Nature. The one by Alec Eickbush might be a typo I think, unless I'm missing something.

They definitions are a bit mixed, let's do the **conditional displacement** gate which corresponds to your formula, I renamed the function for clarity. See appendix of Eickbush paper
Screenshot 2023-11-02 at 18 59 32

dynamiqs/utils/optimal_control.py
```diff
@@ -168,14 +167,97 @@ def snap_gate(
...
+
+    It is defined by
+    $$
+        D(\alpha) = \exp(\alpha a^\dag - \alpha^* a),
+    $$
+    where $a$ and $a^\dag$ are the annihilation and creation operators, respectively.
+
+    Note:
+        The main difference with [dq.displace()][dynamiqs.displace] is that the
+        argument `alpha` can be batched, and the returned tensor is batched accordingly.
```
gautierronan, line 196: If `bdisplace` is strictly better than `displace`, I'd be in favor of merging the two and keeping only the batched version but named `displace`. This means we should do the same for `squeeze`, but it's quite straightforward if you want to do it in the same PR.

gautierronan (changes requested):

gautierronan (approved):
LGTM, thanks for the changes !

PR #277, opened 2023-11-03
Refactor rouchon ME solver

pierreguilmin:
I wasn't very satisfied with what we ended with for the Rouchon solver, the caching felt like a burden while we should actually use it properly to make the code as transparent as possible to the Hamiltonian (constant or time-dependent). I think the main reason was that we sticked with inline lambdas to define what are actually control-flow dependent objects (`M0` and `M1`).

- Now we can use the argument `dt: float` to define `M0` or `M0rev` directly from the forward and backward pass (without creating a bunch of `...rev` objects in the `__init__`).
- I realised we actually need the `sqrtm` normalisation also for time-dependent tensors, because for PWC Hamiltonian it's sometimes better suited than `cholesky` (you just renormalize once for each PWC part). Also they are qualitatively quite different on the tests I made, so `sqrtm` might also be nice for true time-dependent problem where you want to ensure precision.
- The `normalize` option is either `None`, `'sqrt'` or `'cholesky'`. The only "bad case" is if someone chooses `'cholesky'` for a constant tensor, because in this case we should renormalize the Kraus map once, and not the state at every time. But if we document that it's better to use `'sqrtm'` for constant tensors I think it's fine. If you think its' necessary @gautierronan we can also implement back the constant Hamiltonian cholesky normalisation, it will be easier with this structure.
- I had to rework the `@cache` decorator (for the best!) to allow caching on more than one arguments combination (to cache during the backward pass both `M0` on `self.dt` and `M0rev` on `-self.dt`).
- I removed all tensors shapes comments from the bwd pass, I think having them in the fwd pass is enough.
- It should be much easier now to implement cleanly the normalisation for `Rouchon2`.

⚠️ If I'm not mistaken, the only difference with the code on main is https://github.com/dynamiqs/dynamiqs/blob/main/dynamiqs/mesolve/rouchon.py#L161 > @gautierronan  should `M1srev` use `M0` (code on main) or `M0rev` (this PR code)?

```python
# code on main
self.M1s = cache(lambda M0: 0.5 * sqrt(self.dt) * (self.L @ M0 + M0 @ self.L))
self.M1srev = self.M1s

# this PR code
self.M1s = cache(lambda M0: 0.5 * sqrt(self.dt) * (self.L @ M0 + M0 @ self.L))
self.M1srev = cache(lambda M0: 0.5 * sqrt(self.dt) * (self.L @ M0rev + M0rev @ self.L))
```

### DIFF
--- a/dynamiqs/mesolve/rouchon.py
+++ b/dynamiqs/mesolve/rouchon.py
@@ -25,26 +25,28 @@ class MERouchon(MESolver, AdjointFixedSolver):
 
 class MERouchon1(MERouchon):
     @cache(maxsize=2)
-    def R(self, M0: Tensor, dt: float) -> Tensor:
+    def R(self, M0: Tensor, fwd: bool = True) -> Tensor:
         # `R` is close to identity but not exactly, we inverse it to normalize the
         # Kraus map in order to have a trace-preserving scheme
         # -> (b_H, 1, n, n)
+        dt = self.dt if fwd else -self.dt
         return M0.mH @ M0 + dt * self.sum_LdagL
 
     @cache(maxsize=2)
-    def Ms(self, Hnh: Tensor, dt: float) -> tuple(Tensor, Tensor):
+    def Ms(self, Hnh: Tensor, fwd: bool = True) -> tuple(Tensor, Tensor):
         # Kraus operators
         # -> (b_H, 1, n, n), (1, len(L), n, n)
+        dt = self.dt if fwd else -self.dt
         M0 = self.I - 1j * dt * Hnh  # (b_H, 1, n, n)
-        M1s = sqrt(self.dt) * self.L  # (1, len(L), n, n)
+        M1s = sqrt(abs(dt)) * self.L  # (1, len(L), n, n)
 
         if self.options.normalize == 'sqrt':
-            R = self.R(M0, dt)
+            R = self.R(M0, fwd=fwd)
             # we normalize the operators by computing `S = sqrt(R)^-1` and applying it
             # to the `Ms` operators
             S = inv_sqrtm(R)
-            M0 = M0 @ S if dt >= 0 else S @ M0
-            M1s = M1s @ S if dt >= 0 else S @ M1s
+            M0 = M0 @ S if fwd else S @ M0
+            M1s = M1s @ S if fwd else S @ M1s
 
         return M0, M1s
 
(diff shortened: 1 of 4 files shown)


pierreguilmin:
I will implement these changes and both normalisation to the rouchon method for `smesolve` once we have converged on `mesolve`.

dynamiqs/mesolve/rouchon.py
```diff
@@ -24,49 +24,36 @@ class MERouchon(MESolver, AdjointFixedSolver):
...
+        # Kraus map in order to have a trace-preserving scheme
+        # -> (b_H, 1, n, n)
+        return M0.mH @ M0 + dt * self.sum_LdagL
+
+    @cache(maxsize=2)
+    def Ms(self, Hnh: Tensor, dt: float) -> tuple(Tensor, Tensor):
+        # Kraus operators
+        # -> (b_H, 1, n, n), (1, len(L), n, n)
+        M0 = self.I - 1j * dt * Hnh  # (b_H, 1, n, n)
+        M1s = sqrt(self.dt) * self.L  # (1, len(L), n, n)
```
gautierronan, line 39: You use `selt.dt` here but `dt` is a method argument. Throughout the class methods (also in e.g. `forward`), can you normalize the use of `dt` and `self.dt`? Use one or the other every time. This would make more sense IMO.

dynamiqs/mesolve/rouchon.py
```diff
@@ -115,51 +99,44 @@ def backward_augmented(
...
 
-        # compute cached operators
-        # H, Hnh, M0, M0rev: (b_H, 1, n, n)
         H = self.H(t)
         Hnh = self.Hnh(H)
-        M0 = self.M0(Hnh)
-        M0rev = self.M0rev(Hnh)
+
+        # === reverse time
+        M0rev, M1srev = self.Ms(Hnh, -self.dt)
```
gautierronan, line 106: Does this work as intended? I was thinking that `sqrt` of a negative number would raise an error.

dynamiqs/mesolve/rouchon.py
```diff
@@ -115,51 +99,44 @@ def backward_augmented(
...
-            Rrev = self.Rrev(M0rev)
-            Trev = self.Trev(Rrev)
+        if self.options.normalize == 'cholesky':
+            Rrev = self.R(M0rev, -self.dt)
+            Trev = self.T(Rrev)
             rho = inv_kraus_matmul(Trev.mH, rho, upper=True)  # Tr.mH^-1 @ rho @ Tr^-1
 
         # compute rho(t-dt)
-        rho = kraus_map(rho, M0rev) - kraus_map(rho, self.M1srev)
+        rho = kraus_map(rho, M0rev) - kraus_map(rho, M1srev)
```
gautierronan, line 115: If you use a `-dt` in the Kraus operator, then this should be `rho = kraus_map(rho, M0rev) + kraus_map(rho, M1srev)`

gautierronan (changes requested):
Some issues with the sign of the Kraus map in reverse time. Either you use a normal Kraus map (with only additions) and a negative `dt`, or a Kraus map with minus signs, and a positive `dt`.

gautierronan (approved):

PR #318, opened 2023-11-10
[1/2] Split open and closed systems in tests

abocquet:
[1/2] this
[2/2] https://github.com/dynamiqs/dynamiqs/pull/325

### DIFF
--- a/tests/solver_tester.py
+++ b/tests/solver_tester.py
@@ -6,6 +6,7 @@ from typing import Any
 
 import torch
 
+from benchmarks.systems.system import ClosedSystem, OpenSystem
 from dynamiqs.gradient import Gradient
 from dynamiqs.solver import Solver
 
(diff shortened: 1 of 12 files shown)


gautierronan (approved):

pierreguilmin:
Can we wait that I review the PR stack of Ronan's before merging yours @abocquet? I'll review them this afternoon. Cross-merging everything might become a big mess.

tests/solver_tester.py
```diff
@@ -6,46 +6,14 @@
 
 import torch
 
+from benchmarks.systems.system import ClosedSystem, OpenSystem
```
gautierronan, line 9: Why are you importing from the benchmarks module? It should be independent from the tests. I think you mean to import from `sesolve.closed_system.py` and `mesolve.open_system.py`.

gautierronan (changes requested):
Thanks for the PR. I'm not super satisfied with the idea of adding redundent code (with `_test_batching` both in `ClosedSolverTester` and `OpenSolverTester`). But I don't have much better to propose.

If you have an idea to reduce redudancy, I'd be very happy. Otherwise it LGTM apart from the minor comment below.

abocquet:
Thanks for the comment, it was wrong indeed.

As for the redundancy it's less bad (and maybe even better) to have some if it makes the tests easier to read. Let me suggest you https://mtlynch.io/good-developers-bad-tests/ that makes a good point on this issue in general

gautierronan (approved):

PR #322, opened 2023-11-10
Fix cavity tests

pierreguilmin:
Long story short, the tests for cavity were bugged:
- better time now so gradients are not all null at the end of the evolution
- changed `exp_ops` to simplify expressions
- fixed a truncature issue which cause both `Propagator` and `Dopri5` to fail (fixed by taking a smaller alpha)

I think we need to be more careful with the tests, I've started a note to make this a bit more clean.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -41,6 +41,11 @@ $$
 $$
 \nabla_\theta l(t) =
 \begin{pmatrix}
+    \frac{\partial l(t)}{\partial \Delta} \\
+    \frac{\partial l(t)}{\partial \alpha_0} \\
+    \frac{\partial l(t)}{\partial \kappa}
+\end{pmatrix} =
+\begin{pmatrix}
   0.0 \\
   2 \alpha_0 e^{-\kappa t} \\
   -\alpha_0^2 t e^{-\kappa t}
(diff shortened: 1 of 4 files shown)


tests/README.md
```diff
@@ -0,0 +1,67 @@
...
+- Loss:
+
+$$
+l(t) = |\alpha(t)|^2 = \alpha_0^2 e^{-\kappa t}
+$$
+
+- Gradient of the loss:
+
+$$
+\nabla_\theta l(t) =
```
abocquet, line 42:
```suggestion
\nabla_\theta l(t) =
\begin{pmatrix}
\frac{\partial l(t)}{\partial \Delta} \\
\frac{\partial l(t)}{\partial \alpha_0} \\
\frac{\partial l(t)}{\partial \kappa}
\end{pmatrix}
=
```
would be easier to follow

tests/README.md
```diff
@@ -0,0 +1,67 @@
...
+  2 \alpha_0 e^{-\kappa t} \\
+  -\alpha_0^2 t e^{-\kappa t}
+\end{pmatrix}
+$$
+
+- Gradients of the expectation values:
+
+$$
+\begin{aligned}
+    \nabla_\theta \braket{O^{(1)}}(t) &=
```
abocquet, line 54: Same comment here

tests/mesolve/open_system.py
```diff
@@ -83,7 +83,7 @@ def __init__(
         self.H = self.delta * adag @ a
         self.H_batched = [0.5 * self.H, self.H, 2 * self.H]
         self.jump_ops = [torch.sqrt(self.kappa) * a, dq.eye(self.n)]
-        self.exp_ops = [(a + adag) / sqrt(2), 1j * (adag - a) / sqrt(2)]
+        self.exp_ops = [dq.position(self.n), dq.momentum(self.n)]
```
abocquet, line 86: Random thought : should be add a `g` argument in `position` and `momentum` (set to 2 by default)?

abocquet (changes requested):

tests/mesolve/open_system.py
pierreguilmin: Yes, I think it's a good idea 👍

tests/sesolve/closed_system.py
```diff
@@ -194,8 +197,8 @@ def grads_expect(self, t: float) -> Tensor:
         ]).detach()
 
 
-cavity = Cavity(n=8, delta=2 * pi, alpha0=1.0, t_end=1.0)
-gcavity = Cavity(n=8, delta=2 * pi, alpha0=1.0, t_end=1.0, requires_grad=True)
+cavity = Cavity(n=8, delta=2.0, alpha0=0.5, t_end=1.0)
+gcavity = Cavity(n=8, delta=2.0, alpha0=0.5, t_end=1.0, requires_grad=True)
```
abocquet, line 201: Nit: delta is in radians so conventionally we express it in fractions of `2π`

abocquet (approved):

tests/sesolve/closed_system.py
pierreguilmin: Right, I changed it to avoid having `t_end` coincinding with a full rotation, which makes a lot of the gradient we check cancel, and was hiding sign errors. This commit fixes this more clearly and restablishes the $2\pi$ unit.

abocquet (approved):

PR #324, opened 2023-11-10
Backward Euler for sesolve

pierreguilmin:
Too easy to ignore it. The backward Euler method: 1st order, fixed step size, implicit.


abocquet (commented):
Great job going for low hanging fruit! I'm wondering whether `BEuler` is the best name though 🤔 Would `ImplicitEuler` be more explicit (no pun intended)? `BackwardEuler` works too but is a bit confusing as I originally thought it was the backward autodiff with Euler

pierreguilmin:
> Great job going for low hanging fruit! I'm wondering whether `BEuler` is the best name though 🤔 Would `ImplicitEuler` be more explicit (no pun intended)? `BackwardEuler` works too but is a bit confusing as I originally thought it was the backward autodiff with Euler

Right, I don't have a clear opinion on this, I'm fine with `ImplicitEuler` too, wdyt @gautierronan?

gautierronan:
> Right, I don't have a clear opinion on this, I'm fine with `ImplicitEuler` too, wdyt @gautierronan?

"Backward Euler" is the true name of the method. So I think we should keep it like this, and make it clear that it's implicit in the documentation.

gautierronan (approved):
LGTM

abocquet (approved):

PR #325, opened 2023-11-11
[2/3] Batched dissipators

abocquet:
image
⬆️ photon number for a leaky cavity initialized in multiple states and with multiple kappa

- [1/3] https://github.com/dynamiqs/dynamiqs/pull/318
- [2/3] this
- [3/3] (incoming) doc update

Closes DYN-11.


debug.ipynb
gautierronan, line 1: I guess this file should not be on the PR.

dynamiqs/solvers/utils/batching.py
```diff
@@ -5,16 +5,36 @@
...
-    return H.unsqueeze(1)
+    H = H.unsqueeze(1)
+    H = H.unsqueeze(1)
+    return H
 
 
-def batch_y0(y0: Tensor, H: TDTensor) -> Tensor:
+def batch_jump_ops(jump_ops: Tensor) -> Tensor:
+    # handle jump_ops batching
+    # jump_ops: (n_jump_ops, b_jump_ops?, m, n) ->  (n_jump_ops, 1, b_jump_ops, 1, n, n)
```
gautierronan, line 18: I think we should have `jump_ops` of shape `(b_jump_ops?, n_jump_ops, m, n)` instead of `(n_jump_ops, b_jump_ops?, m, n)`. For 2 reasons:

1. We keep the optional dimension outward so the last dimensions are always consistent (batched or not).
2. It's more consistent with other batchable objects, with the batched dimension first.

In case it's more annoying in the backend, we can just swap the two dimensions together right here.

benchmarks/systems/system.py
```diff
@@ -22,6 +24,7 @@ class OpenSystem(ClosedSystem):
     def __init__(self):
         super().__init__()
         self.jump_ops = None
+        self.jump_ops_batched = None
```
gautierronan, line 27: Why is this required in the benchmarks module? This should depend on the benchmarked system if ever we want to have a batched jump_ops.

dynamiqs/mesolve/rouchon.py
```diff
@@ -54,33 +54,34 @@ def Ms(self, Hnh: Tensor, fwd: bool = True) -> tuple(Tensor, Tensor):
     def T(self, R: Tensor) -> Tensor:
         # we normalize the map at each time step by inverting `R` using its Cholesky
         # decomposition `R = T @ T.mT`
-        # -> (b_H, 1, n, n)
+        # -> (b_H, 1, 1, n, n)
```
gautierronan, line 57: `M0` is batched over `L`, so `R(M0)` and `T(R)` are as well. Can you check the dimensions you wrote are consistent with what's actually being returned by the function?

dynamiqs/mesolve/rouchon.py
```diff
@@ -54,33 +54,34 @@ def Ms(self, Hnh: Tensor, fwd: bool = True) -> tuple(Tensor, Tensor):
...
         """
-        # rho: (b_H, b_rho, n, n) -> (b_H, b_rho, n, n)
+        # rho: (b_H, b_L, b_rho, n, n) -> (b_H, b_L, b_rho, n, n)
 
-        H = self.H(t)  # (b_H, 1, n, n)
-        Hnh = self.Hnh(H)  # (b_H, 1, n, n)
-        M0, M1s = self.Ms(Hnh)  # (b_H, 1, n, n), (1, len(L), n, n)
+        H = self.H(t)  # (b_H, 1, 1, n, n)
+        Hnh = self.Hnh(H)  # (b_H, 1, 1, n, n)
+        M0, M1s = self.Ms(Hnh)  # (b_H, 1, 1, n, n), (len(L), 1, b_L, 1, n, n)
```
gautierronan, line 75: Same comment here. I think there are mistakes in the shapes. Can you check them?

dynamiqs/mesolve/rouchon.py
```diff
@@ -54,33 +54,34 @@ def Ms(self, Hnh: Tensor, fwd: bool = True) -> tuple(Tensor, Tensor):
...
 
         # normalize the Kraus Map
         if self.options.normalize == 'cholesky':
-            R = self.R(M0)  # (b_H, 1, n, n)
-            T = self.T(R)  # (b_H, 1, n, n)
+            R = self.R(M0)  # (b_H, 1, 1, n, n)
+            T = self.T(R)  # (b_H, 1, 1, n, n)
```
gautierronan, line 80: Same here.

dynamiqs/mesolve/rouchon.py
```diff
@@ -135,10 +136,19 @@ class MERouchon2(MERouchon):
     @cache(maxsize=2)
     def Ms(self, Hnh: Tensor, fwd: bool = True) -> tuple(Tensor, Tensor):
         # Kraus operators
-        # -> (b_H, 1, n, n), (b_H, len(L), n, n)
+        # -> (b_H, 1, 1, n, n), (len(L), b_H, b_L, 1, n, n)
         dt = self.dt if fwd else -self.dt
-        M0 = self.I - 1j * dt * Hnh - 0.5 * dt**2 * Hnh @ Hnh  # (b_H, 1, n, n)
-        M1s = 0.5 * sqrt(abs(dt)) * (self.L @ M0 + M0 @ self.L)  # (b_H, len(L), n, n)
+        M0 = self.I - 1j * dt * Hnh - 0.5 * dt**2 * Hnh @ Hnh  # (b_H, 1, 1, n, n)
```
gautierronan, line 141: Same here.

dynamiqs/mesolve/rouchon.py
```diff
@@ -135,10 +136,19 @@ class MERouchon2(MERouchon):
...
+
+        batched_mult_left = lambda x, y: torch.einsum('n...ik,...kj->n...ij', x, y)
+        batched_mult_right = lambda x, y: torch.einsum('...ik,n...kj->n...ij', x, y)
```
gautierronan, line 144: Can you move these functions either at the top of the file, or in the utils?

dynamiqs/solvers/utils/utils.py
```diff
@@ -29,12 +29,15 @@ def kraus_map(rho: Tensor, O: Tensor) -> Tensor:
...
+        operators: Kraus operators of shape `(a, b, c, n, n)` or `(x, a, b, c, n, n)`.
 
     Returns:
         Density matrix of shape `(a, ..., n, n)` with the Kraus map applied.
     """
-    return torch.einsum('abij,a...jk,abkl->a...il', O, rho, O.mH)
+
+    if O.dim() == 5:
+        O = O.unsqueeze(0)
+    return torch.einsum('n...ik,...kl,n...lj->...ij', O, rho, O.mH)
```
gautierronan, line 40: I don't think this achieves the same operation as described in the docstring. One of the two needs to be fixed.

gautierronan: Also, I guess `c` is `b_jump_ops` but I didn't understand what the `x` dimension was?

dynamiqs/utils/vectorization.py
```diff
@@ -214,4 +214,4 @@ def slindbladian(H: Tensor, L: Tensor) -> Tensor:
     Returns:
         _(..., n^2, n^2)_ Lindbladian superoperator.
     """
-    return -1j * (spre(H) - spost(H)) + sdissipator(L).sum(-3)
+    return -1j * (spre(H) - spost(H)) + sdissipator(L).sum(0)
```
gautierronan, line 217: If we change the shape ordering of `jump_ops`, then this should be reverted.

gautierronan (changes requested):
Great PR, super useful ! Some comments below.

dynamiqs/solvers/utils/batching.py
abocquet: We decided to ease the most common path here, which is not batched jump operators. The easiest way of passing mutiple jump operators is through a list (stacking tensors requires more code and a little gymnastic). So the best way of writing in the most common case remains `dq.mesolve(H, [L1, L2], ...)` IMO.

If we chose to put the batching direction outside we'd have to require
- either the jump operators to always be reshaped with `(b_jump_ops?, n_jump_ops, m, n)`, which is annoying
- or to have a different syntax for batched and not batched operators
- or to provide with a list of list of tensors if we want to keep not batched operators, which also sounds like a bad idea

Also I'm not sure we have other batchable and stackable (ie there are mulitple) arguments. In this case it is more consistant with other batched object to NOT put the `n_jump_ops` outside this it's not a batching dimension but a "stacking" one. Otherwise you'd have
- `y0`: `(b_H, b_L, b_y0, m, n)`
- `H`: `(b_H, 1, 1, m, n)`
- `L`: `(b_H, b_L, n_L, 1, m, n)`
which is an additional complication

benchmarks/systems/system.py
abocquet: In `test/systems` `self.H_batched` is also defined so it's a matter of consistency. In general it's good practice to define all fields in the `__init__` method of a class in Python

dynamiqs/mesolve/rouchon.py
abocquet: Nice catch!

dynamiqs/solvers/utils/utils.py
abocquet: `c` is `b_jump_ops` and `x` is `n_jump_ops`

abocquet: I hotfixed the doc but it's still not perfect. I think we should talk about it IRL

dynamiqs/solvers/utils/batching.py
gautierronan: Indeed I didn't think of the fact that we pass jump operators as a list. In this case I'm good with keeping this order.

One question: how do we handle the fact that we may want multiple to pass multiple jump operators that are batched? Or alternatively, when we want to batch over a single jump operator and not all?

For instance, I want to simulate a dissipative cat qubit with `L2 = sqrt(kappa2) * (a^2 - alpha^2)` (with `kappa2` fixed), but also batch over `L1 = sqrt(kappa1) a` (with `kappa1` batched). Should I pass `L2` in a batch with a similar shape as `L1`? Should only `L1` be batched? 

Alternatively, what if I want to batch over `kappa2` **and** `kappa1`? What is the accepted input shape?

I don't think there's a particular preferred choice, but we need to make one. Could you clarify this in the PR? Maybe add checks that the input is of the required shape format?

dynamiqs/solvers/utils/utils.py
gautierronan: Ok, I get it now. So I think we need to have two separate operations:

1. `kraus_map` which computes `torch.einsum('n...ik,...kl,n...ij', O, rho, O.mH)`.
2. `kraus_matmul` which computes `torch.einsum('...ik,...kl,...ij', O, rho, O.mH)`, i.e. without the sum. I think this second one is completely equivalent to `O @ rho @ O.mH`, but to be checked.

In both cases, the `...` clearly designate all batched dimensions which are common to `O` and `rho`. This way, it's more robust if we ever decide to have additional batching dimensions, and we don't have to check `O.dim() == 5`.

Could introduce these two operators and make the appropriate changes? I guess we only need `kraus_matmul` for the product of `M0 @ rho @ M0.mH`.

gautierronan (changes requested):
Another comment below regarding the kraus_map operation. On top of this PR, could also make another PR to fix

1. the documentation of `mesolve`, `sesolve` and `smesolve`
2. the "batching" tutorial

dynamiqs/solvers/utils/batching.py
abocquet: This is a good question. As of now all dissipators are expected to have the same batching. That being said, it's still possible to code your example with something like this (might not work out of the box but you'll get the idea):
```python
kappa_2 = 10
kappa_1 = torch.linspace(0, 1, 11)

L2  = kappa_2 * (a @ a).repeat(len(kappa_1), 1, 1)
L1  = kappa_1 a.repeat(len(kappa_1), 1, 1)

qt.mesolve(H, [L2, L1], ...)
```

It's not as straight forward as specifying `L2 = kappa_2 * a @ a` but it may be a good thing, because otherwise there is a topic on how you handle dissipators with different batch sizes etc...

If you want to batch over $\kappa_2$ and $\kappa_1$, probably something like this would work:
```
kappa_2 = torch.linspace(1, 10, 10)
kappa_1 = torch.linspace(0, 1, 11)

kappa_2, kappa_1 = torch.meshgrid(kappa_2, kappa_1)
kappa_2 = kappa_2.ravel()
kappa_1 = kappa_1.ravel()

L2  = kappa_2 * (a @ a).repeat(len(kappa_1), 1, 1)
L1  = kappa_1 a.repeat(len(kappa_1), 1, 1)

qt.mesolve(H, [L2, L1], ...)
```

It's not as easy as if we direclty made the product in `mesolve` but doing so would disqualify all cases where we want to have hand made batching (like `(x1, y1), (x1, y2), (x2, y1)` but not `(x2, y2)`). To simplify a bit I would add a helper method that would make the carthesian product of several batched tensors.

I'll add the check for same dimensionality in this PR but let's keep API changes (if we want them) for another one

abocquet:
Absolutely, I will also adress how to do dissipators batching once we agree on how we want to handle it

dynamiqs/solvers/utils/batching.py
pierreguilmin: Sorry, don't have the time to do a full review right now (but pls wait for me for merging @abocquet), but on this specific point (multi-batched Hamiltonian, initial state on jump operators), I think we should:
- keep the current `sesolve`, `mesolve`, `smesolve` APIs, it's going to be unreadable with arbitrarily mulyi-batched `H`, `rho0` and `L`
- one day, if needed, provide simple utility functions to convert a multi-batched object to a single-batched object and the reverse (because internally we'll always handle a large Hamiltonian), but it's a bit of an edge case

So this feature is not really missing from the API, we just miss a bit of sugar. What IS missing is the ability to run non-product simulations of batched parameters (i.e. we run today all H of a batched Hamiltonian with all rho0 of a batched initial states, you can't _choose_ specific configurations). I think it's quite an edge case though, I don't think we want to complexify the API or put effort on this.

abocquet: 100% agreed, I think we are on the same page

Conclusion : 
- until further notice the niche examples Ronan gave are to be done with the code examples in my n-1 message (give or take)
- multi-batching is something that we should think of someday but until there is a usecase/something else to do, let's put it in a corner of our heads and talk about it on the occasion

gautierronan: If I understood correctly, we only accept that all jump operators have the same batching dimension for now. Can you just add a check that this is indeed the case somewhere, and raise an error if not?

dynamiqs/solvers/utils/utils.py
gautierronan: I think it's really better not to use `einsum` when it's not required. It's a complicated function which is particularly inefficient for small-sized systems. At some point we will probably remove all `einsum`s from the code.

So here, using a simple `@` product when it's possible should be favored I think. It also avoids having an `.unsqueeze()` under the hood. This way `kraus_matmul` stays quite straightforward.

dynamiqs/solvers/utils/batching.py
gautierronan: Actually, to simplify usability, I would propose that the following two ways of batching are allowed:

1. All jump operators have the same batching dimension `b_L`.
2. A single jump operator has a batching dimension `b_L`, and all others have no batching dimensions (`ndim=2`).

In case 2, we simply batch all other jump operators internally with the same shape. All other batching shapes should be disallowed. WDYT?

pierreguilmin: I'm fine with either propositions (either force all batched with same dim, or accept both all batched with same dim and a single one batched). Let's not document this new feature in mesolve documentation or in the batching tutorial for now, it needs a bit more thinking and experimentations in the backend before going public.

abocquet: Addressed in https://github.com/dynamiqs/dynamiqs/pull/325/commits/f20ab9860564027a287fbd8aea8dcc80f46302bb

dynamiqs/solvers/utils/utils.py
abocquet: This goes againt the comment left by @pierreguilmin  in the Kraus map documentation:
```
This is equivalent to `torch.sum(operators @ rho[None,...] @ operators.mH, dim=0)`.
    The use of einsum yields better performances on large matrices, but may cause a
    small overhead on smaller matrices (N <~ 50).
```
I am not litterate on this so until we have a significant benchmark to prove one side or the other I'll leave it as is

dynamiqs/mesolve/mesolve.py
```diff
@@ -160,6 +160,36 @@ def mesolve(
...
+    batched_jump_ops, not_batched_jump_ops = [], []
+
+    for jump_op in jump_ops:
+        if jump_op.dim() == 3:
+            batched_jump_ops.append(jump_op)
+        elif jump_op.dim() == 2:
+            not_batched_jump_ops.append(jump_op)
+        else:
+            raise ValueError(
+                f"All jump operators must be of shape 2 or 3, got shape {jump_op.shape}"
```
gautierronan, line 174:
```suggestion
                f"All jump operators must have 2 dimensions or 3 dimensions if batched, but a jump operator had {jump_op.ndim} dimensions with shape {jump_op.shape}."
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -160,6 +160,36 @@ def mesolve(
             'Argument `jump_ops` must be a non-empty list, otherwise consider using'
             ' `sesolve`.'
         )
+
+    # check all jump ops are batched in a similar way or not at all
+    batched_jump_ops, not_batched_jump_ops = [], []
+
+    for jump_op in jump_ops:
+        if jump_op.dim() == 3:
+            batched_jump_ops.append(jump_op)
```
gautierronan, line 169:
```suggestion
            if jump_op.shape[0] == 1:
                not_batched_jump_ops.append(jump_op.squeeze(0))
            else:
                batched_jump_ops.append(jump_op)
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -160,6 +160,36 @@ def mesolve(
             'Argument `jump_ops` must be a non-empty list, otherwise consider using'
             ' `sesolve`.'
         )
+
+    # check all jump ops are batched in a similar way or not at all
```
gautierronan, line 164: This whole piece of code should probably not be inside `mesolve`, but inside `batch_jump_ops` to have all batching taken care of in the same place.

dynamiqs/mesolve/mesolve.py
gautierronan: (i.e. allow a jump operator of shape (1, n, n))

dynamiqs/solvers/utils/utils.py
gautierronan: But here this is not a `torch.sum` that is required, but really just a `@` which is always faster than einsum (at least for square matrices). 

Anyway, I'll fix this in a subsequent PR.

dynamiqs/solvers/utils/batching.py
```diff
@@ -5,16 +5,36 @@
 
 def batch_H(H: TDTensor) -> TDTensor:
     # handle H batching
-    # x: (b_H?, n, n) ->  (b_H, 1, n, n)
+    # x: (b_H?, n, n) ->  (b_H, 1, 1, n, n)
     if H.dim() == 2:  # (n, n)
         H = H.unsqueeze(0)  # (1, n, n)
```
gautierronan, line 10:
```suggestion
    if H.dim() == 2:  # (n, n)
        return H[None, None, None, ...]
    elif H.dim() == 3: # (b_H, n, n)
        return H[:, None, None, ...]
    else:
        raise ValueError(f"Expected `H` of dimension 2 or dimension 3 if batched, got {H.ndim} dimensions with shape {H.shape}.")
```

gautierronan: IMO it makes it more explicit how we're batching instead of using repeated `.unsqueeze` calls. Also added a check for the dimension. Could you do the same for `batch_y0` and `batch_jump_ops` ?

gautierronan (changes requested):
Here's another set of comments regarding batching functions. Sorry for the multi-review. After these comments I'm good I think !

To-do in subsequent PRs:

- [ ] Fix documentation of `mesolve` and `smesolve` regarding batching
- [ ] Fix tutorial of batching
- [ ] Fix the use of einsum for `M0 @ rho @ M0.mH` in Rouchon solvers.

dynamiqs/mesolve/mesolve.py
abocquet: It's not easy to put evything in `batch_jump_ops` because it would then have to also include `to_tensor` inside of it and make a function with a lot of arguments and much bigger than what we do for H and rho. Instead, I made another function `prepare_jump_ops_batching` that does the pre-formatting

dynamiqs/solvers/utils/batching.py
abocquet: I cannot do it on `H` because it's not a `Tensor` but a `TDTensor` (which only has `unsqueeze` and does not support slicing). It's cleaner for L and rho0 so I'll do it

abocquet:
I fixed  the use of einsum for M0 @ rho @ M0.mH in Rouchon solvers, I had not understood the fact that the sum was not there in this case

dynamiqs/solvers/utils/batching.py
gautierronan: Ah right I forgot !

dynamiqs/solvers/utils/utils.py
```diff
@@ -28,13 +28,14 @@ def kraus_map(rho: Tensor, O: Tensor) -> Tensor:
     TODO Fix documentation
 
     Args:
-        rho: Density matrix of shape `(a, ..., n, n)`.
-        operators: Kraus operators of shape `(a, b, n, n)`.
+        rho: Density matrix of shape `(a, b, c, n, n)`.
+        operators: Kraus operators of shape `(x, a, b, c, n, n)`.
```
gautierronan, line 32:
```suggestion
        rho: Density matrix of shape `(..., n, n)`.
        operators: Kraus operators of shape `(a, ..., n, n)`.
```

dynamiqs/solvers/utils/utils.py
```diff
@@ -28,13 +28,14 @@ def kraus_map(rho: Tensor, O: Tensor) -> Tensor:
...
 
     Returns:
-        Density matrix of shape `(a, ..., n, n)` with the Kraus map applied.
+        Density matrix of shape `(x, a, b, c, n, n)` with the Kraus map applied.
```
gautierronan, line 35:
```suggestion
        Density matrix of shape `(..., n, n)` with the Kraus map applied.
```

gautierronan (changes requested):
Last minor comments on the documentation of `kraus_map`. Thanks for all the changes !

gautierronan (dismissed):
Thanks @abocquet for the PR ! Nice work :) We'll just need to fix the CI which currently seems bugged.

@pierreguilmin Can you make a quick review, and then we merge?

gautierronan (dismissed):

dynamiqs/mesolve/mesolve.py
```diff
@@ -168,18 +169,18 @@ def mesolve(
         )
 
     # format and batch all tensors
-    # H: (b_H, 1, n, n)
-    # rho0: (b_H, b_rho0, n, n)
+    # H: (b_H, 1, 1, n, n)
+    # jump_ops: (1, b_jump_ops, 1, n, n)
```
pierreguilmin, line 173: Let's rename `b_jump_ops` in `b_L` for consistency with the rest of the code. I know it's not what's done for the state batching `b_rho0`, we could fix it as well but I propose not to do it in this PR, which is already quite cluttered with comments update.

pierreguilmin: Also in `smesolve`.

dynamiqs/mesolve/rouchon.py
```diff
@@ -152,21 +186,24 @@ def forward(self, t: float, rho: Tensor) -> Tensor:
...
         """
-        # rho: (b_H, b_rho, n, n) -> (b_H, b_rho, n, n)
+        # rho: (b_H, b_L, b_rho, n, n) -> (b_H, b_L, b_rho, n, n)
 
-        H = self.H(t)  # (b_H, 1, n, n)
-        Hnh = self.Hnh(H)  # (b_H, 1, n, n)
-        M0, M1s = self.Ms(Hnh)  # (b_H, 1, n, n), (b_H, len(L), n, n)
+        H = self.H(t)  # (b_H, 1, 1, n, n)
+        Hnh = self.Hnh(H)  # (b_H, 1, 1, n, n)
+        M0, M1s = self.Ms(Hnh)  # (b_H, 1, 1, n, n), (len(L), b_H, b_L, 1, n, n)
```
pierreguilmin, line 199:
```suggestion
        M0, M1s = self.Ms(Hnh)  # (b_H, b_L, 1, n, n), (len(L), b_H, b_L, 1, n, n)
```

dynamiqs/smesolve/rouchon.py
```diff
@@ -33,10 +33,10 @@ def __init__(self, *args, **kwargs):
...
     def forward(self, t: float, rho: Tensor) -> Tensor:
-        # rho: (b_H, b_rho, ntrajs, n, n) -> (b_H, b_rho, ntrajs, n, n)
+        # rho: (b_H, b_L, b_rho, ntrajs, n, n) -> (b_H, b_L, b_rho, ntrajs, n, n)
 
         # compute cached operators
-        # H, Hnh, M0_tmp: (b_H, 1, ntrajs, n, n)
+        # H, Hnh, M0_tmp: (b_H, b_L, 1, ntrajs, n, n)
         H = self.H(t)
         Hnh = self.Hnh(H)
         M0_tmp = self.M0_tmp(Hnh)
```
pierreguilmin, line 42: Fix line 51
```diff
- # M0: (b_H, 1, ntrajs, n, n)
+ # M0: (b_H, b_L, 1, ntrajs, n, n)
```

dynamiqs/smesolve/smesolve.py
```diff
@@ -217,18 +217,18 @@ def smesolve(
         )
```
pierreguilmin, line 217: Can you remove the inline comments indicating returned tensors size just below the docstring? Also in sesolve, they're outdated.

dynamiqs/sesolve/sesolve.py
```diff
@@ -163,8 +163,8 @@ def sesolve(
 
     # get saved tensors and restore correct batching
     if result.ysave is not None:
-        result.ysave = result.ysave.squeeze(0, 1)
+        result.ysave = result.ysave.squeeze(0, 1, 2)
     if result.exp_save is not None:
-        result.exp_save = result.exp_save.squeeze(0, 1)
+        result.exp_save = result.exp_save.squeeze(0, 1, 2)
```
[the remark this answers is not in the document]
pierreguilmin: Ho because you also batched `H` along an extra dimension for `sesolve`. What about making moving small batching functions specifc to `sesolve` on top of the file? Otherwise we need to fix all the comments of the SE solvers 😭

dynamiqs/smesolve/smesolve.py
```diff
@@ -217,18 +217,18 @@ def smesolve(
...
+    # H: (b_H, 1, 1, 1, n, n)
+    # rho0: (b_H, b_jump_ops, b_rho0, ntrajs, n, n)
     # exp_ops: (len(exp_ops), n, n)
-    # jump_ops: (len(jump_ops), n, n)
+    # jump_ops: (len(jump_ops), b_jump_ops, n, n)
     H = to_td_tensor(H, dtype=options.cdtype, device=options.device)
     rho0 = to_tensor(rho0, dtype=options.cdtype, device=options.device)
     H = batch_H(H).unsqueeze(2)
-    rho0 = batch_y0(rho0, H).unsqueeze(2).repeat(1, 1, ntrajs, 1, 1)
+    jump_ops = batch_jump_ops(jump_ops, dtype=options.cdtype, device=options.device)
```
pierreguilmin, line 227: Don't you need to unsqueeze this guy too? Did you test that the SME solver still works? I can provide a code example if needed.

dynamiqs/mesolve/me_solver.py
```diff
@@ -8,12 +8,13 @@
 class MESolver(Solver):
     def __init__(self, *args, jump_ops: Tensor):
         super().__init__(*args)
-        self.L = jump_ops[None, ...]  # (1, len(L), n, n)
-        self.sum_LdagL = (self.L.mH @ self.L).squeeze(0).sum(dim=0)  # (n, n)
+        self.L = jump_ops  # (len(L), 1, b_L, 1, n, n)
+        # self.sum_LdagL = (self.L.mH @ self.L).squeeze(0).sum(dim=0)  # (n, n)
+        self.sum_LdagL = torch.einsum("l...ik,l...kj->...ij", self.L.mH, self.L)
```
pierreguilmin, line 13:
```suggestion
        self.sum_LdagL = torch.einsum('l...ik,l...kj->...ij', self.L.mH, self.L)
```

dynamiqs/mesolve/rouchon.py
```diff
@@ -19,6 +19,34 @@ def inv_kraus_matmul(A: Tensor, B: Tensor, upper: bool) -> Tensor:
...
+    r"""Computes the batched matrix product of two tensors where the left tensor
+    has one more dimension than the right one.
+     Args:
+         x: Tensor of shape (N, ..., n, k)
+         y: Tensor of shape (..., k, m)
+
+     Returns:
+        The batched product of `x @ y`
+         `(N, ..., n, m)`.
+    """
```
pierreguilmin, line 32: Pls fix extra spaces, missing newlines and missing end points (taking as reference `dynamiqs.utils.utils` docstrings for example).

dynamiqs/mesolve/rouchon.py
```diff
@@ -54,33 +82,33 @@ def Ms(self, Hnh: Tensor, fwd: bool = True) -> tuple(Tensor, Tensor):
...
         if self.options.normalize == 'cholesky':
-            R = self.R(M0)  # (b_H, 1, n, n)
-            T = self.T(R)  # (b_H, 1, n, n)
+            R = self.R(M0)  # (b_H, b_L, 1, n, n)
+            T = self.T(R)  # (b_H, b_L, 1, n, n)
             rho = inv_kraus_matmul(T.mH, rho, upper=True)  # T.mH^-1 @ rho @ T^-1
 
         # compute rho(t+dt)
-        rho = kraus_map(rho, M0) + kraus_map(rho, M1s)  # (b_H, b_rho, n, n)
+        rho = M0 @ rho @ M0.mH + kraus_map(rho, M1s)  # (b_H, b_L, b_rho, n, n)
```
pierreguilmin, line 111: The rational for the einsum for M0 was that it's faster, let's keep this for now, I'll write smwh that we need to fix it.

dynamiqs/solvers/utils/batching.py
```diff
@@ -1,20 +1,89 @@
...
-    return H.unsqueeze(1)
+    H = H.unsqueeze(1)
+    H = H.unsqueeze(1)
+    return H
+
+
+def batch_jump_ops(jump_ops: List[ArrayLike], dtype=None, device=None) -> Tensor:
```
pierreguilmin, line 20: Use `list` instead of `List`, and `from __future__ import annotations` at the top of the file to use new-style typing. Pls also type `dtype` and `device`.

dynamiqs/solvers/utils/batching.py
```diff
@@ -1,20 +1,89 @@
...
-def batch_y0(y0: Tensor, H: TDTensor) -> Tensor:
+    jump_ops = torch.stack(jump_ops)
+    if jump_ops.dim() == 3:  # (n_jump_ops, n, n)
+        return jump_ops[:, None, None, None, ...]  # (n_jump_ops, 1, 1, 1, n, n)
+    elif jump_ops.dim() == 4:  # (n_jump_ops, b_jump_ops, n, n)
+        return jump_ops[:, None, :, None, ...]  # (n_jump_ops, 1, b_jump_ops, 1, n, n)
+    else:
+        raise ValueError(
+            "Expected `jump_ops` of dimension 2 or dimension 3 if batched, got"
+            f" {jump_ops.ndim} dimensions with shape {jump_ops.shape}."
```
pierreguilmin, line 66: Use single quotes.

dynamiqs/mesolve/rouchon.py
gautierronan: Are you sure? I'm quite skeptical that einsum is faster if we're not summing over the first dimension.

abocquet: We need to do benchmarks.

dynamiqs/solvers/utils/batching.py
abocquet: We need a coding guidelines document to write down all these things

tests/solver_tester.py
```diff
@@ -158,12 +159,39 @@ def _test_batching(
...
+
+        # batched jump_ops and y0
+        result = run(system.H, system.jump_ops_batched, system.y0_batched)
+        assert result.ysave.shape == (b_jump_ops, b_y0, ntsave, m, n)
+        assert result.exp_save.shape == (b_jump_ops, b_y0, n_exp_ops, ntsave)
+
+        # batched H and jump_ops
+        result = run(system.H_batched, system.jump_ops_batched, system.y0_batched)
+        assert result.ysave.shape == (b_H, b_jump_ops, b_y0, ntsave, m, n)
+
```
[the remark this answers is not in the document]
abocquet: It was there but the comment was wrong 🤗

dynamiqs/smesolve/smesolve.py
abocquet: Yes an example would be good please 🙏

dynamiqs/utils/vectorization.py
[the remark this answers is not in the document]
abocquet: I feel it's more natural to have the multiple jump ops dimension on the left because one will define all jump ops and then put them in a list. People will do so for `mesolve` for instance so even in term of API consistency it makes more sense to use `L _(N, ..., n, n)`.
I don't want to  argue on this so you choose.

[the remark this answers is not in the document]
abocquet: In the meantime I made a private function for this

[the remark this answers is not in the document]
gautierronan: For consistency with `mesolve`, we should modify the API of `slindbladian` to accept a list of jump operators, so of format `list[Tensor]`, where each Tensor can be batched. And in this case, I agree with @abocquet.

pierreguilmin (approved):
I'm happy to merge this if you're satisfied @abocquet! 😊

pierreguilmin:
Merging to move fwd, we can always fix small things later 🙂

PR #328, opened 2023-11-12
Refactor adjoint: minor fixes to adjoint solver [1/4]

pierreguilmin:
This is the first of a stack of PR that aims at clarifying and re-organising the ODE solvers code, notably the part related to the adjoint.

**This PR**: A bunch of minor fixes, to read commit by commit, notably:
- 6093471b2f48949b5a926c18db797055263dda5a > `AdjointAutograd` needs `odefun_backward` and `odefun_adjoint` which are not defined by the `solver` class in use
- ef1bd35281271aa3f66b07eea733cab77b4069ac > @abocquet why did we add this? in any case if we want to add it back, this should happen as upstream as possible in the code, i.e. directly in the options class

We slowly start moving things around to make a smooth transition to the new code.

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -182,11 +182,11 @@ def mesolve(
     jump_ops = to_tensor(jump_ops, dtype=options.cdtype, device=options.device)
 
     # convert tsave to a tensor
-    tsave = to_tensor(tsave, dtype=options.rdtype, device=options.device)
+    tsave = to_tensor(tsave, dtype=options.rdtype, device='cpu')
     check_time_tensor(tsave, arg_name='tsave')
 
     # define the solver
-    tmeas = torch.empty(0, dtype=options.rdtype, device=options.device)
+    tmeas = torch.empty(0, dtype=options.rdtype, device='cpu')
     solver = SOLVER_CLASS(H, rho0, tsave, tmeas, exp_ops, options, jump_ops=jump_ops)
 
     # compute the result
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -51,7 +51,7 @@ class AdaptiveSolver(AutogradSolver):
         `self.tstop`."""
 
         # initialize the progress bar
-        self.pbar = tqdm(total=self.tstop[-1].item(), disable=not self.options.verbose)
+        self.pbar = tqdm(total=self.tstop[-1], disable=not self.options.verbose)
 
         # initialize the step counter
         self.step_counter = 0
@@ -63,7 +63,7 @@ class AdaptiveSolver(AutogradSolver):
 
         # run the ODE routine
         t, y, ft = self.t0, self.y0, f0
-        for tnext in self.tstop.cpu().numpy():
+        for tnext in self.tstop:
             y, ft, dt, error = self.integrate(t, tnext, y, ft, dt, error)
             self.save(y)
             t = tnext
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -23,6 +23,15 @@ class AdaptiveSolver(AutogradSolver):
         Series in Computational Mathematics.
     """
 
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        # initialize the progress bar
+        self.pbar = tqdm(total=self.tstop[-1], disable=not self.options.verbose)
+
+        # initialize the step counter
+        self.step_counter = 0
+
     @property
     @abstractmethod
     def order(self) -> int:
@@ -49,12 +58,6 @@ class AdaptiveSolver(AutogradSolver):
         """Integrates the ODE forward from time `self.t0` to time `self.tstop[-1]`
         starting from initial state `self.y0`, and save the state for each time in
         `self.tstop`."""
-
-        # initialize the progress bar
-        self.pbar = tqdm(total=self.tstop[-1], disable=not self.options.verbose)
-
-        # initialize the step counter
-        self.step_counter = 0
 
         # initialize the ODE routine
         f0 = self.odefun(self.t0, self.y0)
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -168,6 +168,17 @@ class AdaptiveSolver(AutogradSolver):
 
 
 class AdjointAdaptiveSolver(AdaptiveSolver, AdjointSolver):
+    @abstractmethod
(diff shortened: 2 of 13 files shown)


dynamiqs/solvers/ode/adaptive_solver.py
```diff
@@ -14,18 +14,46 @@
...
+        Series in Computational Mathematics.
+    """
+
+    @property
+    @abstractmethod
+    def order(self) -> int:
+        pass
+
+    @property
+    @abstractmethod
```
gautierronan, line 41: I think `tableau` should not be an abstract method of `AdaptiveSolver`. In principle, we should have `AdaptiveSolver` -> `RungeKutta` -> DormandPrince5` and then we could have `tableau` an abstract method of `RungeKutta`. For now we have a single adaptive method, and the next ones will also be Runge Kuttas with Butcher tableaus. But we could also have Adaptive Solvers without Butcher tableaus.

gautierronan, line 41: I think `tableau` should not be an abstract method of `AdaptiveSolver`. In principle, we should have `AdaptiveSolver` -> `RungeKutta` -> DormandPrince5` and then we could have `tableau` an abstract method of `RungeKutta`. For now we have a single adaptive method, and the next ones will also be Runge Kuttas with Butcher tableaus. But we could also have Adaptive Solvers without Butcher tableaus.

dynamiqs/solvers/ode/adaptive_solver.py
```diff
@@ -168,15 +172,22 @@ def update_tstep(self, dt: float, error: float) -> float:
 
 
 class AdjointAdaptiveSolver(AdaptiveSolver, AdjointSolver):
+    """Integrate an augmented ODE of the form $(1) dy / dt = fy(y, t),
+    $(2) da / dt = fa(a, y)$ in backward time with initial condition $y(t_0)$ using an
```
gautierronan, line 179: "reverse" time and not "backward".

dynamiqs/solvers/ode/fixed_solver.py
```diff
@@ -56,36 +65,44 @@ def run_autograd(self):
...
         for t in times[:-1].cpu().numpy():
             y = self.forward(t, y)
             self.pbar.update(self.dt)
 
         return y
 
+
+class AdjointFixedSolver(FixedSolver, AdjointSolver):
+    """Integrate an augmented ODE of the form $(1) dy / dt = fy(y, t)$ and
+    $(2) da / dt = fa(a, y)$ in backward time with initial condition $y(t_0)$ using a
```
gautierronan, line 91: Same here: "reverse time"

dynamiqs/solvers/ode/fixed_solver.py
```diff
@@ -31,14 +31,8 @@ class FixedSolver(AutogradSolver):
...
         super().__init__(*args)
+
         self.dt = self.options.dt
-        if isinstance(self.dt, Tensor):
-            if self.dt.numel() == 1:
-                self.dt = self.dt.item()
-            else:
-                raise ValueError(
-                    f'`dt` should be a number or a 0-d tensor, but is {self.dt}.'
-                )
```
gautierronan: We indeed need to check that `dt` is a float, and not a Tensor. Otherwise this was creating a bug somewhere in the adjoint (forgot where) and also it respects the required typing.

`dt` is a Tensor for instance if you pass:
```python
dq.solver.Rouchon(dt=tsave[1]-tsave[0])
```

Can you add back the conversion from 0-D Tensor to float where you think it's appropriate? (in the options class maybe).

dynamiqs/mesolve/mesolve.py
```diff
@@ -182,11 +182,11 @@ def mesolve(
     jump_ops = to_tensor(jump_ops, dtype=options.cdtype, device=options.device)
 
     # convert tsave to a tensor
-    tsave = to_tensor(tsave, dtype=options.rdtype, device=options.device)
+    tsave = to_tensor(tsave, dtype=options.rdtype, device='cpu')
```
gautierronan, line 185: Since we use `tsave` as a numpy array (in order to keep floats everywhere), shouldn't we convert it to a numpy array from the get-go, instead of calling `.numpy()` slightly later in the code? 

Or otherwise we can move `tsave` to the GPU and never call `.cpu.numpy()`. This was my preferred solution and still is, from this debate we had some time ago. Just mentionning it in case you changed your mind.

gautierronan (changes requested):
Really great PR. Lots of useful changes. Some minor comments below.

dynamiqs/solvers/ode/adaptive_solver.py
pierreguilmin: Fixed in #331. 👍

dynamiqs/solvers/ode/adaptive_solver.py
pierreguilmin: Nice remark 👌 I didn't add the `abstractmethod` here, it was there before, let me write your comment down to fix in an independent PR later, and introduce an `AdaptiveRungeKutta` class.

dynamiqs/solvers/ode/fixed_solver.py
pierreguilmin: Ok, I add it back. Do we really need to check that it's a 0-D tensor with one element? It seems a bit far fetched that someone would put a 1d tensor for dt no?

I propose to add in the options
```diff
class _ODEFixedStep(Solver):
    SUPPORTED_GRADIENT = (Autograd,)

    def __init__(self, *, dt: float):
+        # convert dt because it's common to pass a 0-D tensor instead of a float
+        if isinstance(dt, Tensor):
+            dt = dt.item()
        self.dt = dt
```

dynamiqs/solvers/ode/adaptive_solver.py
gautierronan: Indeed ! Probably from me actually !

dynamiqs/mesolve/mesolve.py
pierreguilmin: Yes we should, I was just lazy to recode the following line `check_time_tensor` for NumPy arrays.

> Or otherwise we can move tsave to the GPU and never call .cpu.numpy(). This was my preferred solution and still is, from this debate we had some time ago. Just mentionning it in case you changed your mind.
Let me think about it, I propose to keep this aside for now and choose either fix later (make times Tensor everywhere, or convert to numpy directly).

dynamiqs/solvers/ode/fixed_solver.py
gautierronan: Indeed, as long as `dt.item()` works (and it should for N-D tensors as long as they have 1 element), it's good.

dynamiqs/mesolve/mesolve.py
gautierronan: Sound good !

gautierronan (approved):

PR #329, opened 2023-11-12
Refactor adjoint: add base ODE solver class [2/4]

pierreguilmin:
The _integrate_ abstraction is great, but was not pushed further for the fwd pass. It's funny because it was pushed to the end for the adjoint backward pass (with `integrate_augmented`). We just write a proper base class for the ODE solvers to abstract this properly, now the _integrate_ logic is exactly the same for the fwd and bwd pass. Next PR will fix the bwd _integrate_ abstraction which is a bit clunky with the current code.


gautierronan (approved):
Looks good !

PR #333, opened 2023-11-13
Fix `wigner` wrong device/dtype for clenshaw

abocquet:
Fix `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!` when running a simulation on CUDA


pierreguilmin:
I think it's better to create the tensor directly on the correct device rather than transferring after creation. Also typing will
be incorrect here (complex instead of float).

pierreguilmin (approved):

pierreguilmin:
@gautierronan the dtype/device are fixed for clenshaw method but not fft, I'll add it to the backlog.

pierreguilmin:
Thanks a lot for the fix @abocquet 👍

PR #335, opened 2023-11-15
Add basis, alias of fock

gautierronan:
Add `dq.basis` and `dq.basis_dm` to replicate the API of qutip. (I'm personally more used to using `qt.basis` so I keep making the mistake when using dynamiqs).

### DIFF
--- a/dynamiqs/utils/states.py
+++ b/dynamiqs/utils/states.py
@@ -10,7 +10,7 @@ from .operators import displace
 from .tensor_types import ArrayLike, Number, get_cdtype, to_tensor
 from .utils import tensprod, todm
 
-__all__ = ['fock', 'fock_dm', 'coherent', 'coherent_dm']
+__all__ = ['fock', 'fock_dm', 'basis', 'basis_dm', 'coherent', 'coherent_dm']
 
 
 def fock(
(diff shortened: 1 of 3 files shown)


dynamiqs/utils/states.py
```diff
@@ -97,6 +97,28 @@ def fock_dm(
...
 
 
+def basis(
+    dim: int | tuple[int, ...],
+    number: int | tuple[int, ...],
+    *,
+    dtype: torch.complex64 | torch.complex128 | None = None,
+    device: str | torch.device | None = None,
+):
+    """Alias for [`dq.fock()`](/python_api/utils/states/fock.html)."""
```
pierreguilmin, line 107: Correct syntax is
```suggestion
    """Alias for [`dq.fock()`][dynamiqs.fock]."""
```

pierreguilmin (changes requested):
I'm good otherwise 👍

abocquet (approved):

dynamiqs/utils/states.py
gautierronan: I tried this but the link was broken for some reason, both in the main function page or in the API summary table.

gautierronan: Nevermind it works if you use `[dynamiqs.fock]` and I was using `(dynamiqs.fock)`. Fixed.

abocquet (approved):

pierreguilmin (approved):

PR #336, opened 2023-11-15
Add ndim and shape to TDTensor

gautierronan:
Just a simple addition to `TDTensor` API which could be useful at some point (e.g. in https://github.com/dynamiqs/dynamiqs/pull/325).


abocquet (approved):
I love it 🔥

PR #337, opened 2023-11-15
Fix cuda and cuda:0 bug

gautierronan:
Fix this quite annoying bug that arises from:
```python
import torch
x = torch.randn(3, device='cuda')
print(x.device == torch.device('cuda')) # False
print(x.device == torch.device('cuda:0')) # True
```
Typically occurs when we do `H.to('cuda')` instead of `H.to('cuda:0')`

### DIFF
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -21,16 +21,21 @@ def obj_type_str(x: Any) -> str:
 
 def to_device(device: str | torch.device | None) -> torch.device:
     if device is None:
-        return torch.ones(1).device  # default device
+        torch_device = torch.ones(1).device  # default device
     elif isinstance(device, str):
-        return torch.device(device)
+        torch_device = torch.device(device)
     elif isinstance(device, torch.device):
-        return device
+        torch_device = device
     else:
         raise TypeError(
             'Argument `device` must be a string, a `torch.device` or `None` but has'
             f' type {obj_type_str(device)}.'
         )
+
+    if torch_device.index:
+        return torch_device
+    else:
+        return torch.device(torch_device.type, 0)  # default device index to 0
 
 
 def hdim(x: Tensor) -> int:
--- a/dynamiqs/solvers/utils/td_tensor.py
+++ b/dynamiqs/solvers/utils/td_tensor.py
@@ -41,7 +41,6 @@ def check_callable(
     expected_device: torch.device,
 ):
     # check type, dtype and device match
-
     if not isinstance(x0, Tensor):
         raise TypeError(
             f'The time-dependent operator must be a {type_str(Tensor)}, but has type'


abocquet (commented):
Do we know why PyTorch doesn't do this already? There must be a reason and it may apply to us too

abocquet:
Then it seems a bit reckless to "fix" it by ourselves without knowing why it has been done so IMHO

pierreguilmin:
If you have two GPUs you can't select on which one the code runs now 😕 I agree that we need a fix though.

dynamiqs/_utils.py
```diff
@@ -21,17 +21,22 @@ def obj_type_str(x: Any) -> str:
...
-        return device
+        torch_device = device
     else:
         raise TypeError(
             'Argument `device` must be a string, a `torch.device` or `None` but has'
             f' type {obj_type_str(device)}.'
         )
 
+    if torch_device.type == 'cuda' and not torch_device.index:
+        return torch.device(torch_device.type, index=0)  # default cuda index to 0
```
pierreguilmin, line 36: You should use `torch.cuda.current_device()` instead of `0`.

pierreguilmin (changes requested):
Good catch!

gautierronan:
> Then it seems a bit reckless to "fix" it by ourselves without knowing why it has been done so IMHO

I don't get your comment. We're only fixing our own device check here.

dynamiqs/_utils.py
```diff
@@ -21,17 +21,22 @@ def obj_type_str(x: Any) -> str:
...
         raise TypeError(
             'Argument `device` must be a string, a `torch.device` or `None` but has'
             f' type {obj_type_str(device)}.'
         )
 
+    if torch_device.type == 'cuda' and not torch_device.index:
+        return torch.device(torch_device.type, index=torch.cuda.current_device())
+    else:
+        return torch_device
+
```
pierreguilmin, line 39:
```suggestion
    if torch_device.type == 'cuda' and not torch_device.index:
        torch_device.index = torch.cuda.current_device()
    
    return torch_device
```

dynamiqs/_utils.py
```diff
@@ -21,17 +21,22 @@ def obj_type_str(x: Any) -> str:
...
     elif isinstance(device, torch.device):
-        return device
+        torch_device = device
     else:
         raise TypeError(
             'Argument `device` must be a string, a `torch.device` or `None` but has'
             f' type {obj_type_str(device)}.'
         )
 
+    if torch_device.type == 'cuda' and not torch_device.index:
```
pierreguilmin, line 35:
```suggestion
    if torch_device.type == 'cuda' and torch_device.index is None:
```

pierreguilmin (changes requested):

PR #338, opened 2023-11-16
Use attribute doc style for properties docstring

pierreguilmin:
Minor fix to f2e021c.


pierreguilmin:
cc @gautierronan for next time, it's just a convention 😉

PR #345, opened 2023-11-19
Improve sharp bits documentation page

pierreguilmin:
Small pass on the sharp bits page that I didn't review yet (#305). It's nice! I just reorganised a bit to make every message crystal clear, and add some details.

PR #346, opened 2023-11-19
Fix pycon lexer not recognised by sybil

pierreguilmin:
On top of #345.

We were not testing the examples under the `pycon` fences. ~@gautierronan can we juste use `python` as proposed in this PR, or should we try an find a workaround to keep `pycon`?~ --> Added the `pycon` parser to sybil.

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -4,7 +4,12 @@ import pytest
 import torch
 from matplotlib import pyplot as plt
 from sybil import Sybil
-from sybil.parsers.myst import DocTestDirectiveParser, PythonCodeBlockParser, SkipParser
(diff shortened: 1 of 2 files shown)


gautierronan:
Well the syntax highlighting is not quite the right one if we don't use `pycon`. Is it hard to make a workaround?

pierreguilmin:
> Well the syntax highlighting is not quite the right one if we don't use `pycon`. Is it hard to make a workaround?

Right, let me see.

pierreguilmin:
Fixed, it's super simple to add a custom parser to sybil 👍

gautierronan (approved):

PR #350, opened 2023-11-20
Split plot functions into multiple files

pierreguilmin:
I had to tweak around a little bit to make it work with the doc, this is why I introduced the `namespace` file. Also the ordering of the table in Python API was incorrect, that's why I set it manually.


gautierronan (approved):
I didn't catch why you introduced `namespace`. Otherwise LGTM

pierreguilmin:
> I didn't catch why you introduced namespace. Otherwise LGTM

Well otherwise we have a lot of subsections in the doc for the plotting section, with only one or two functions per subsection.

PR #351, opened 2023-11-20
Fix adjoint

gautierronan:
Adjoint tests were bugged for 2 reasons:

1. Time was not being evolved in the right direction inside `solver.step()` function
2. The graphs of `fy` and `fa` were being freed at every time step

Fixes DYN-77.

### DIFF
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -165,11 +165,19 @@ class AdjointAdaptiveSolver(AdaptiveSolver, AdjointODESolver):
         """Returns $fa(a, t)$."""
         pass
 
+    def _odefun_backward(self, t: float, y: Tensor) -> Tensor:
+        # t is passed in as negative time
+        return self.odefun_backward(-t, y)
+
+    def _odefun_adjoint(self, t: float, y: Tensor) -> Tensor:
+        # t is passed in as negative time
+        return self.odefun_adjoint(-t, y)
+
     def init_augmented(self, t0: float, y0: Tensor, a0: Tensor) -> tuple:
-        fy0 = self.odefun_backward(t0, y0)
-        fa0 = self.odefun_adjoint(t0, a0)
-        dty0 = self.init_tstep(t0, y0, fy0, self.odefun_backward)
-        dta0 = self.init_tstep(t0, a0, fa0, self.odefun_adjoint)
+        fy0 = self._odefun_backward(t0, y0)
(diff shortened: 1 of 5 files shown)


dynamiqs/solvers/ode/adaptive_solver.py
```diff
@@ -201,18 +201,13 @@ def integrate_augmented(
...
             # next line, because there are no more references to the original tensors
-            # (the old `y`, `a`, `fyt` and `fat` go out of scope)
-            y, a, fyt, fat = (
-                new_leaf_tensor(y),
-                new_leaf_tensor(a),
-                new_leaf_tensor(fyt),
-                new_leaf_tensor(fat),
-            )
+            # (the old `y` and `a` go out of scope)
+            y, a = new_leaf_tensor(y), new_leaf_tensor(a)
```
pierreguilmin, line 205: Aren't we going to store a large computation graph to track `fyt` and `fat`?

pierreguilmin (commented):
Nice 😍 Sorry, I think I was the one who introduced these two errors with my refactoring PRs. Here's a small proposal to abstract the negative time for the people implementing the solvers.

dynamiqs/solvers/ode/adaptive_solver.py
gautierronan: TBH I don't understand this part so well. For sure if you add back the `new_leaf_tensor` call on `fyt` and `fat` the tests don't pass anymore. Also, it doesn't seem like the run time of the backward pass explodes, so I think we're not passing through the graph of `fyt` and `fat` multiple times. 

I'm guessing they go out of scope naturally and so their graph is detached by PyTorch.

gautierronan:
Not a big fan of https://github.com/dynamiqs/dynamiqs/pull/351/commits/3e4f23efa14592060e4b24268110807e1ade4aa5. I think it adds complexity to the solver (which is already quite complicated) for not much. If our users want to implement a custom solver which works with the adjoint, they will have to make the effort of understanding the adjoint code anyway. 

As you wish, we can keep it if you prefer it this way.

dynamiqs/solvers/ode/adaptive_solver.py
pierreguilmin: Right, but I think it would be great if we could make it explicit using the `new_leaf_tensor` function, or a comment when it goes out of scope. Otherwise it's very empirical, and unreadable for someone who don't know the code. We can do this in another PR though, nothing urgent.

pierreguilmin (approved):

PR #352, opened 2023-11-20
Fix adjoint memory usage

gautierronan:
Following up on this comment https://github.com/dynamiqs/dynamiqs/pull/351#discussion_r1399121270. There was still a bug in the adjoint which follows from a poor management of the computation graph. We were backpropagating through the entire graph every time step (and the graph was increasing in size at each time step), so backprop was exploding.

This fixes the bug, and also the tests now pass really cleanly (even with `atol=rtol=1e-6`). 

However, the fix is not entirely satisfying because it requires recomputing the functions f(y, t) and f(a, t) twice at each time step during the backward pass. This is not a big overhead so it's a good enough fix for now, but still quite annoying.  Happy to hear other solutions.

### DIFF
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -205,9 +205,12 @@ class AdjointAdaptiveSolver(AdaptiveSolver, AdjointODESolver):
             y, a = new_leaf_tensor(y), new_leaf_tensor(a)
 
             with torch.enable_grad():
+                # compute derivatives
+                fyt, fat = self.odefun_backward(t, y), self.odefun_adjoint(t, a)
+
                 # perform a single step of size dt
-                fyt_new, y_new, y_err = self.step(t, y, fyt, dt, self.odefun_backward)
-                fat_new, a_new, a_err = self.step(t, a, fat, dt, self.odefun_adjoint)
+                _, y_new, y_err = self.step(t, y, fyt, dt, self.odefun_backward)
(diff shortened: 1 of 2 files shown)


dynamiqs/solvers/ode/adaptive_solver.py
```diff
@@ -205,9 +205,12 @@ def integrate_augmented(
             y, a = new_leaf_tensor(y), new_leaf_tensor(a)
```
pierreguilmin, line 205: Pretty sure we can add back the following now (it's really this line that enforces the graph to be disconnected between step n and n+1)
```suggestion
            # the computation graph attached to `y`, `a`, `fyt` and `fat` is
            # automatically freed after next line, because there are no more references
            # to the original tensors (the old `y`, `a`, `fyt` and `fat` go out of
            # scope)
            y, a, fyt, fat = (
                new_leaf_tensor(y),
                new_leaf_tensor(a),
                new_leaf_tensor(fyt),
                new_leaf_tensor(fat),
            )
```

pierreguilmin (changes requested):
Brilliant!

dynamiqs/solvers/ode/adaptive_solver.py
gautierronan: I mean, we can add it, but it's useless I think. Here you are intantiating `fyt` as a leaf tensor, and in the next line, we're saying that `fyt` is instead `self.odefun_backward(t, y)`. So we're assigning another value.

pierreguilmin: Correct!

pierreguilmin (approved):

PR #353, opened 2023-11-21
Small fixes to doc

gautierronan:
Various small changes to the doc. Slightly modified the basic examples to make them simpler/more canonical. I also removed the "uncomment for GPU" lines since we'll soon have a dedicated tutorial, and I wasn't sure it was working (at least it was not being checked by doctest).


README.md
```diff
@@ -57,9 +56,6 @@ jump_ops = [np.sqrt(kappa) * a]
 rho0 = qt.coherent_dm(n, alpha0)
 tsave = np.linspace(0, 1.0, 101)
 
-# uncomment the next line to run the simulation on GPU
-# torch.set_default_device('gpu')
-
```
pierreguilmin, line 62: > I also removed the "uncomment for GPU" lines since we'll soon have a dedicated tutorial, and I wasn't sure it was working (at least it was not being checked by doctest).

I think we should put back this line here and in the next example. The main goal of these two examples is to showcase the main features of the library in the smallest amount of code. It's important that people who don't know GPUs or PyTorch see that they just need one extra line and everything works the same.

pierreguilmin (changes requested):
Nice!

README.md
gautierronan: But does it work? We don't need to specify `device='cuda'` in the `options`? I didn't try but I'm quite skeptical this won't raise an error.

gautierronan: I added it back on the first example. Note that the GPU example was not working, because we had `torch.set_default_device('gpu')` instead of `torch.set_default_device('cuda')`.

gautierronan:
@pierreguilmin An idea on the CI? It's passing locally but not on github. No idea what the error message is saying.

README.md
pierreguilmin: Thanks! (related to https://linear.app/dynamiqs/issue/DYN-107/add-a-gpu-runner-to-the-ci for this kind of testing)

pierreguilmin: I'd say we should add back the line to both examples (bc it shows that you can also do the gradient computation on GPU, as simply as running the simulation).

pierreguilmin:
> @pierreguilmin An idea on the CI? It's passing locally but not on github. No idea what the error message is saying.

It breaks for unrelated reasons on my PR too, I guess it's a new version of sybil or smght like that. I didn't have the time to investigate yet 🤷🏼‍♂️

docs/getting_started/examples.md
```diff
@@ -26,7 +26,7 @@ rho0 = qt.coherent_dm(n, alpha0)
 tsave = np.linspace(0, 1.0, 101)
 
 # uncomment the next line to run the simulation on GPU
-# torch.set_default_device('gpu')
+# torch.set_default_device('cuda')
```
abocquet, line 29:
```suggestion
torch.set_default_device('cuda' if torch.cuda.is_available() else 'cpu')
```
would be even better

abocquet (changes requested):

README.md
```diff
@@ -58,7 +58,7 @@ rho0 = qt.coherent_dm(n, alpha0)
 tsave = np.linspace(0, 1.0, 101)
 
 # uncomment the next line to run the simulation on GPU
-# torch.set_default_device('gpu')
+# torch.set_default_device('cuda' if torch.cuda.is_available() else 'cpu')
```
pierreguilmin, line 61: Maybe we can keep the line uncommented with @abocquet proposed fix, and change the comment
```python
# run the simulation on a GPU if available, otherwise use the CPU
torch.set_default_device('cuda' if torch.cuda.is_available() else 'cpu')
```

pierreguilmin (approved):

pierreguilmin (approved):

PR #355, opened 2023-11-22
Wavy homepage

gautierronan:
Inspired from https://codepen.io/raykuo/pen/QgaKrr. You should `task docserve` locally to see the PR.

Not sure I like it, but has the benefit of simplicity. I looked into three.js but it seems really complicated to make something good-looking.

Related to DYN-104.


pierreguilmin:
Niiiice 😍 I like it! A few remarks if you want to put more time on it, otherwise I'm fine with moving fwd:
- On large screens the wave goes a bit too high and overlaps with the text. I thin ideally it should be at the center/just above the three text box.
Screenshot 2023-11-22 at 19 41 53 (2)
Screenshot 2023-11-22 at 19 41 53 1(2)
- It'd be nice to add a lil' more complexity (more shadowed small waves).
- Could we have the gradient back on the orange part? or does it look weird?

Concerning the overall theme, I have other ideas, but it makes a lot of sense to use a wave (function)!

abocquet:
Here is a cool example with code also : https://codepen.io/soju22/pen/xxGqJxj

gautierronan:
New version. Tell me what you think @pierreguilmin @abocquet.

Screenshot 2023-11-23 at 15 04 19

Screenshot 2023-11-23 at 15 04 23

pierreguilmin:
I think I prefer the first version. The colors were a bit flashier, and the wave was more surprising with the large amplitude.

**v1**
https://github.com/dynamiqs/dynamiqs/assets/38159029/ee3fb42e-958f-4816-962d-b94f90d32408

**v2**
https://github.com/dynamiqs/dynamiqs/assets/38159029/32963a70-8413-46cc-912c-e60574658214

abocquet:
Thanks for the movies Pierre
I prefer the second one because it's faster

abocquet (dismissed):
That's really cool! Maybe leave more orange space between the logo and the wave (this is purely subjective so take it or not)

gautierronan:
@pierreguilmin @abocquet Here's a V3. I created custom wave patterns for all 4 waves. Tell me what you think.

abocquet:
I think there is even less space than before 😅

gautierronan:
> I think there is even less space than before 😅 Looks awesome otherwise

I agree, improved in the last commit.

pierreguilmin:
Nice 😍 Much better than v2 IMO. We can merge if you're satisfied, minor subjective comments below:

- I still prefer the first version, because it felt the whole page was dynamic, here it just looks like a moving band, but if you both like it we can start with that, I prefer it than the current static version anyway.
- Also, I liked better the more flashy old colors (that matched the logo and the rest of the doc), but same it's as you prefer.

Screenshot 2023-11-24 at 13 27 43
Screenshot 2023-11-24 at 13 27 31

pierreguilmin (approved):

pierreguilmin:
Also FYI the result looks quite different between small (top) vs large (bottom) screens

Screenshot 2023-11-24 at 13 30 37

Screenshot 2023-11-24 at 13 30 28 (2)

gautierronan:
> Nice 😍 Much better than v2 IMO. We can merge if you're satisfied, minor subjective comments below:
> 
>     * I still prefer the first version, because it felt the whole page was dynamic, here it just looks like a moving band, but if you both like it we can start with that, I prefer it than the current static version anyway.
> 
>     * Also, I liked better the more flashy old colors (that matched the logo and the rest of the doc), but same it's as you prefer.

My opinion is that I think the last version is less intrusive than the first, and I kinda prefer the tuned down colors (more smoothing, less aggressive). But I'm open to either changes.

PR #362, opened 2023-11-25
Add even cat example to `plot_fock()`

pierreguilmin:
Also allows to demonstrate the use of `xticksall` and setting `ymax` to `None`.

Screenshot 2023-11-25 at 12 16 00


gautierronan (approved):
Shouldn't this be the default behavior actually? And then users can pass `ymax=1.0` if they want?

pierreguilmin:
> Shouldn't this be the default behavior actually? And then users can pass `ymax=1.0` if they want?

IMO it's like the wigner, we plot a quantity which is mathematically bounded by 1.0, so this is the default, and then if one wants to "zoom" in he can choose to do it. I think this default also allows directly comparing plots (i.e. you can easily compare two states distributions e.g. fock 1 and a coherent state without changing the arguments).

abocquet (approved):

PR #363, opened 2023-11-25
Simplify batching internally

pierreguilmin:
Closes DYN-132. The PR looks big but it's mostly due to renamings, there's not much new logic.

Now we never specify dimensions in solvers comments, just `(..., n, n)`. All objects passed to `Solver` must be matmul compatible, the rest will work.

The saver to handle our "maybe batched tensors" is `torch.view`, it's a very neat way to handle both the batched or not batched case: if `H` has shape `(bH?, n, n)`, `H = H.view(-1, 1, n, n)` has shape `(bH, 1, n, n)` with `bH = 1` if not batched (no need to conditionally unsqueeze).

I also took the chance to do some renamings for consistency. The philosophy is that we have a set of names for the outside world (`sesolve`, `mesolve`, `smesolve` APIs: `H`, `psi0`/`rho0`, `jump_ops`, ...) and a set of internal names for convenience (`H`, `y0`, `L`)
- in `sesolve.py/mesolve.py/smesolve.py` renamed asap `psi0`/`rho0` to `y0`, and `jump_ops` to `L`
- renamed `jump_ops` argument of `Solver` to `L` directly
- batch size naming fixed and unified everywhere: `(len(L), 1, b_L, 1, n, n)` -> `(nL, 1, bL, 1, n, n)`

The code to format `L` was condensed, one drawback is that errors are not as comprehensive as before. I think however we should implement proper error handling for shape compatibilities (see DYN-51), and meanwhile it's ok if we skip it IMO (it will be fully refactored anyway once we fix DYN-51).


pierreguilmin:
If you guys agree I'd also like to do the following renamings internally (in another PR):
- `exp_ops` > `E` (like `L` for `jump_ops`), and we will have `Esave` instead of `exp_save` (more symmetric with `ysave`), and we can use `nE` instead of `len(exp_ops)` to specify shapes
- `meas_save` > `Lmsave` (we use `Lm` for measured `jump_ops`), again more symmetric and we can use `nLm` instead of `len(meas_save)` to specify shapes

dynamiqs/smesolve/smesolve.py
```diff
@@ -204,38 +205,52 @@ def smesolve(
...
+    tmeas = to_tensor(tmeas, **kw)
+    check_time_tensor(tmeas, arg_name='tmeas', allow_empty=True)
 
-    # convert etas to a tensor and check
-    # etas: (len(L), 1, 1, 1, 1)
-    etas = to_tensor(etas, **time_kwargs)
+    # === convert and check etas
+    # etas: (nL, ...)
+    etas = to_tensor(etas, dtype=options.rdtype, device=options.device)
     etas = etas[..., None, None, None, None]
```
gautierronan, line 254:
```suggestion
    etas = etas.view(-1, 1, 1, 1, 1)
```

gautierronan (approved):
Very nice ! Thanks for taking care of all this :) Will simplify everything in the future.

gautierronan (approved):

dynamiqs/smesolve/rouchon.py
```diff
@@ -19,40 +19,32 @@ class SMERouchon1(SMERouchon):
...
-        H = self.H(t)
-        Hnh = self.Hnh(H)
-        M0_tmp = self.M0_tmp(Hnh)
+        H = self.H(t)  # (..., n, n)
+        Hnh = self.Hnh(H)  # (..., n, n)
+        M0_tmp = self.M0_tmp(Hnh)  # (..., n, n)
 
         # sample Wiener process
-        dw = self.sample_wiener(self.dt)  # (len(Lm), b_H, b_L, b_rho, ntrajs)
+        dw = self.sample_wiener(self.dt)  # (nLm, ...)
```
abocquet, line 38: the last dimension is still `ntrajs` no?

abocquet (approved):

dynamiqs/smesolve/rouchon.py
pierreguilmin: Yes, but for smesolve I also included `ntrajs` into `...`. For cartesian batching we will have `... = bH, bL, by, ntrajs` and for stacked batching `... = bHLy, ntrajs`.

PR #364, opened 2023-11-26
Small fixes to documentation

gautierronan:
Two small fixes to documentation:
- Fix `rho0` to `psi0` in readme examples (thought I had done so in the previous PR, but apparently not)
- Move around utility functions such that the API table ordering matches the ordering in the left-side menu (it's only copy-pasting around, not actual changes).


gautierronan:
> IMO we should choose the ordering from most useful to least used functions in `__all__` of each file, and make the API table ordering pick that somehow. It's quite annoying to tie the ordering to the arbitrary ordering of the API table, you never know where you should add a method, and can't choose the best ordering.

Not sure how to do that, but we can add the "members" list below the table to select the ordering. It's just a bit more text.

pierreguilmin:
I think the [CMasher package](https://cmasher.readthedocs.io/index.html) has some nice cyclic colormap (we'd need to change the figure background color for the two last ones):
18187fcd-8285-4908-a78d-6f0b5cfa17e6

@abocquet any you like? (copper is my favorite)

pierreguilmin (approved):
I fixed the ordering here and there to align everything (file, __all__, mkdocs.yml and API table).

PR #365, opened 2023-11-27
Experimental implementation of time-dependent `jump_ops`

pierreguilmin:
This should not be merged with `main`, it's just a patch.

If one of the `jump_ops` is a function with signature `(t: float) -> Tensor`, this PR picks a modified adaptive solver that supports time-dependent `jump_ops`. No support for batched `jump_ops`.

@AnilMurani and @Tzekh there you go, if you pull this branch you can use time-dependent jump operators, while we find time to implement proper support for it.

Related to DYN-19.

Example:
```python
import dynamiqs as dq
import numpy as np

n = 32
alpha0 = np.sqrt(2)
alpha1 = 2
k2 = 4
a = dq.destroy(n)
H = dq.zero(n)

def L(t):
    if t <= 1/3:
        return dq.zero(n)
    elif 1/3 <= t <= 2/3:
        return np.sqrt(k2) * (a @ a - alpha0**2 * dq.eye(n))
    else:
        return np.sqrt(k2) * (a @ a - alpha1**2 * dq.eye(n))

jump_ops = [L]

psi0 = dq.coherent(n, 0.0)
tsave = np.linspace(0, 1, 12)
```
```python
dq.plot_wigner_mosaic(result.states, n=12, nrows=3, cross=True)
```
cdcefacc-1265-4184-9f5f-a8324d6b3afc

### DIFF
--- a/dynamiqs/mesolve/adaptive_t.py
+++ b/dynamiqs/mesolve/adaptive_t.py
@@ -1,0 +1,74 @@
+from __future__ import annotations
+
+import torch
+from torch import Tensor
+
+from ..solvers.ode.adaptive_solver import AdjointAdaptiveSolver, DormandPrince5
+from ..solvers.solver import Solver
+from ..solvers.utils import cache
+
+
+class METSolver(Solver):
+    def __init__(self, *args, L: Tensor):
+        super().__init__(*args)
+        self.L = L  # (nL, ..., n, n)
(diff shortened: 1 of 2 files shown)


abocquet:
@pierreguilmin this PR should probably be closed because of JAX support, WDYT?

PR #366, opened 2023-11-29
dyn 133 implement stacked batching

abocquet:
- Add flat batching
- Clean code

PR #369, opened 2023-12-06
Fix flat batching

pierreguilmin:
On top of https://github.com/dynamiqs/dynamiqs/pull/366. Related to DYN-133.

- The tests were coming hard to read IMO, I know it’s usually better to be explicit but we are in a situation where we need to test tons of configuration, so I think it's more readable to factor. Moreover it will help not copy-pasting all these lines for the incoming `smesolve` tests.
- The way we test the batching is a bit outdated, I added a minimal test for the flat batching which is good enough for now IMO.
- I refactord a bit how we handle the `flat_batching` option in `mesolve` to remove repeated code and to have a uniformized code with `sesolve` and `smesolve`.
- The `smesolve` batching is broken but I'm a lil' tired, I'll fix it once I have written proper tests for `smesolve`, because it's super annoying to fix without them.

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -10,7 +10,7 @@ from ..solver import Dopri5, Euler, Propagator, Rouchon1, Rouchon2, Solver
 from ..solvers.options import Options
 from ..solvers.result import Result
 from ..solvers.utils.td_tensor import to_td_tensor
-from ..solvers.utils.utils import format_L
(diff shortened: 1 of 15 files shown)


abocquet (approved):

abocquet:
Simple nit : it would have been faster to review if it had been split into 
- change existing code
- add support for sesolve and smesolve

gautierronan (approved):
Nice ! I had a few comments on the smesolve side but I'll wait for the fix.

PR #372, opened 2023-12-07
Add `TimeTensor` [1/4]: replace `TDTensor` by `TimeTensor`

pierreguilmin:
Related to DYN-137. I'd like to stack a few PRs before merging this one to make sure everything works well with this new abstraction. I'm pushing this one for review to make sure you guys agree with the foundations.

**This PR** replaces `TDTensor` by `TimeTensor`. It re-implements `ConstantTimeTensor` and `CallableTimeTensor`, now with support for left/right addition with another `Tensor` or `TimeTensor` and left/right multiplication with a scalar (`Number | Tensor`).

Note that we won't implement all the additions between all different types, we will just need internally additions between any `TimeTensor` and `Tensor`, addition between any `TimeTensor` and `ConstantTimeTensor`, and addition between any `TimeTensor` subclass and the same subclass.

Left to do before merging the PR stack:
- [x] Add `TimeTensor` [2/4] > implement `PWCTimeTensor`
- [x] Add `TimeTensor` [3/4] > implement `ParameterizedTimeTensor`
- [x] Add `TimeTensor` [4/4] > ~~fix the doc~~ fix caching


dynamiqs/time_tensor.py
```diff
@@ -0,0 +1,257 @@
...
+__all__ = ['tt']
+
+
+def to_time_tensor(
+    x: ArrayLike | callable[[float], Tensor],
+    *,
+    dtype: torch.dtype | None = None,
+    device: str | torch.device | None = None,
+) -> TimeTensor:
+    dtype = get_cdtype(dtype) if dtype is None else dtype  # assume complex by default
```
abocquet, line 21:
```suggestion
    dtype = dtype or get_cdtype(None)
```

dynamiqs/time_tensor.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @abstractmethod
+    def adjoint(self) -> TimeTensor:
+        f = lambda t: self.f(t).adjoint()
+        f0 = self.f0.adjoint()
+        return CallableTimeTensor(f, f0)
+
+    def __neg__(self) -> TimeTensor:
+        f = lambda t: -self.f(t)
+        f0 = -self.f0
```
abocquet, line 225: Why is this necessary ? It doesn't impact the shape

dynamiqs/time_tensor.py
```diff
@@ -0,0 +1,257 @@
...
+        return CallableTimeTensor(f, f0)
+
+    def __mul__(self, other: Number | Tensor) -> TimeTensor:
+        super().__mul__(other)
+        f = lambda t: self.f(t) * other
+        f0 = self.f0 * other
```
abocquet, line 231: Same

dynamiqs/time_tensor.py
```diff
@@ -0,0 +1,257 @@
...
+def time_tensor_add(x: TimeTensor, y: Tensor | TimeTensor) -> TimeTensor:
+    if isinstance(x, ConstantTimeTensor):
+        if isinstance(y, Tensor):
+            return ConstantTimeTensor(x.tensor + y)
+        elif isinstance(y, ConstantTimeTensor):
+            return ConstantTimeTensor(x.tensor + y.tensor)
+        elif isinstance(y, CallableTimeTensor):
+            f = lambda t: y.f(t) + x.tensor
+            f0 = y.f0 + x.tensor
+            return CallableTimeTensor(f, f0)
```
abocquet, line 244: You are missing the default case, it should at least crash instead of silently returning `None`

dynamiqs/time_tensor.py
```diff
@@ -0,0 +1,257 @@
...
+            f0 = x.f0 + y
+            return CallableTimeTensor(f, f0)
+        elif isinstance(y, ConstantTimeTensor):
+            f = lambda t: x.f(t) + y.tensor
+            f0 = x.f0 + y.tensor
+            return CallableTimeTensor(f, f0)
+        elif isinstance(y, CallableTimeTensor):
+            f = lambda t: x.f(t) + y.f(t)
+            f0 = x.f0 + y.f0
+            return CallableTimeTensor(f, f0)
```
abocquet, line 257: Same

abocquet (changes requested):

tests/mesolve/open_system.py
```diff
@@ -161,7 +161,7 @@ def __init__(
         self.loss_op = dq.sigmaz()
 
         # prepare quantum operators
-        self.H = lambda t: self.eps * torch.cos(self.omega * t) * dq.sigmax()
+        self.H = dq.tt(lambda t: self.eps * torch.cos(self.omega * t) * dq.sigmax())
```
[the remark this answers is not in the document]
pierreguilmin: You prefer `dq.time_tensor`?

dynamiqs/time_tensor.py
pierreguilmin: Arf, I'm not fan of this construct but ok for here (it's very risky whenever the argument may be a list, string, or any object that may evaluate to False somehow).

tests/mesolve/open_system.py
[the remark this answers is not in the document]
gautierronan: I like a shorter name for the public API. It may be a function that will be called often and in long expressions.

dynamiqs/time_tensor.py
pierreguilmin: I prefer to keep a correct initial value _in case of_ using `f0` for smthg else. Otherwise we should call it `shape`, but it's annoying bc you can't directly apply view operations on `torch.Size` object.

gautierronan (commented):
Don't have much more comments than @abocquet. LGTM.

abocquet (approved):

abocquet (changes requested):
There is still a silent fail

abocquet (dismissed):

gautierronan (dismissed):

abocquet (dismissed):

abocquet (dismissed):

PR #374, opened 2023-12-08
Add TimeTensor [2/4]: implement `PWCTimeTensor`

pierreguilmin:
Related to DYN-137.
Stacked on #372.
Replaces #233 except adapting the `Propagator` solver (will do later, see DYN-47).

The batching is held by the values tensor:
```python
import dynamiqs as dq
import torch
n = 8
a = dq.destroy(n)
# define pwc H  = drive x a
times = torch.linspace(0, 1.0, 11)
nseeds = 30
drive = dq.rand_complex((nseeds, 10))  # (30, 10)
Hpwc = dq.totime((times, drive, a))
Hpwc.shape  # (30, 8, 8) --> shape at a given time
```

We can sum tensors with non-aligned times:
```python
n = 8
a = dq.destroy(n)

# static
Hstatic = a.mH @ a
print(Hstatic.shape)  # (8, 8)

# total: H = adag @ a + drive * a + drive.conj() * adag
H = Hstatic + Hpwc + Hpwc.mH
print(H.shape)  # (30, 8, 8)

# test at time 0.0
Hse = Hstatic.expand(nseeds, n, n)
drive0 = drive[0][..., None, None]
assert torch.equal(H(0.0), Hse + drive0 * a + drive0.conj() * a.mH)

# test at time 0.12
drive1 = drive[1][..., None, None]
assert torch.equal(H(0.12), Hse + drive1 * a + drive1.conj() * a.mH)
```

### DIFF
--- a/dynamiqs/time_tensor.py
+++ b/dynamiqs/time_tensor.py
@@ -8,7 +8,14 @@ from torch import Tensor
 
 from ._utils import obj_type_str, type_str
 from .solvers.utils.utils import cache
-from .utils.tensor_types import ArrayLike, Number, get_cdtype, to_device, to_tensor
+from .utils.tensor_types import (
+    ArrayLike,
+    Number,
+    dtype_complex_to_real,
+    get_cdtype,
+    to_device,
+    to_tensor,
+)
 
 __all__ = ['totime']
 
@@ -22,6 +29,18 @@ def totime(
     dtype = dtype or get_cdtype(dtype)  # assume complex by default
     device = to_device(device)
 
+    # pwc time tensor
+    if isinstance(x, tuple) and len(x) == 3:
+        times, values, tensor = x
+        if dtype in (torch.complex64, torch.complex128):
+            rdtype = dtype_complex_to_real(dtype)
+        else:
+            rdtype = dtype
+        times = to_tensor(times, dtype=rdtype, device=device)
+        values = to_tensor(values, dtype=dtype, device=device)
+        tensor = to_tensor(tensor, dtype=dtype, device=device)
+        pwc = _PWCTimeTensor(times, values, tensor)
+        return PWCTimeTensor([pwc])
     # constant time tensor
     if isinstance(x, get_args(ArrayLike)):
         x = to_tensor(x, dtype=dtype, device=device)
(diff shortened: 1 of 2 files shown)


dynamiqs/time_tensor.py
```diff
@@ -21,6 +28,18 @@ def totime(
     dtype = dtype or get_cdtype(dtype)  # assume complex by default
     device = to_device(device)
 
+    # pwc time tensor
+    if isinstance(x, tuple) and len(x) == 3:
+        times, values, tensor = x
```
gautierronan, line 33: Can you also check that `times`, `values` and `tensor` have the proper type and shape, and otherwise raise an error to explain the required shapes?

abocquet (approved):
This is great, thank you 🚀

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,170 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
...
+class PWCTimeTensor(TimeTensor):
+    # Arbitrary sum of PWC tensors.
+    #
+    # A single PWC tensor is defined by a tuple of 3 tensors (times, values, tensor),
+    # where
+    # - times: (nv+1) are the time points between which the PWC tensor take constant
+    #          values, where nv is the number of time intervals
+    # - values: (nv, ...) are the constant complex values for each time interval, where
+    #           (...) is an arbitrary batching size
+    # - tensor: (n, n) is the tensor value
```
gautierronan, line 292: Two questions:

1. Shouldn't we allow formats `(n, m)`? I don't see any reason we should restrict to `(n, n)`, although this would be the main use.
2. Should we allow also the format `(nv, n, n)`? I can see applications where you would have a piece-wise constant tensor but the tensor itself changes from one piece to the next. We can still apply propagator solvers in this case.

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,170 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
...
+        # - ...: values batching (`self.values` carries all the shape transformation)
+        # - npwc: number of piecewise constant tensors
+        # - n: Hilbert space dimension
+
+        self.times = times  # (nv+1)
+
+        self.values = values  # (npwc, nv, ...)
+        self.npwc = values.shape[0]
+        self.nv = values.shape[1]
+        self.dots = values.shape[2:]  # (...)
```
gautierronan, line 319:
```suggestion
        self.nbatch = values.shape[2:]  # (...)
```
(maybe more explicit)

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,170 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
...
+    # v2 = [[1, 2, 3]]
+    # times: 0     5     10        20        30 = [0, 5, 10, 20, 30]
+    #        |-----|-----|---------|---------|
+    # v1n:   |  1  |  1  |    2    |    0    |  = [1, 1, 2, 0]
+    # v2n:   |  1  |  2  |    3    |    3    |  = [1, 2, 3, 3]
+    # t = [0, 5, 10, 20, 30]
+    # v = [[1, 1, 2, 0], [1, 2, 3, 3]]
+
+    # merge times
+    t = torch.cat((t1, t2)).sort()[0].unique(sorted=True)  # (nv+1)
```
gautierronan, line 419:
```suggestion
    t = torch.cat((t1, t2)).unique(sorted=True)  # (nv+1)
```
Isn't this already sorting?

gautierronan (changes requested):
Very nice work! A few comments/questions below regarding the format.

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,170 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
...
+
+
+class PWCTimeTensor(TimeTensor):
+    # Arbitrary sum of PWC tensors.
+    #
+    # A single PWC tensor is defined by a tuple of 3 tensors (times, values, tensor),
+    # where
+    # - times: (nv+1) are the time points between which the PWC tensor take constant
+    #          values, where nv is the number of time intervals
+    # - values: (nv, ...) are the constant complex values for each time interval, where
```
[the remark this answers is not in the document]
pierreguilmin: It's a bit more natural with `(nv, ...)` IMO, because we always use `values` after having selected one of its elements `values[idx, ...]`.

dynamiqs/time_tensor.py
pierreguilmin: Yep 👍

dynamiqs/time_tensor.py
pierreguilmin: 1. We _could_ actually allow arbitrary batching for the tensor, but I don't see why we _should_ as it's a tiny bit more complex.
2. Ongoing discussion on Slack.

dynamiqs/time_tensor.py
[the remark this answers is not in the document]
pierreguilmin: Actually I agree, let me fix it.

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,103 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
...
+    # Defined by a tuple of 2 tensors (times, values), where
+    # - times: (nv+1) are the time points between which the PWC tensor take constant
+    #          values, where nv is the number of time intervals
+    # - values: (..., nv) are the constant values for each time interval, where
+    #           (...) is an arbitrary batching size
+
+    def __init__(self, times: Tensor, values: Tensor):
+        self.times = times  # (nv+1)
+        self.values = values  # (..., nv)
+        self.nv = len(self.values)
```
gautierronan, line 293:
```suggestion
        self.nv = self.values.shape[-1]
```

gautierronan (approved):
Very nice, thanks for the changes !

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,103 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
...
+    @property
+    def device(self) -> torch.device:
+        return self.tensors.device
+
+    @property
+    def shape(self) -> torch.Size:
+        return torch.Size((*self.factors[0].shape, self.n, self.n))  # (..., n, n)
+
+    def __call__(self, t: float) -> Tensor:
+        static = self.static.expand(*self.shape)  # (..., n, n)
```
gautierronan, line 342: What is the point of this line? I think the sum should still work with broadcasting. Maybe it should be inside the first `if` statement.

dynamiqs/time_tensor.py
```diff
@@ -239,3 +278,103 @@ def __add__(self, other: Tensor | TimeTensor) -> TimeTensor:
             return CallableTimeTensor(f, f0)
         else:
             return NotImplemented
+
+
+class _PWCTensor:
```
gautierronan, line 283: Can we call this `_PWCFactor` ?

gautierronan: Fixed in #374.

pierreguilmin: I'd rather fix it here if possible to avoid mixing the PRs, is that ok for you? There are still a few things to fix in this PR to handle caching properly, and it will conflict with the renaming in your PR if we don't do it here.

gautierronan: Of course !

PR #381, opened 2023-12-19
Remove normalization in plot_wigner

gautierronan:
This is useful for instance when plotting the wigner of a trace 0 density matrix (e.g. a density matrix obtained from Liouvillian diagonalization). It will get renormalized like crazy.

Closes DYN-151.


abocquet (dismissed):
To answer your question, to me yes we can remove the warning. If the user chooses to change the default value they should know what they are doing so there is no need to guide them with a warning

pierreguilmin:
> To answer your question, to me yes we can remove the warning. If the user chooses to change the default value they should know what they are doing so there is no need to guide them with a warning

The default value is `normalize=True`, so we should either keep it if we choose to keep the normalisation. However IMO if we allow plotting the Wigner of non-physical states we can just remove altogether the `normalize` option, the warning and the normalisation code.

abocquet:
I'm in favor of keeping the normalization as the warning saved my life once or twice (too small timesteps, wrong GPU configuration, low truncature)

pierreguilmin:
> I'm in favor of keeping the normalization as the warning saved my life once or twice (too small timesteps, wrong GPU configuration, low truncature)

But then you see visually that it's incorrect no?

abocquet:
It's hard to see when  state has a norm between 0.9 and 1.1 at least

pierreguilmin:
Right. So we can bring back the warning then, but maybe not normalize the state ourselves. Maybe the name should be `check_normalize` or `check_state` then. Actually a bunch of other methods should check this (e.g. `plot_fock`) and we could have a common argument name to skip or not the checks.

pierreguilmin (approved):

PR #383, opened 2023-12-20
Use PWC Hamiltonian format in propagator solvers

pierreguilmin:
Closes DYN-47.

Replaces #233. I copy-paste a comment from there to motivate the PR (mostly performance related 🚀):
> Notably it allows to use the propagator solver for optimal control problems with PWC drives, which leads to a huge speedup (e.g. optimisation of Fock 2 preparation with cavity + transmon goes from ~3 minutes to ~20 seconds on my CPU). I think it’s important to have a good enough PWC Hamiltonian sesolve solver, because this is the first step of many optimal control proof-of-concept, it's a quick way to know wether something is doable or not.

The important part is that we need to evolve between each time in `self.tstop`, and also each time in `H.times` (otherwise we will _skip_ a PWC part). Thus we create a new `self.tstop_new` times array in the `Propagator` base class to evolve in-between the correct times.

<details>
  <summary>Handcrafted test to check that the caching and the stopping times are correct</summary>

```python
import dynamiqs as dq
import numpy as np
from matplotlib import pyplot as plt

dq.mplstyle()

n = 16
a = dq.destroy(n)

times = [1, 2]
values = [1]

H = dq.totime((times, values, a.mH @ a))
jump_ops = [dq.zero(n)]
psi0 = dq.coherent(n, 2.0)
tsave = np.linspace(0, 3.0, 11)  # [0.0, 0.3, 0.6, ..., 3.0]
exp_ops = [dq.position(n)]
solver = dq.solver.Propagator()
x = dq.sesolve(H, psi0, tsave, exp_ops=exp_ops, solver=solver).expects.real[0]
plt.plot(tsave, x, '--o')
for t in times:
    plt.axvline(t, color='gray', ls='--')
plt.xlim(tsave[0], tsave[-1]);
```

050e7a38-d294-4840-9c81-787d4a0c8db8

```text
Starting from t=0.00, evolving for delta_t=0.3
  └── compute propagator: id(H)=6298916400, delta_t=0.3
Starting from t=0.30, evolving for delta_t=0.3
Starting from t=0.60, evolving for delta_t=0.3
Starting from t=0.90, evolving for delta_t=0.1
  └── compute propagator: id(H)=6298916400, delta_t=0.1  # delta_t changed
Starting from t=1.00, evolving for delta_t=0.2
  └── compute propagator: id(H)=6298916560, delta_t=0.2  # H changed
Starting from t=1.20, evolving for delta_t=0.3
  └── compute propagator: id(H)=6298916560, delta_t=0.3  # delta_t changed
Starting from t=1.50, evolving for delta_t=0.3
Starting from t=1.80, evolving for delta_t=0.2
  └── compute propagator: id(H)=6298916560, delta_t=0.2  # delta_t changed
Starting from t=2.00, evolving for delta_t=0.1
  └── compute propagator: id(H)=6298916800, delta_t=0.1  # H changed
Starting from t=2.10, evolving for delta_t=0.3
  └── compute propagator: id(H)=6298916800, delta_t=0.3  # delta_t changed
Starting from t=2.40, evolving for delta_t=0.3
Starting from t=2.70, evolving for delta_t=0.3
```

</details>

### DIFF
--- a/dynamiqs/solvers/propagator.py
+++ b/dynamiqs/solvers/propagator.py
@@ -5,7 +5,8 @@ from abc import abstractmethod
 import numpy as np
 from torch import Tensor
 
-from ..time_tensor import ConstantTimeTensor
+from .._utils import obj_type_str
+from ..time_tensor import ConstantTimeTensor, PWCTimeTensor
 from .solver import AutogradSolver
 from .utils.utils import tqdm
 
@@ -28,20 +29,38 @@ class Propagator(AutogradSolver):
         super().__init__(*args, **kwargs)
 
         # check that Hamiltonian is time-independent
-        if not isinstance(self.H, ConstantTimeTensor):
+        if not isinstance(self.H, (ConstantTimeTensor, PWCTimeTensor)):
             raise TypeError(
-                'Solver `Propagator` requires a time-independent Hamiltonian.'
+                f'Solver {obj_type_str(self.options.solver)} requires a'
+                ' time-independent or piecewise-constant Hamiltonian (types'
+                ' `ArrayLike`, `ConstantTimeTensor` or `PWCTimeTensor`), but the'
+                f' passed Hamiltonian has type {obj_type_str(self.H)}.'
             )
-        self.H = self.H(self.t0)
+        if isinstance(self.H, PWCTimeTensor):
+            # We build a custom times array `tstop_new` gathering `self.tstop` and
+            # `self.H.times`. This ensures that each PWC part of the Hamiltonian is
+            # taken into account during the evolution.
+
+            # keep only `H.times` values in the interval [t0, tsave[-1])
+            mask = (self.t0 <= self.H.times) & (self.H.times < self.tsave[-1])
+            extra_tstop = self.H.times[mask]
+            extra_tstop = extra_tstop.numpy(force=True)
+
+            # build new times to stop
+            self.tstop_new = np.unique(np.concatenate((self.tstop, extra_tstop)))
+        else:
+            self.tstop_new = self.tstop
 
     def run_autograd(self):
         t1, y = self.t0, self.y0
-        for t2 in tqdm(self.tstop, disable=not self.options.verbose):
+        for t2 in tqdm(self.tstop_new, disable=not self.options.verbose):
             if t2 != self.t0:
                 # round time difference to avoid numerical errors when comparing floats
                 delta_t = round_truncate(t2 - t1)
                 y = self.forward(t1, delta_t, y)
-            self.save(y)
+            # if the time is in the original `self.tstop` times, save the result
+            if t2 in self.tstop:
+                self.save(y)
             t1 = t2
 
     @abstractmethod
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -146,3 +146,7 @@ def cache(func=None, *, maxsize: int = 1):
         return grad_cached_func(*args, grad_enabled=torch.is_grad_enabled(), **kwargs)
 
     return wrapper
+
+
+def merge_tensors(*x: Tensor) -> Tensor:
+    return torch.cat(x).unique(sorted=True)
--- a/dynamiqs/mesolve/propagator.py
+++ b/dynamiqs/mesolve/propagator.py
@@ -10,17 +10,20 @@ from .me_solver import MESolver
 class MEPropagator(MESolver, Propagator):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.lindbladian = slindbladian(self.H, self.L)  # (..., n^2, n^2)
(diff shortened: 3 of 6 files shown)


dynamiqs/solvers/propagator.py
```diff
@@ -28,20 +29,38 @@ def __init__(self, *args, **kwargs):
...
+            # `self.H.times`. This ensures that each PWC part of the Hamiltonian is
+            # taken into account during the evolution.
+
+            # keep only `H.times` values in the interval [t0, tsave[-1])
+            mask = (self.t0 <= self.H.times) & (self.H.times < self.tsave[-1])
+            extra_tstop = self.H.times[mask]
+            extra_tstop = extra_tstop.numpy(force=True)
+
+            # build new times to stop
+            self.tstop_new = np.unique(np.concatenate((self.tstop, extra_tstop)))
```
gautierronan, line 50: Maybe we should call this `self.tstop` as well no? The `_new` doesn't add much to the variable name in terms of meaning IMO.

dynamiqs/solvers/propagator.py
```diff
@@ -28,20 +29,38 @@ def __init__(self, *args, **kwargs):
...
         t1, y = self.t0, self.y0
-        for t2 in tqdm(self.tstop, disable=not self.options.verbose):
+        for t2 in tqdm(self.tstop_new, disable=not self.options.verbose):
             if t2 != self.t0:
                 # round time difference to avoid numerical errors when comparing floats
                 delta_t = round_truncate(t2 - t1)
                 y = self.forward(t1, delta_t, y)
-            self.save(y)
+            # if the time is in the original `self.tstop` times, save the result
+            if t2 in self.tstop:
```
gautierronan, line 62: And here I guess this should be `self.tsave` instead? i.e. we stop on `self.tstop` and we save on `self.tsave`. Makes more sense I think.

gautierronan (changes requested):
Small comments on the naming of variables. Otherwise looks good.

dynamiqs/solvers/propagator.py
pierreguilmin: We can't, it will override the base class `self.tstop` name that may be used internally.

dynamiqs/solvers/propagator.py
pierreguilmin: This goes a bit against the base class `tstop` logic. `tstop` = the base class asks to stop (and save or does whatever it needs to do), `tstop_new` = the propagator subclass asks to stop. I think it might be harmful to use `self.tsave` here, it breaks the base class abstraction.

dynamiqs/solvers/propagator.py
gautierronan: Actually I didn't notice this was only in the propagator file. I guess it's fine then. How about `self.tstop_pwc` as a name then? Or something else that carries more meaning.

dynamiqs/solvers/propagator.py
gautierronan: Actually why do you need to check `if t2 in self.tstop` ? This check should already be achieved within `self.save` with the `self.tsave_mask`. Otherwise maybe we should adapt it there?

dynamiqs/solvers/propagator.py
gautierronan: Still, I think it should be fine to override `self.tstop` no? At least if it's not, it's how it should be IMO. `self.tstop` should only carry information about when we require to stop. Here, we need to stop more often than before, but it shouldn't change the rest of the process.

pierreguilmin: You can look at it two ways:
- either `self.tstop` is a read-only property of the base `Solver` class which denotes when `self.save(y)` should be called, and subclasses can do whatever they want on top with how they want to manage time (e.g. fixed solvers stop every dt, adaptive solvers stop often, propagator stop every pwc times, etc...)
- either `self.tstop` is a mutable property of the base class which denotes when the specific solver stops, subclasses can modify it but they need to call `self.save(y)` at the times of the original `self.tstop`

I wrote the base class with the first case in mind, I think it's much easier not to tamper with the internal of the base class (especially with the double saving `tsave`, `tmeas`, and the `tstop_counter`). I agree that it works to override `self.tstop` here, but it's really a hack IMO (for example `tstop_counter` in the base class becomes completely unrelated to `tstop`). Maybe what's confusing is that we should rename the base method `save` to `stop` to make this more explicit (for example we could stop in the base solver to log GPU usage or whatever, for now we stop to save so we called it `save`).

dynamiqs/solvers/propagator.py
pierreguilmin: I'm not sure to understand. If you read the first bullet point of previous comment, this is how the base solver class was designed: subclasses must stop at every time in `tstop`, and call `save()`. I think it's the clearest way to manage saving indepently of what the solver does.

PR #386, opened 2023-12-21
Test `smesolve()`

pierreguilmin:
Closed DYN-110.

This is a very basic test, it leverages the fact that a coherent state is immune to measurement, so the trajectory is the same wether or not the state is monitored. Later we should write proper tests, to also ensure that the measurement statistics are correct, and for real stochastic trajectories.

### DIFF
--- a/tests/smesolve/monitored_system.py
+++ b/tests/smesolve/monitored_system.py
@@ -1,0 +1,156 @@
+from __future__ import annotations
+
+from math import cos, exp, pi, sin
+from typing import Any
+
+import torch
+from torch import Tensor
+
+import dynamiqs as dq
+from dynamiqs import TimeTensor
+from dynamiqs.gradient import Gradient
+from dynamiqs.solver import Solver
+from dynamiqs.solvers.result import Result
+from dynamiqs.utils.tensor_types import ArrayLike
+
+from ..mesolve.open_system import OpenSystem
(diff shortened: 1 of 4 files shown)


gautierronan (dismissed):
Looks good. Did you simply do a "visual" check that the correctness and gradient tests are passing well? I'm asking because the `rtol` and `atol` are quite large, but it's maybe expected since it's just first-order solvers.

gautierronan:
> > Looks good. Did you simply do a "visual" check that the correctness and gradient tests are passing well? I'm asking because the `rtol` and `atol` are quite large, but it's maybe expected since it's just first-order solvers.
> 
> I'm not sure I understand your question? I copy-pasted the `rtol` and `atol` from the corresponding solver in `mesolve` tests.

I just meant launching with a very small `rtol` and `atol`, looking at the analytical values vs simulated values, and checking they're approximately the same.

Since we're using quite large `rtol`/`atol`, sometimes the tests pass, but when you print analytical vs simulated, it's clear that it's not actually passing.

gautierronan (approved):

pierreguilmin:
> I just meant launching with a very small rtol and atol, looking at the analytical values vs simulated values, and checking they're approximately the same.

Ho I see, I just did the check, it looks good 👍🏼

PR #388, opened 2023-12-21
Change API of slindbladian

gautierronan:
Closes DYN-117.

Make the API of `dq.slindbladian` consistent with `dq.mesolve`, by accepting a list of jump operators.

Ideally, I think we should also accept ArrayLike and make the proper conversions, but it's not super important for now.

### DIFF
--- a/dynamiqs/utils/vectorization.py
+++ b/dynamiqs/utils/vectorization.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from math import sqrt
 
 import torch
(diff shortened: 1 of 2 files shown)


dynamiqs/utils/vectorization.py
```diff
@@ -174,16 +176,16 @@ def sdissipator(L: Tensor) -> Tensor:
     $$
 
     Args:
-        L _(..., n, n)_: Jump operator (an arbitrary operator).
+        L _(..., n, n)_: Jump operator.
```
pierreguilmin, line 179: Also fix this comment in `dissipator()`.

dynamiqs/utils/vectorization.py
```diff
@@ -209,9 +211,11 @@ def slindbladian(H: Tensor, L: Tensor) -> Tensor:
...
-        L _(..., N, n, n)_: Sequence of jump operators (arbitrary operators).
+        jump_ops _(list of tensors of shape (..., n, n), or tensor of shape (N, ..., n,
+            n))_: List of jump operators.
 
     Returns:
         _(..., n^2, n^2)_ Lindbladian superoperator.
     """
-    return -1j * (spre(H) - spost(H)) + sdissipator(L).sum(0)
+    Ls = torch.stack(jump_ops) if isinstance(jump_ops, list) else jump_ops
+    return -1j * (spre(H) - spost(H)) + sdissipator(Ls).sum(0)
```
pierreguilmin, line 221: We should also fix the `lindbladian()` function.

pierreguilmin (changes requested):
Nice!

PR #391, opened 2023-12-22
Add time derivative calculation functionality

### LINKED ISSUE #389: Calculating derivative with respect to time
Let's say that I want to calculate the expectation value of some observable `C`, thus defining a function `f(t)=<0|exp(itH)*C*exp(-itH)|0>`. In order to calculate its derivative `dfdt` with respect to time (tensor `tsave`) I would use `torch.autograd.grad`. However, currently, as far as I have investigated, this is not possible since in `dynamiqs/solvers/solver.py` line 39 `tsave` is converted to `numpy` array, thus removing it from the computational graph. 

Is there another way to calculate derivative with respect to time or this functionality has yet to be implemented?

Minimal example:
```
import dynamiqs as dq
import torch
import numpy as np
from scipy import interpolate

torch.set_default_device("cuda")

# parameters
omega = torch.tensor([4.0], requires_grad=True)

# define Hamiltonian
ham = (omega/2) * (dq.tensprod(dq.sigmax(), dq.eye(2)) + dq.tensprod(dq.eye(2), dq.sigmax()))

# initial state
psi0 = torch.tensor([1,0,0,0])

# define tsave
tsave = torch.linspace(1.0, 3.0, 201, requires_grad=True)

# define observable
total_magnetization = dq.tensprod(dq.sigmaz(), dq.eye(2)) + dq.tensprod(dq.eye(2), dq.sigmaz())

# run simulation
result = dq.sesolve(
    ham, psi0, tsave, 
    gradient=dq.gradient.Autograd(),
    options=dict(verbose=False),
)

# calculate expectation value
exp_val = dq.expect(total_magnetization, result.states).real


print("expectation value:")
print(exp_val)


# calculate derivative with respect to time
dexpval_dx = torch.autograd.grad(
    exp_val, tsave, torch.ones_like(exp_val), create_graph=True
)[0]



x = tsave.cpu().detach().numpy()
y = exp_val.cpu().detach().numpy()
dfdx_exact = interpolate.UnivariateSpline(x, y, k=5, s=0).derivative()(x)

plt.plot(x, exp_val.cpu().detach().numpy(), label="f(x)")
plt.scatter(x, dexpval_dx.cpu().detach().numpy(), label="dfdx AD", lw=3.0, s=2)
plt.plot(x, dfdx_exact, label="dfdx interpolated")
plt.legend()
plt.show()
```

The code fails at line where `torch.autograd.grad` is used.

vytautas-a:
Implementation of time derivative functionality: derivative can be calculated with respect to `tsave`.

Closes #389

### DIFF
--- a/docs/getting_started/sharp-bits.md
+++ b/docs/getting_started/sharp-bits.md
@@ -7,6 +7,8 @@ In dynamiqs we use PyTorch tensors to represent quantum states and operators. A 
 ```python
 import dynamiqs as dq
 import torch
+
+torch.set_default_device("cpu")
 ```
 
 ## Main differences with QuTiP
--- a/dynamiqs/plots/plots_hinton.py
+++ b/dynamiqs/plots/plots_hinton.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 import matplotlib as mpl
-import matplotlib.patches as patches
 import numpy as np
+from matplotlib import patches
 from matplotlib.axes import Axes
 from matplotlib.collections import PatchCollection
 from matplotlib.colors import Normalize
--- a/dynamiqs/solvers/ode/adaptive_solver.py
+++ b/dynamiqs/solvers/ode/adaptive_solver.py
@@ -70,11 +70,18 @@ class AdaptiveSolver(ODESolver):
                 dt = t1 - t
 
             # compute the next step
+            # print("t:", t)
+            # print("y:", y)
+            # print("ft:", ft)
+            # print("dt:", dt)
+            # print(self.odefun)
+
             ft_new, y_new, y_err = self.step(t, y, ft, dt, self.odefun)
+
             error = self.get_error(y_err, y, y_new)
             if error <= 1:
                 t, y, ft = t + dt, y_new, ft_new
-                self.pbar.update(dt)
+                self.pbar.update(float(dt))
 
             # check max steps are not reached
             self.increment_step_counter(t)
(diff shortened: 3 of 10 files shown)


pierreguilmin:
Thanks for your PR @vytautas-a! We're in the process of migrating the repository to JAX, we'll come back to your work when we're done (early January). Have a great end of the year 😊

gautierronan:
Hi @vytautas-a, sorry for the long delay in response. As stated above, we were migrating to JAX, and the migration is now pretty much over. I hope you'll get to enjoy this new version of dynamiqs.

In turns out that differentiating with respect to time is now naturally supported. You can find below a short example:
```python
import dynamiqs as dq
import jax.numpy as jnp
import jax

def position(omega, T):
    H = omega * dq.number(16)
    psi0 = dq.coherent(16, 1.0)
    tsave = jnp.linspace(0.0, T, 11)
    exp_ops = [dq.destroy(16)]
    result = dq.sesolve(H, psi0, tsave, exp_ops=exp_ops)
    return result.expects[0, -1].real

grad_position = jax.grad(position, argnums=(0, 1))
print(grad_position(1.0, 1.0))
# (Array(-0.84147686, dtype=float32, weak_type=True),
#  Array(-0.8414724, dtype=float32, weak_type=True))
```
There are some warnings that pop up when differentiating w.r.t time. I'm not sure where they come from, but will try to fix those at some point.

In any case, thanks a lot for your contribution. I'm sorry it didn't end up making it to main. We'd be very happy to see other contributions from you in the future though !

PR #395, opened 2023-12-28
Switch `dynamiqs.utils` from PyTorch to JAX [2/?]

pierreguilmin:
Screenshot 2023-12-28 at 18 58 35

On top of #394.

A lot of things are packed in this PR, sorry for the gigantic change. It transitions the following files from PyTorch to JAX:
```
dynamiqs
└── utils
    ├── operators.py
    ├── optimal_control.py
    ├── states.py
    ├── tensor_types.py
    ├── utils.py
    ├── vectorization.py
    └── wigners.py (partially)
tests
└── utils
    └── test_utils.py
```

The doctest is running on all these files.

Most of the code is similar, some things are more natural to do which is nice! 😊 A few random remarks:
- https://jax.readthedocs.io/en/latest/jax.typing.html advises to type all public APIs argument with `ArrayLike`, so I followed their advice. It's quite amazing, now all our functions work on any QuTiP quantum objects for example (or even lists of QuTiP quantum objects).
- I removed the `device` argument as it is managed differently in JAX.
- I added "array_like" or "array" to all type annotations specifying tensors shape for clarity.
- I removed the functions `to_tensor`, `to_numpy` and `from_qutip` which are replaced by `jnp.asarray`.
- I had to modify `_utils` as a side effet, but it has not been switched to JAX yet.
- JAX array printing is much cleaner, it does not displays extra zeros. 😍
- I did not fully switch `dynamiqs/utils.wigners.py` to JAX, just the `"clenshaw"` method.
- ChatGPT is terrible with JAX, most of the answers are useless. 😭
- JAX does not have a `x.mH`, so I use `dag(x)` everywhere.
- The single main changes are (1) the batched tensor product is now a one-liner using `jnp.vectorize`, (2) `sqrtm()` is natively implemented in JAX.
- JAX mimics NumPy broadcasting which is amazing, for example we can now use the following broadcasting magic:
    ```python
    import dynamiqs as dq
    psi1 = dq.fock(2, 0)  # (2, 1)
    psis = [dq.fock(8, i) for i in range(5)]  # (5, 8, 1)
    dq.tensprod(psi1, psis)  # (5, 16, 1)
    ```

### DIFF
--- a/dynamiqs/utils/optimal_control.py
+++ b/dynamiqs/utils/optimal_control.py
@@ -61,7 +61,7 @@ def rand_complex(
     r"""Returns an array filled with random complex numbers uniformly distributed in
     the complex plane.
 
-    Each element of the returned array is sampled uniformly in the disc of radius
+    Each element of the returned array is sampled uniformly in the disk of radius
     $\text{rmax}$.
 
     Notes-: Uniform sampling in the complex plane
--- a/dynamiqs/utils/optimal_control.py
+++ b/dynamiqs/utils/optimal_control.py
@@ -1,103 +1,72 @@
 from __future__ import annotations
 
-import torch
-from torch import Tensor
+import jax
+import jax.numpy as jnp
+from jax import Array
+from jaxtyping import ArrayLike
 
-from .._utils import check_time_tensor
 from ..utils.operators import displace
 from ..utils.states import fock
 from ..utils.utils import tensprod, tobra
-from .tensor_types import dtype_complex_to_real, get_cdtype, get_rdtype, to_device
+from .tensor_types import dtype_complex_to_real, get_cdtype, get_rdtype
 
-__all__ = ['rand_real', 'rand_complex', 'pwc_pulse', 'snap_gate', 'cd_gate']
+__all__ = ['rand_real', 'rand_complex', 'snap_gate', 'cd_gate']
 
 
 def rand_real(
-    size: int | tuple[int, ...],
+    shape: int | tuple[int, ...],
     *,
+    min: float = 0.0,
     max: float = 1.0,
-    requires_grad: bool = False,
-    seed: int | None = None,
-    dtype: torch.float32 | torch.float64 | None = None,
-    device: str | torch.device | None = None,
-) -> Tensor:
-    r"""Returns a tensor filled with uniformly distributed random real numbers.
+    seed: int = 0,
+    dtype: jnp.float32 | jnp.float64 | None = None,
+) -> Array:
+    r"""Returns an array filled with uniformly distributed random real numbers.
 
-    Each element of the returned tensor has a random value in the interval
-    $[-\texttt{max}, \texttt{max})$. Formally, each element is defined by
-    $$
-        x = \texttt{max} \cdot (2 \cdot \texttt{rand(0,1)} - 1),
-    $$
-    where $\texttt{rand(0,1)}$ is a random number uniformly distributed between 0 and 1.
+    Each element of the returned array is sampled uniformly in
+    $[\text{min}, \text{max})$.
 
     Args:
-        size _(int or tuple of ints)_: Size of the returned tensor.
-        max: Maximum absolute value of the random real numbers.
-        requires_grad: Whether gradients need to be computed with respect to the
-            returned tensor.
+        shape _(int or tuple of ints)_: Shape of the returned array.
+        min: Minimum (inclusive) value.
+        max: Maximum (exclusive) value.
         seed: Seed for the random number generator.
-        dtype: Real data type of the returned tensor.
-        device: Device of the returned tensor.
+        dtype: Complex data type of the returned array.
 
     Returns:
-        _(*size)_ Tensor filled with random real numbers.
+        _(array of shape (*shape))_ Array filled with random real numbers.
 
     Examples:
-        >>> x = dq.rand_real((2, 5), max=5.0, seed=42)
-        >>> x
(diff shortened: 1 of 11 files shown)


abocquet (dismissed):

dynamiqs/utils/optimal_control.py
```diff
@@ -1,103 +1,72 @@
...
-    element is defined by
-    $$
-        x = re^{i\theta}\ \text{with}\
-        \begin{cases}
-            r = \texttt{rmax} \cdot \sqrt{\texttt{rand(0,1)}} \\\\
-            \theta = 2\pi \cdot \texttt{rand(0,1)}
-        \end{cases}
-    $$
-    where $\texttt{rand(0,1)}$ is a random number uniformly distributed between 0 and 1.
+    Each element of the returned array is sampled uniformly in the disc of radius
```
gautierronan, line 64:
```suggestion
    Each element of the returned array is sampled uniformly in the disk of radius
```
(disc also exists, but you use disk below).

dynamiqs/utils/optimal_control.py
```diff
@@ -122,107 +91,34 @@ def rand_complex(
...
-    theta = 2 * torch.pi * rand()
-    x = r * torch.exp(1j * theta)
-
-    # start tracking operations on the tensor if requires_grad is True
-    x.requires_grad_(requires_grad)
-
-    return x
-
-
-def pwc_pulse(times: Tensor, values: Tensor) -> callable[[float], Tensor]:
```
gautierronan, line 170: Is this function not needed anymore?

dynamiqs/utils/tensor_types.py
```diff
@@ -1,202 +1,81 @@
...
 
-# type for objects convertible to a torch.Tensor using `to_tensor`
-ArrayLike = Union[tuple, list, np.ndarray, Tensor, Qobj]
 
-# data type conversion dictionaries
-DTYPE_TO_REAL = {torch.complex64: torch.float32, torch.complex128: torch.float64}
-DTYPE_TO_COMPLEX = {torch.float32: torch.complex64, torch.float64: torch.complex128}
+def _get_default_dtype() -> jnp.float32 | jnp.float64:
+    default_dtype = jnp.array(0.0).dtype
+    return jnp.float64 if default_dtype == jnp.float64 else jnp.float32
```
gautierronan, line 22:
```suggestion
    return jnp.array(0.0).dtype
```

dynamiqs/utils/utils.py
```diff
@@ -29,80 +31,88 @@
...
-        tensor([[1.+0.j],
-                [0.+0.j]])
+        Array([[1.+0.j],
+               [0.+0.j]], dtype=complex64)
         >>> dq.dag(dq.fock(2, 0))
-        tensor([[1.-0.j, 0.-0.j]])
+        Array([[1.-0.j, 0.-0.j]], dtype=complex64)
     """
-    return x.mH
+    x = jnp.asarray(x)
```
gautierronan, line 55: I don't think we should convert every time, I think we should have `dag(x: Array) -> Array` or `dag(x:ArrayLike) -> ArrayLike` instead. The function could be called many times in a given computation.

gautierronan: Same comment for all other utility functions with the same typing.

dynamiqs/utils/utils.py
```diff
@@ -156,87 +168,67 @@ def ptrace(x: Tensor, keep: int | tuple[int, ...], dims: tuple[int, ...]) -> Ten
...
-
-    # x: (..., x1, x2)
-    # y: (..., y1, y2)
-
-    batch_dims = x.shape[:-2]
-    x1, x2 = x.shape[-2:]
-    y1, y2 = y.shape[-2:]
-    kron_dims = torch.Size((x1 * y1, x2 * y2))
+    return reduce(_bkron, args)
+    # JAXDO: jax.lax.reduce?
```
gautierronan, line 223: Yes should be jax.lax.reduce I think, for tracing.

gautierronan (changes requested):

dynamiqs/utils/optimal_control.py
pierreguilmin: It's replaced by `dq.totime()` for pwc arrays.

dynamiqs/utils/tensor_types.py
pierreguilmin: It does not return `jnp.float32` or `jnp.float64` but a weird object 😅 So then when you do `dtype is jnp.float32` it returns `False`. Maybe we should extend the typing of our dtype to match any possible dtype (e.g. in JAX you can do `dtype='complex32'`). Wdyt?

dynamiqs/utils/utils.py
pierreguilmin: It's what JAX recommends to do (typing args with `ArrayLike` and returning `Array`) > https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices. When jitting all these checks vanish.

dynamiqs/utils/tensor_types.py
gautierronan: Quite strange, but I guess we should do this:
```python
dtype_array = jnp.array(0.0, dtype=jnp.float32).dtype
dtype_numpy = np.dtype(np.float32)
print(dtype_array is dtype_numpy)
# True
```

dynamiqs/utils/utils.py
gautierronan: Ah ok, all good then !

PR #397, opened 2023-12-28
Switch `dynamiqs.plots` from PyTorch to JAX [4/?]

pierreguilmin:
On top of #396.

Transition the following files from PyTorch to JAX:

```
dynamiqs
└── plots
    └── plots_fock.py
    └── plots_hinton.py
    └── plots_misc.py
    └── plots_wigner.py
```

The doctest is running except on examples using `sesolve`/`mesolve`.

### DIFF
--- a/dynamiqs/plots/plots_hinton.py
+++ b/dynamiqs/plots/plots_hinton.py
@@ -1,29 +1,48 @@
 from __future__ import annotations
 
+from itertools import product
+
+import jax.numpy as jnp
 import matplotlib as mpl
 import matplotlib.patches as patches
 import numpy as np
+from jax import Array
+from jax.typing import ArrayLike
 from matplotlib.axes import Axes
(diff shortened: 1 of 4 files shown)


dynamiqs/plots/plots_hinton.py
```diff
@@ -170,49 +191,49 @@ def plot_hinton(
 
         ![plot_hinton_large](/figs-code/plot_hinton_large.png){.fig}
     """  # noqa: E501
+    x = jnp.asarray(x)
```
abocquet, line 194: From what I understood, `jnp` can be thought of as "`np` with tracing" so we probably don't need it in plot functions

abocquet (changes requested):

gautierronan (commented):
Agree with @abocquet. Otherwise looks good.

PR #398, opened 2024-01-04
Move to diffrax 1/?

abocquet:
Manages to make test pass for euler without gradient

To support (among others):
- autodiff
- batching
- time dep
- support no grad


dynamiqs/sesolve/sesolve.py
```diff
@@ -0,0 +1,86 @@
...
+        # TODO : use vmap ?
+        res["expects"] = tuple([split_complex(bexpect(op, psi)) for op in exp_ops])
+        return res
+
+    solution = diffeqsolve(
+        diffrax.ODETerm(f),
+        solver_class(),
+        t0=tsave[0],
+        t1=tsave[-1],
+        dt0=tsave[1],  # TODO: there is probably a smarter choice
```
gautierronan, line 65: Yes, either we let diffrax choose it, or we could try something related to the norm of the Hamiltonian/jump operators at time t=0? e.g. the inverse of the largest eigenvalue difference of the Hamiltonian?

abocquet: I beleive it's a mandatory argument

pierreguilmin: I think it works to pass `dt=None` if a stepsize controller is defined, otherwise we should pass `options.dt` for fixed step solvers.

abocquet: It doesn't work : 
> ValueError: Constant step size solvers cannot select step size automatically; please pass a value for `dt0`.

PR #406, opened 2024-01-06
Fix time array implementation

pierreguilmin:
On top of #405.

I also made a base class for `PWCTimeArray` and `ModulatedTimeArray` as they had the exact same implementation (suggested by @gautierronan a while ago).

```python
import dynamiqs as dq
import jax.numpy as jnp

n = 16
a = dq.destroy(n)

H = dq.totime(dq.dag(a) @ a)
H = dq.totime(lambda t: dq.dag(a) @ a)
H = dq.totime(([1, 2, 3], [1, 3], dq.dag(a) @ a))
H = dq.totime((lambda t: jnp.array([3.0]), dq.dag(a) @ a))
```

### DIFF
--- a/dynamiqs/time_array.py
+++ b/dynamiqs/time_array.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
-from abc import abstractmethod, abstractproperty
-from typing import Tuple, get_args
+from abc import abstractmethod
+from typing import get_args
 
+import numpy as np
 from jax import Array
 from jax import numpy as jnp
 
 from ._utils import check_time_array, obj_type_str, type_str
-from .utils.array_types import ArrayLike, Number, dtype_complex_to_real, get_cdtype
+from .types import Scalar
+from .utils.array_types import ArrayLike, dtype_complex_to_real, get_cdtype
 
 __all__ = ['totime']
 
@@ -109,7 +111,7 @@ def _factory_pwc(
         )
 
     factors = [_PWCFactor(times, values)]
-    arrays = array.unsqueeze(0)  # (1, n, n)
+    arrays = array[None, ...]  # (1, n, n)
     return PWCTimeArray(factors, arrays)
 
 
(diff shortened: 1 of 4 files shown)


dynamiqs/time_array.py
```diff
@@ -483,45 +426,59 @@ def __init__(
...
-    def shape(self) -> Tuple[int, ...]:
-        return jnp.Size((*self.factors[0].shape, self.n, self.n))  # (..., n, n)
+    def shape(self) -> tuple[int, ...]:
+        return tuple(self.factors[0].shape, self.n, self.n)  # (..., n, n)
 
     def __call__(self, t: float) -> Array:
-        values = jnp.stack([x(t) for x in self.factors], dim=-1)  # (..., nf)
+        values = jnp.stack([x(t) for x in self.factors], axis=-1)  # (..., nf)
         values = values.reshape(*values.shape, 1, 1)  # (..., nf, n, n)
         return (values * self.arrays).sum(-3) + self.static  # (..., n, n)
```
gautierronan, line 439: Since now we have a jit, maybe we could do:
```python
if self.static is None:
  return (values * self.arrays).sum(-3)
else:
  return (values * self.arrays).sum(-3) + self.static
```
to avoid an additional sum with a zero-valued tensor. Unless the typing of `self.static` is not handled by the jit.

gautierronan: Actually may not be so smart to add if statements for the jit.

dynamiqs/time_array.py
```diff
@@ -288,158 +284,110 @@ def __init__(self, f: callable[[float], Array], f0: Array):
...
         return _PWCFactor(self.times, self.values.conj())
 
-    def __call__(self, t: float) -> Array:
+    def __call__(self, t: Scalar) -> Array:
         if t < self.times[0] or t >= self.times[-1]:
             return jnp.zeros_like(self.values[..., 0])  # (...)
         else:
             # find the index $k$ such that $t \in [t_k, t_{k+1})$
             idx = jnp.searchsorted(self.times, t, side='right') - 1
             return self.values[..., idx]  # (...)
```
gautierronan, line 384: I think this needs to be modified with some `lax` function to account for the if statement with the jit.

gautierronan (changes requested):

dynamiqs/time_array.py
pierreguilmin: I'm guessing the jit optimisation ignores addition with a all-0 array as part of the optimizations they do.

dynamiqs/time_array.py
pierreguilmin: Correct, let's fix JIT-stuff in another PR if you're ok with that.

abocquet (approved):

pierreguilmin:
@gautierronan I added todo comments. @abocquet once merged you'll have to reimplement `repeat` (both in `time_array.py` and tests).

PR #408, opened 2024-01-07
Fix mesolve correctness tests

gautierronan:
Small refactorings, and fixes the correctness for mesolve (the sesolve one still passes too).

@abocquet I removed the `self.constant_H` logic: with a `ConstantTimeArray`, the result is already "cached" by default since the same array is returned when calling `TimeArray.__call__` at any time `t`.

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -7,7 +7,14 @@ import jax.numpy as jnp
 import numpy as np
 from jaxtyping import ArrayLike
 
-from .._utils import _get_adjoint_class, _get_solver_class, save_fn
+from .._utils import (
+    SolverArgs,
+    _get_adjoint_class,
+    _get_solver_class,
+    merge_complex,
+    save_fn,
+    split_complex,
+)
 from ..gradient import Autograd, Gradient
 from ..options import Options
 from ..result import Result
@@ -47,6 +54,7 @@ def mesolve(
     Ls = [totime(L) for L in jump_ops]
     Ls = format_L(Ls)
     term = LindbladTerm(H=H, Ls=Ls)
+    exp_ops = jnp.asarray(exp_ops)
 
     solution = dx.diffeqsolve(
         term,
@@ -54,8 +62,8 @@ def mesolve(
         t0=tsave[0],
         t1=tsave[-1],
         dt0=dt,
-        y0=todm(rho0),
-        args=(options, exp_ops),
+        y0=split_complex(todm(rho0)),
+        args=SolverArgs(save_states=options.save_states, exp_ops=exp_ops),
         saveat=dx.SaveAt(ts=tsave, fn=save_fn),
         stepsize_controller=stepsize_controller,
         adjoint=adjoint_class(),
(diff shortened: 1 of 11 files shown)


gautierronan:
A small benchmark on the recurrent `jnp.stack` of jump operators:
```python
import jax
import jax.numpy as jnp
import timeit

def stack_and_sum(x: list[jax.Array]):
    return jnp.sum(jnp.stack(x), axis=0)

def sum_stacked(x: jax.Array):
    return jnp.sum(x, axis=0)

jitted_stack_and_sum = jax.jit(stack_and_sum)
jitted_sum_stacked = jax.jit(sum_stacked)

nL = 3
for N in [10, 100, 1000]:
    print(f'=== N = {N} ===')
    # prepare arrays
    Ls = [jax.random.normal(jax.random.PRNGKey(0), (N, N)) for _ in range(nL)]
    Ls_stack = jnp.stack(Ls)

    # timeit
    %timeit stack_and_sum(Ls)
    %timeit sum_stacked(Ls_stack)
    %timeit jitted_stack_and_sum(Ls)
    %timeit jitted_sum_stacked(Ls_stack)
```

```
=== N = 10 ===
148 µs ± 1.18 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
7.41 µs ± 93.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
6.42 µs ± 348 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
5.39 µs ± 109 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
=== N = 100 ===
165 µs ± 1.91 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
11.6 µs ± 47.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
9.15 µs ± 165 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
9.06 µs ± 83.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
=== N = 1000 ===
2.77 ms ± 29.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
965 µs ± 111 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
359 µs ± 76.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
1.23 ms ± 253 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
(not sure what happened on the last one, but overall stacking is fine if the code is jitted)

dynamiqs/mesolve/mesolve.py
```diff
@@ -46,29 +51,31 @@ def mesolve(
 
     # === solve differential equation with diffrax
     H = totime(H)
-    Ls = format_L(jump_ops)
+    Ls = [totime(L) for L in jump_ops]
```
abocquet, line 54: I don't understand why we move `totime` here

abocquet (approved):

dynamiqs/mesolve/mesolve.py
gautierronan: No strong reason, I just thought it was easier to read (this way, `format_L` just takes care of the batching dimension). I can put it back if you want !

pierreguilmin (approved):

dynamiqs/mesolve/mesolve.py
abocquet: Ok I was just curious, we can leave it here

pierreguilmin:
If we choose to stack each L at every step, we probably don't need to format them to have the same dimension, broadcasting will do the job.

gautierronan:
> If we choose to stack the jump operators at every step, we probably don't need to format them to have the same dimension, broadcasting will do the job.

True. I'm not 100% sure we should stack at every step, but for now it seemed like a good simple solution. Let's think about it some more, and if indeed we end up stacking we can remove the dimension formatting.

PR #410, opened 2024-01-09
abocquet/test jit sesolve

abocquet:
Not very critical in terms of jitting (expect for `_PWCFactor`) actually but it does fix a couple of bugs


gautierronan (changes requested):

dynamiqs/time_array.py
```diff
@@ -376,12 +379,16 @@ def conj(self) -> _Factor:
...
+        idx = jnp.searchsorted(self.times, t, side='right') - 1
+        return jax.lax.select(
+            t < self.times[0],
+            jnp.zeros_like(self.values[..., 0]),
+            jax.lax.select(
+                t >= self.times[-1],
+                jnp.zeros_like(self.values[..., 0]),
+                self.values[..., idx],
+            ),
+        )
```
pierreguilmin, line 391: I think https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html#jax.lax.cond is better suited here.
> In general [select()](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.select.html#jax.lax.select) leads to evaluation of both branches, although the compiler may elide computations if possible. For a similar function that usually evaluates only a single branch, see [cond()](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html#jax.lax.cond).

pierreguilmin: Also there's no need to nest select, and `jnp.searchsorted` should only be called in the second branch (as in the original Python code).

tests/solver_tester.py
```diff
@@ -136,6 +138,42 @@ def _test_correctness(
...
+    def _test_jit(
+        self,
+        system: System,
+        solver: Solver,
+        *,
+        options: dict[str, Any] | None = None,
+        ntsave: int = 11,
+        ysave_atol: float = 1e-3,
+        esave_rtol: float = 1e-3,
+        esave_atol: float = 1e-5,
```
pierreguilmin, line 150: You can remove this.

pierreguilmin (dismissed):
A few comments but I don't want to block if you want to move fwd.

dynamiqs/time_array.py
gautierronan: Adressed in https://github.com/dynamiqs/dynamiqs/pull/411

pierreguilmin (approved):

gautierronan (approved):

PR #413, opened 2024-01-11
Add progress bar

abocquet:
Adds a progress bar to mesolve and sesolve. Performances are not optimal yet since an event is triggered at each timestep

### DIFF
--- a/dynamiqs/mesolve/lindblad_term.py
+++ b/dynamiqs/mesolve/lindblad_term.py
@@ -15,7 +15,7 @@ class LindbladTerm(ProgressBarTerm):
     vector_field: Callable[[Scalar, PyTree, PyTree], PyTree]
 
     def __init__(
-        self, H: TimeArray, Ls: TimeArray, update_progressbar: Optional[Callable]
+        self, H: TimeArray, Ls: TimeArray, update_progressbar: callable | None
     ):
         super().__init__(update_progressbar)
         self.H = H
--- a/dynamiqs/mesolve/lindblad_term.py
+++ b/dynamiqs/mesolve/lindblad_term.py
@@ -1,24 +1,28 @@
 from __future__ import annotations
-from typing import Callable
+from typing import Callable, Optional
 
-import diffrax as dx
 import jax.numpy as jnp
 from jaxtyping import Array, PyTree, Scalar
 
+from progress_bar import ProgressBarTerm
 from ..time_array import TimeArray
 from ..utils.utils import dag
 
 
-class LindbladTerm(dx.ODETerm):
+class LindbladTerm(ProgressBarTerm):
     H: TimeArray  # (n, n)
     Ls: list[TimeArray]  # (nL, n, n)
     vector_field: Callable[[Scalar, PyTree, PyTree], PyTree]
 
-    def __init__(self, H: TimeArray, Ls: TimeArray):
+    def __init__(
+        self, H: TimeArray, Ls: TimeArray, update_progressbar: Optional[Callable]
+    ):
+        super().__init__(update_progressbar)
         self.H = H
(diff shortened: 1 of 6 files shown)


dynamiqs/mesolve/lindblad_term.py
```diff
@@ -1,24 +1,28 @@
...
 
-class LindbladTerm(dx.ODETerm):
+class LindbladTerm(ProgressBarTerm):
     H: TimeArray  # (n, n)
     Ls: list[TimeArray]  # (nL, n, n)
     vector_field: Callable[[Scalar, PyTree, PyTree], PyTree]
 
-    def __init__(self, H: TimeArray, Ls: TimeArray):
+    def __init__(
+        self, H: TimeArray, Ls: TimeArray, update_progressbar: Optional[Callable]
```
pierreguilmin, line 18: New-gen typing
```suggestion
        self, H: TimeArray, Ls: TimeArray, update_progressbar: callable | None
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -42,11 +43,14 @@ def mesolve(
     # === stepsize controller
     stepsize_controller, dt = _stepsize_controller(solver)
 
+    # === progress bar
+    bar, update_progressbar = make_progressbar(options.progress_bar, tsave[-1])
```
pierreguilmin, line 47: We maybe need to specify `tsave[0]` too bc we solve from `t0=tsave[0]`? Ideally `t0` should be an option in `options`.

dynamiqs/options.py
```diff
@@ -44,6 +44,7 @@ def __init__(
         verbose: bool = True,
         dtype: jnp.complex64 | jnp.complex128 | None = None,
         cartesian_batching: bool = True,
+        progress_bar: bool = True,
```
pierreguilmin, line 47: This is redundant with `verbose`. `progress_bar` is more explicit but longer, pick what you prefer!

dynamiqs/progress_bar.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+from typing import Callable, Union
```
pierreguilmin, line 3: New-gen typing, we probably don't need this

Callable -> callable
Union[a, b] -> a | b

pierreguilmin: This is why we `from __future__ import annotations`

dynamiqs/progress_bar.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __init__(self, update_progressbar: Union[Callable, None]):
+        self.update_progressbar = update_progressbar
+
+    def vector_field(self, t: Scalar, _psi: PyTree, _args: PyTree):
+        if self.update_progressbar is not None:
+            host_callback.id_tap(self.update_progressbar, t)
+
+
+def make_progressbar(progress_bar: bool, tfinal: float):
```
pierreguilmin, line 22: Can yout type the return?

pierreguilmin (dismissed):
Nice!  Ideally we push this upstream to diffrax directly.

dynamiqs/progress_bar.py
abocquet: TIL

dynamiqs/progress_bar.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+
+import diffrax as dx
+from jaxtyping import PyTree, Scalar
+from jax.experimental import host_callback
+from tqdm import tqdm
```
gautierronan, line 7:
```suggestion
from tqdm.auto import tqdm
```
(to handle both notebooks and terminals)

PR #415, opened 2024-01-15
JIT `wigner()`

abocquet:
Clenshaw only. Not the cleanest but I speeds up plot mosaic a bit. On a CPU:
image

For later PRs:
- [ ] FFT method
- [ ] implement a batching parameter to have a tradeoff between memory efficiency and speed


dynamiqs/utils/wigners.py
```diff
@@ -45,17 +45,29 @@ def wigner(
...
     xvec = jnp.linspace(-xmax, xmax, npixels, dtype=rdtype)
     yvec = jnp.linspace(-ymax, ymax, npixels, dtype=rdtype)
 
     if method == 'clenshaw':
         state = todm(state)
-        w = _wigner_clenshaw(state, xvec, yvec, g)
+        expanded = False
+        if state.ndim == 2:
+            state = state[None, ...]
+            expanded = True
```
pierreguilmin, line 57: Do we really need to do this every time we want to batch over a potentially non-existing dimension?  (especially having to remember `expanded` is a bit annoying) Is there a shorter way/trick/one liner/utility function/context manager we could use?

dynamiqs/utils/wigners.py
```diff
@@ -45,17 +45,29 @@ def wigner(
     """
     state = jnp.asarray(state)
```
pierreguilmin, line 46: Ideally we can vmap `wigner()` directly once we fix fft.

dynamiqs/utils/wigners.py
```diff
@@ -45,17 +45,29 @@ def wigner(
...
     if method == 'clenshaw':
         state = todm(state)
-        w = _wigner_clenshaw(state, xvec, yvec, g)
+        expanded = False
+        if state.ndim == 2:
+            state = state[None, ...]
+            expanded = True
+
+        n = state.shape[-1]
+        w = jax.vmap(_wigner_clenshaw, in_axes=(0, None, None, None, None))(
```
pierreguilmin, line 60:
```suggestion
        w = jax.vmap(_wigner_clenshaw, in_axes=(-3, None, None, None, None))(
```
Can you try and see if it works on doubly-batched arrays? (e.g. `(8, 5, 16, 16)`)

pierreguilmin (approved):
Great 😍

abocquet:
My bad the screenshot was wrong
image

dynamiqs/utils/wigners.py
abocquet: I'm open to any idea

dynamiqs/utils/wigners.py
abocquet: I'm not sure what you want to do with `-3` 🤔 
It would work with a double vmap

Maybe jitting wigner direclty would be better, I'm going to see what we can do

pierreguilmin: I thought "iterate over the third axis starting from last of the third element", to support `(..., n, n)` and not just `(b, n, n)`, but it's maybe not the correct way to do. We want the function to work with arbitrarily batched input.

dynamiqs/utils/wigners.py
pierreguilmin: I'll think about it 😊

abocquet:
You were right, it's much more natural to batch `wigner`

pierreguilmin:
I'm confused, it's jitted but it's not batched?

abocquet:
No, if you want to batch you'd have to use `vmap`

pierreguilmin:
Isn't the goal of the PR to make wigner batch-compatible to use it in `plot_wigners_mosaic`?

pierreguilmin (approved):

dynamiqs/utils/wigners.py
```diff
@@ -15,6 +15,8 @@
 __all__ = ['wigner']
 
 
+@ft.partial(jnp.vectorize, signature='(n,m)->(k),(l),(k,l)', excluded={1, 2, 3})
+@ft.partial(jax.jit, static_argnums=(3, 4))
```
pierreguilmin, line 19: Should we reverse both? I don't know if vectorize jits by default.

pierreguilmin (approved):
Beautiful!

dynamiqs/utils/wigners.py
abocquet: It's sensitively equivalent

dynamiqs/utils/wigners.py
```diff
@@ -73,6 +72,7 @@ def wigner(
     return xvec, yvec, w
 
 
+@ft.partial(jax.jit, static_argnums=(4,))
```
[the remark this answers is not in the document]
abocquet: Yes

pierreguilmin (approved):

PR #417, opened 2024-01-18
[2/2] Wigner gif

abocquet:
And fixes a bug with exp ops that appeared with JAX upgrade I think
Part of 
- https://github.com/dynamiqs/dynamiqs/pull/416
- [2/2]

### DIFF
--- a/dynamiqs/plots/plots_wigner.py
+++ b/dynamiqs/plots/plots_wigner.py
@@ -234,7 +234,7 @@ def plot_wigner_mosaic(
 
 def plot_wigner_gif(
     states: ArrayLike,
-    length_s: int = 5,
+    length_s: float = 5.0,
     fps: int = 10,
     w: float = 5.0,
     h: float = 5.0,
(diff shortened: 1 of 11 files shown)


dynamiqs/plots/plots_wigner.py
```diff
@@ -224,3 +230,64 @@ def plot_wigner_mosaic(
             clear=False,
         )
         ax.set(xlabel='', ylabel='', xticks=[], yticks=[])
+
+
+def plot_wigner_gif(
+    states: ArrayLike,
+    length_s: int = 5,
```
gautierronan, line 237: What is this? Could we find a more explicit name in the absence of documentation?

abocquet: It plots a wigner in a gif. I will write a doc

gautierronan: How about `gif_duration` then?

gautierronan: And it should be a float I guess.

gautierronan (approved):

PR #418, opened 2024-01-19
JIT ptrace

abocquet:
And adds a few random functions 

closes DYN-21

### DIFF
--- a/dynamiqs/rand.py
+++ b/dynamiqs/rand.py
@@ -1,0 +1,28 @@
+from jax import numpy as jnp, Array
+import jax
+
+from dynamiqs.utils import dag  # todo: clean this dependency
+
+
+def hermitian(n: int, key=None) -> Array:
+    key = key if key is not None else jax.random.PRNGKey(0)
+    key1, key2 = jax.random.split(key, 2)
+    mat = 2 * jax.random.uniform(key1, (n, n)) - 1
+    mat = mat + 1j * (2 * jax.random.uniform(key2, (n, n)) - 1)
+    mat = 0.5 * (mat + dag(mat))
+    return mat
+
+
+def dm(n: int, key=None) -> Array:
+    mat = hermitian(n, key)
+    mat = mat.at[jnp.diag_indices(n)].set(jnp.abs(jnp.diag(mat)) + jnp.sqrt(2) * n)
+    mat /= jnp.trace(mat)
+    return mat
+
+
+def ket(n: int, key=None) -> Array:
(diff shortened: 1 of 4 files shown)


dynamiqs/rand.py
```diff
@@ -0,0 +1,34 @@
...
+    mat = jax.random.normal(key1, dims) + 1j * jax.random.normal(key2, dims)
+    return mat
+
+
+def hermitian(n: int, key=None) -> Array:
+    key = key if key is not None else jax.random.PRNGKey(0)
+    key1, key2 = jax.random.split(key, 2)
+    mat = 2 * jax.random.uniform(key1, (n, n)) - 1
+    mat = mat + 1j * (2 * jax.random.uniform(key2, (n, n)) - 1)
+    mat = 0.5 * (mat + dag(mat))
```
gautierronan, line 19:
```suggestion
    mat = complex_matrix((n, n), key=key)
    return 0.5 * (mat + dag(mat))
```
Why not something like this?

dynamiqs/rand.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def complex_matrix(dims: tuple[int, ...], key=None) -> Array:
+    key = key if key is not None else jax.random.PRNGKey(0)
+    key1, key2 = jax.random.split(key, 2)
+    mat = jax.random.normal(key1, dims) + 1j * jax.random.normal(key2, dims)
+    return mat
+
+
+def hermitian(n: int, key=None) -> Array:
```
gautierronan, line 14:
```suggestion
def hermitian_matrix(dims: tuple[int, ...], key=None) -> Array:
```
For consistency with `complex_matrix`.

gautierronan: Same comment for `dm` and `ket`.

gautierronan (changes requested):
Nice ! Nit comments below.

dynamiqs/rand.py
abocquet: Honestly I copy-pasted qutip's code, but this may be simpler indeed

dynamiqs/rand.py
abocquet: It's just that `complex` is a reserved keywork of Python

gautierronan: My comment was also about the typing of the `dims` argument. You use `dims` in `complex_matrix`, and `n` in other functions.

gautierronan (approved):

PR #421, opened 2024-01-29
fix hinton labels

abocquet:
image

is coherent with the swapped labels

image

### DIFF
--- a/dynamiqs/plots/plots_hinton.py
+++ b/dynamiqs/plots/plots_hinton.py
@@ -259,8 +259,8 @@ def plot_hinton(
         ax.xaxis.set_ticklabels(tickslabel)
         ax.yaxis.set_ticklabels(tickslabel)
 
-    ket_ticks(ax.xaxis)
-    bra_ticks(ax.yaxis)
+    bra_ticks(ax.xaxis)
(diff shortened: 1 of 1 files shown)


pierreguilmin:
It's a matter of convention. For

$$
\rho=\begin{pmatrix}a & b\\\c & d\end{pmatrix}=a\ket0\bra0 + b \ket0\bra1 + c\ket1\bra0 + d\ket1\bra1,
$$

the value of the plot at row $i$ and column $j$ is always $\rho_{ij}$. Now there are two ways to interpret the labels:
- With the original labels the element at row $\bra i$ and colum $\ket j$ is $\braket{i|\rho|j}$.
- With the new labels the element at row $\ket i$ and column $\bra j$ is the the corresponding element $\ket i\bra j$ of $rho$.

IMO we should keep the original labels, as they align with the standard mathematical notation for matrices (usually when plotting a matrix we plot its action on the column, for example the first column is $\rho\ket0$).

abocquet:
The two ways to interpret the labels are not completely interchangeable, I believe there is a transposition to switch from one to another

pierreguilmin:
Can I close this @abocquet?

PR #427, opened 2024-02-02
Simplify rand.py and add dtype

pierreguilmin:
On top of #405.

### DIFF
--- a/dynamiqs/rand.py
+++ b/dynamiqs/rand.py
@@ -11,15 +11,11 @@ from .utils.array_types import dtype_complex_to_real, get_cdtype
 
 
 def matrix(
+    key: PRNGKeyArray,
     shape: tuple[int, ...],
     *,
-    key: PRNGKeyArray | None = None,
     dtype: jnp.complex64 | jnp.complex128 | None = None,
 ) -> Array:
-    # seed PRNG key if None
-    if key is None:
-        key = jax.random.PRNGKey(42)
-
     rdtype = dtype_complex_to_real(get_cdtype(dtype))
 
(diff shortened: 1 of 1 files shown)


dynamiqs/rand.py
```diff
@@ -1,32 +1,72 @@
...
+
+def matrix(
+    shape: tuple[int, ...],
+    *,
+    key: PRNGKeyArray | None = None,
+    dtype: jnp.complex64 | jnp.complex128 | None = None,
+) -> Array:
+    # seed PRNG key if None
+    if key is None:
+        key = jax.random.PRNGKey(42)
```
abocquet, line 21: This is IMO a turbo footgun. I removed default argument from my preceding PR because with default people can **think** they are drawing random numbers but they are not until they provide with the key

dynamiqs/rand.py
```diff
@@ -1,32 +1,72 @@
...
-def matrix(dims: tuple[int, ...], key) -> Array:
     key1, key2 = jax.random.split(key, 2)
-    mat = jax.random.normal(key1, dims) + 1j * jax.random.normal(key2, dims)
-    return mat
+    x = jax.random.normal(key1, shape, dtype=rdtype)
+    y = jax.random.normal(key2, shape, dtype=rdtype)
+    return x + 1j * y
+
+
+def herm(
```
abocquet, line 31: As usual, I'd prefer a slightly longer but clear name (e.g `hermitian`)

abocquet (changes requested):
Thank you for completing this module! I'm strongly in favor of letting the key argument (cf comment)

gautierronan (dismissed):

PR #429, opened 2024-02-05
JIT `sesolve`

pierreguilmin:
Everyone is a PyTree, everyone is happy ☀️ We turn all objects to equinox module to make sesolve JIT-compatible.

Notes:
- I had to remove the dtype argument of options bc jnp.dtype is not a JAX type
- I remove result timing as it's not straightfwd to implement

Here's an example:
```python
import dynamiqs as dq
from jax import numpy as jnp
from time import time

n = 64
a = dq.destroy(n)
H = dq.dag(a) @ a
psi0 = dq.coherent(n, 1.0)

# JIT and execute
start = time()
dq.sesolve(H, psi0, tsave).ysave.block_until_ready()
print(f'{(time()-start)*1000:.2f} ms')  # 484.34 ms

# execute again
start = time()
dq.sesolve(H, psi0, tsave).ysave.block_until_ready()
print(f'{(time()-start)*1000:.2f} ms')  # 1.02 ms

# check that a different H with same dtype and shape doesn't re-JIT
H = dq.dag(a) @ a + a + dq.dag(a)
start = time()
dq.sesolve(H, psi0, tsave).ysave.block_until_ready()
print(f'{(time()-start)*1000:.2f} ms')  # 1.29 ms
```

### DIFF
--- a/dynamiqs/__init__.py
+++ b/dynamiqs/__init__.py
@@ -6,7 +6,7 @@ from .mesolve import mesolve
 from .plots import *
 from .result import Result
 from .sesolve import sesolve
-from .time_array import TimeArray, totime
+from .time_array import TimeArray, TimeArrayLike, totime
 from .utils import *
 
 # get version from pyproject.toml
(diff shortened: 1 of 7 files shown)


gautierronan (dismissed):
LGTM

PR #430, opened 2024-02-06
Fix sesolve tests

pierreguilmin:
Fixes correctness and gradient test for euler and adaptive (for the cavity system).


tests/solver_tester.py
```diff
@@ -11,227 +11,79 @@
...
+        true_Esave = system.expects(system.tsave)
         logging.warning(f'true_Esave = {true_Esave}')
-        assert jnp.allclose(result.Esave, true_Esave, rtol=esave_rtol, atol=esave_atol)
+        logging.warning(f'Esave      = {result.Esave}')
+        assert jnp.allclose(true_Esave, result.Esave, rtol=esave_rtol, atol=esave_atol)
 
     def test_correctness(self):
         pass
 
-    def _test_jit(
```
gautierronan, line 140: Why remove the jit test?

gautierronan (approved):
LGTM. Just one question below. Also, I'll look into the tdqubit tests asap.

tests/solver_tester.py
pierreguilmin: I don't see the point of this test? If it's to test that the code works with JIT, it's jitted by default so already tested by the rest. If it's to test wether the jitted code runs faster, the answer will always be yes (I don't see a bug/situation where it's not the case). Moreover the no-jit code takes quite long to run (on the order of a few seconds), so once we test all the solvers it would slow down the CI quite a lot.

PR #431, opened 2024-02-07
Implement base solver

pierreguilmin:
This separates the API layer from the diffrax solvers implementation. It also allows implementing custom solvers very easily (not necessarily based on diffrax). I will use this to implement the propagator solver in a subsequent PR.

### DIFF
--- a/dynamiqs/core/diffrax_solver.py
+++ b/dynamiqs/core/diffrax_solver.py
@@ -31,6 +31,8 @@ class DiffraxSolver(BaseSolver):
     term: dx.ODETerm
 
     def __init__(self, *args):
+        # this dummy init is needed because of the way the class hierarchy is set up,
+        # to have subsequent init working properly
         super().__init__(*args)
 
     def run(self) -> PyTree:
--- a/dynamiqs/core/diffrax_solver.py
+++ b/dynamiqs/core/diffrax_solver.py
@@ -4,8 +4,6 @@ import warnings
 from collections import namedtuple
 
 import diffrax as dx
-
-# from equinox import AbstractVar
 from jaxtyping import PyTree
 
 from .._utils import bexpect
--- a/dynamiqs/core/diffrax_solver.py
+++ b/dynamiqs/core/diffrax_solver.py
@@ -1,0 +1,99 @@
+from __future__ import annotations
+
+import warnings
+from collections import namedtuple
+
+import diffrax as dx
+
+# from equinox import AbstractVar
+from jaxtyping import PyTree
+
+from .._utils import bexpect
+from ..gradient import Adjoint, Autograd
+from ..result import Result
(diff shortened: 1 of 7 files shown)


dynamiqs/solver.py
```diff
@@ -1,27 +1,42 @@
...
 from jaxtyping import Scalar
 
 from .gradient import Adjoint, Autograd, Gradient
 
+NoneType = type(None)
+
 
 class Solver(eqx.Module):
-    SUPPORTED_GRADIENT: ClassVar[tuple[Gradient]] = ()
+    SUPPORTED_GRADIENT: ClassVar[tuple[Gradient | NoneType]] = (NoneType,)
```
abocquet, line 14: Is is true to say that all solvers can accept `None` as gradient? If so we could probably strip this `NoneType` type away

dynamiqs/core/diffrax_solver.py
```diff
@@ -0,0 +1,99 @@
+from __future__ import annotations
+
+import warnings
+from collections import namedtuple
+
+import diffrax as dx
+
+# from equinox import AbstractVar
```
abocquet, line 8:
```suggestion
```

dynamiqs/solver.py
pierreguilmin: It's true for now so let's remove it yes 👍🏼

dynamiqs/core/diffrax_solver.py
```diff
@@ -0,0 +1,99 @@
...
+
+class DiffraxSolver(BaseSolver):
+    diffrax_solver: dx.AbstractSolver
+    stepsize_controller: dx.AbstractAdaptiveStepSizeController
+    dt0: float | None
+    max_steps: int
+    term: dx.ODETerm
+
+    def __init__(self, *args):
+        super().__init__(*args)
```
gautierronan, line 36:
```suggestion
```

gautierronan (approved):
LGTM :)

dynamiqs/core/diffrax_solver.py
pierreguilmin: Fun fact, we need this 😅 
```python
import equinox as eqx

class A(eqx.Module):
    a: int

class B(A):
    b: int

#    def __init__(self, *args):
#        super().__init__(*args)

class C(B):
    def __init__(self, *args):
        super().__init__(*args)
        self.b = 2 * self.a

C(1)  # TypeError: __init__() missing 1 required positional argument: 'b'
```
If you uncomment the init of `B` it works as intended...

Maybe I'm missing smthg obvious here, cc @abocquet.

gautierronan: Ok that's weird ! Not super important anyway, let's keep it ^^

PR #434, opened 2024-02-07
Simplify sesolve mesolve

pierreguilmin:
On top of #433.

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -38,7 +38,7 @@ def mesolve(
     y0 = jnp.asarray(psi0, dtype=options.cdtype)
     y0 = todm(y0)
     ts = jnp.asarray(tsave, dtype=options.rdtype)
-    E = jnp.asarray(exp_ops, dtype=options.cdtype) if exp_ops is not None else None
+    Es = jnp.asarray(exp_ops, dtype=options.cdtype) if exp_ops is not None else None
 
     # === select solver class
     solvers = {Euler: MEEuler, Dopri5: MEDopri5}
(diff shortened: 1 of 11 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -1,25 +1,26 @@
...
+from ..time_array import TimeArray
 from ..utils.utils import todm
 from .mediffrax import MEDopri5, MEEuler
 
 
 @partial(jax.jit, static_argnames=('solver', 'gradient', 'options'))
 def mesolve(
     H: ArrayLike | TimeArray,
-    jump_ops: ArrayLike,
+    jump_ops: ArrayLike | list[ArrayLike | TimeArray],
```
abocquet, line 23:
```suggestion
    jump_ops: ArrayLike | TimeArray | list[ArrayLike | TimeArray],
```

abocquet (dismissed):

dynamiqs/mesolve/mesolve.py
pierreguilmin: I think we want `jump_ops` to be an iterable. If the user has a single time-dependent `jump_ops` he should pass a list with one element. The first `ArrayLike` typing is to account for a list of `jump_ops` already stacked. Maybe for clarity we can replace the `jump_ops` typing by the list only, and the `exp_ops` typing by a list too (even though it's a little imprecise, because it does not reflect that a stacked array can be passed, but it's much clearer that the expected object should have dimension `(nL, n, n)`).

gautierronan: Agreed as well !

PR #435, opened 2024-02-07
Fix doctest figures

pierreguilmin:
On top of #434.

### DIFF
--- a/tests/test_plots.py
+++ b/tests/test_plots.py
@@ -1,6 +1,7 @@
 import jax.numpy as jnp
 import numpy as np
 import pytest
+from matplotlib import pyplot as plt
 
 from dynamiqs import coherent, plot_wigner, plot_wigner_mosaic, todm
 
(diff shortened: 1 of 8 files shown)


tests/test_plots.py
```diff
@@ -16,6 +17,11 @@ def setup(self):
         self.psis = jnp.asarray(self.psis)
         self.rhos = jnp.asarray(self.rhos)
 
+    @pytest.fixture(autouse=True)
+    def teardown(self):
+        yield
```
abocquet, line 22: Why do you need to yield here ? A small comment would be useful

abocquet (changes requested):

tests/test_plots.py
abocquet: LGTM otherwise

pierreguilmin: It's because it's an end-of-test fixture, first the test runs and then this is called. This is how pytest proposes to write this kind of fixture.

abocquet, line 22:
```suggestion
        yield  # see https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended
```

abocquet (approved):

gautierronan (approved):

PR #437, opened 2024-02-08
Batch mesolve

abocquet:
Here I use the convention that the dissipators batching dimension is always outwards because it makes the reasoning much simpler for this code, but if there is a good reason to adopt another one, we can switch.

On top of #436

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
         # os: [ubuntu-latest, macOS-latest, windows-latest]
         os: [ubuntu-latest]
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Requirements
 
-The project was written using Python 3.8+, you must have a compatible version of Python (i.e. >= 3.8) installed on your computer.
+The project was written using Python 3.9+, you must have a compatible version of Python (i.e. >= 3.9) installed on your computer.
 
 ## Setup
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 [P. Guilmin](https://github.com/pierreguilmin), [R. Gautier](https://github.com/gautierronan), [A. Bocquet](https://github.com/abocquet), [E. Genois](https://github.com/eliegenois)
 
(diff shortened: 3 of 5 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -27,6 +28,39 @@ def mesolve(
...
+        options.cartesian_batching,
+        args,
+        out_axes,
+        H.ndim > 2,
+        jump_ops.ndim > 3,
+        psi0.ndim > 2,
+    )
+
+    # === apply vectorized function
+    return f(H, jump_ops, psi0, tsave, exp_ops, solver, gradient, options)
```
gautierronan, line 51:
```suggestion
    # compute in_axes to vectorize over
    batch_axes = (H.ndim > 2, jump_ops.ndim > 3, psi0.ndim > 2, None, None, None, None, None)
    in_axes = compute_in_axes(batch_axes, options.cartesian_batching)
    
    # the result is vectorized over ysave and Esave
    out_axes = Result(None, None, None, None, 0, 0)

    batched_mesolve = jax.vmap(_mesolve, in_axes=in_axes, out_axes=out_axes)

    # === apply vectorized function
    return batched_mesolve(H, jump_ops, psi0, tsave, exp_ops, solver, gradient, options)
```

dynamiqs/_utils.py
```diff
@@ -41,3 +42,27 @@ def bexpect(O: Array, x: Array) -> Array:
     if isket(x):
         return jnp.einsum('ij,bjk,kl->b', dag(x), O, x)  # <x|O|x>
     return jnp.einsum('bij,ji->b', O, x)  # tr(Ox)
+
+
+def compute_batching(f, cartesian_batching, other_in_axes, out_axes, *should_batch):
```
gautierronan, line 47:
```suggestion
def compute_in_axes(batch_axes, cartesian_batching):
```
(and corresponding function modifications so as to return the `in_axes` argument)

gautierronan (changes requested):
Nice! I like the overall idea of the PR, but I'd suggest the changes below to remedy two issues that I find in the current form:
 - it's lacking readability (mesolve is the first function people will look at when reading source code: it should be very easy to follow IMO),
 - `compute_batching` does not support batching over arguments that are not within the first function arguments (useful for `smesolve`).

abocquet:
I understand your concern but 
- the code proposal does not work as is since we need a for loop to vmap properly. If we chose to split the code according to your comment the loop would be in the mesolve code. The complexity does not dise appear, it is moved outside of `compute_batching` 
- one does not need to understand the batching mechanism during the first read. The goal of `compute_batching` is to abstract this mechanism away. This way of splitting concerns allows for understanding mesolve aside for batching and come back later if this is the part of interest 
- regarding smesolve this is a problem we can address later and probably the `ntraj` dimension (if this is what you are thinking about), could be in the first arguments of `_smsesolve` which solves the problem

abocquet:
I shot myself with git but the PR is still open, we can continue discussing it while I open a new PR

PR #440, opened 2024-02-08
Batch mesolve

abocquet:
Re-edition of https://github.com/dynamiqs/dynamiqs/pull/437 that died because of an infortunate force-push 

# Original comment : 
Here I use the convention that the dissipators batching dimension is always outwards because it makes the reasoning much simpler for this code, but if there is a good reason to adopt another one, we can switch.

# @gautierronan comment

Nice! I like the overall idea of the PR, but I'd suggest the changes below (or similar) to remedy two issues that I find in the current form:
- it's lacking readability (mesolve is the first function people will look at when reading source code: it should be very easy to follow IMO),
- compute_batching does not support batching over arguments that are not within the first function arguments (useful for smesolve).

## Answer 

I understand your concern but 
- the code proposal does not work as is since we need a for loop to vmap properly. If we chose to split the code according to your comment the loop would be in the mesolve code. The complexity does not disappear, it is moved outside of compute_batching
-  I am not sure how to make very easy to read of piece of code with 16 possible cases (32 for smesolve) without explicitly code them all or do tedious list/dict based case handling
- one does not need to understand the batching mechanism during the first read. The goal of compute_batching is to abstract this mechanism away. This way of splitting concerns allows for understanding mesolve aside for batching and come back later if this is the part of interest
- regarding smesolve this is a problem we can address later and probably the ntraj dimension (if this is what you are thinking about), could be in the first arguments of _smesolve which solves the problem

### DIFF
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -2,8 +2,6 @@ from __future__ import annotations
 
 from typing import Any
 
-import jax
-import numpy as np
 from jax import numpy as jnp
 from jaxtyping import Array
 
@@ -43,27 +41,3 @@ def bexpect(O: Array, x: Array) -> Array:
     if isket(x):
         return jnp.einsum('ij,bjk,kl->b', dag(x), O, x)  # <x|O|x>
     return jnp.einsum('bij,ji->b', O, x)  # tr(Ox)
-
-
-def compute_vmap(
-    f: callable,
-    cartesian_batching: bool,
-    is_batched: list[bool],
-    out_axes: list[int | None],
-) -> callable:
-    if any(is_batched):
-        if cartesian_batching:
-            # iteratively map over the first axis of each batched argument
-            idx_batched = np.where(is_batched)[0]
-            # we apply the successive vmaps in reverse order, so that the output
-            # batched dimensions are in the correct order
-            for i in reversed(idx_batched):
-                in_axes = [None] * len(is_batched)
-                in_axes[i] = 0
-                f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
-        else:
-            # map over the first axis of all batched arguments
-            in_axes = list(np.where(is_batched, 0, None))
-            f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
-
-    return f
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -55,8 +55,8 @@ def compute_vmap(
         if cartesian_batching:
             # iteratively map over the first axis of each batched argument
             idx_batched = np.where(is_batched)[0]
-             # we apply the succesive vmaps in reverse order, so that the output
-             # batched dimensions are in the correct order
+            # we apply the successive vmaps in reverse order, so that the output
+            # batched dimensions are in the correct order
             for i in reversed(idx_batched):
                 in_axes = [None] * len(is_batched)
                 in_axes[i] = 0
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import Any
 
+import jax
 from jax import numpy as jnp
 from jaxtyping import Array
 
@@ -41,3 +42,27 @@ def bexpect(O: Array, x: Array) -> Array:
     if isket(x):
         return jnp.einsum('ij,bjk,kl->b', dag(x), O, x)  # <x|O|x>
     return jnp.einsum('bij,ji->b', O, x)  # tr(Ox)
+
+
+def compute_batching(f, cartesian_batching, other_in_axes, out_axes, *should_batch):
+    # Build a list of batch mask (ie binary numbers). Each digit is 1 if we
+    # should batch over the dimension and 0 otherwise.
+    # For example if `should_batch` = (True, False, True), the batch_mask will be
+    # [001, 100]
+    batch_mask = [2**k * v for k, v in enumerate(should_batch)]
+
+    # When doing cartesian batching, we need to batch over all dimensions
+    # one by one, but when cartesian batching is off, we want to batch over
+    # every dimension at once, hence the sum.
+    # For example if `should_batch` = (True, False, True), the batch_mask will be
+    # [001, 100] if we want to do cartesian batching and [101] otherwise.
+    if not cartesian_batching:
+        batch_mask = [sum(batch_mask)]
+
+    for mask in reversed(batch_mask):
+        if mask == 0:  # skip the no batching case
+            continue
+        mask = tuple([0 if mask & (1 << i) else None for i in range(len(should_batch))])
+        f = jax.vmap(f, in_axes=mask + other_in_axes, out_axes=out_axes)
+
+    return f
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import Any
 
 import jax
+import numpy as np
 from jax import numpy as jnp
 from jaxtyping import Array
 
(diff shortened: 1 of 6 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -19,11 +20,43 @@
...
+    )
+
+    # === apply vectorized function
+    return f(H, jump_ops, psi0, tsave, exp_ops, solver, gradient, options)
+
+
+@partial(jax.jit, static_argnames=('solver', 'gradient', 'options'))
+def _mesolve(
+    H: ArrayLike | TimeArray,
+    jump_ops: ArrayLike | list[ArrayLike | TimeArray],
```
pierreguilmin, line 56:
```suggestion
    jump_ops: list[ArrayLike | TimeArray],
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -19,11 +20,43 @@
...
+    psi0: ArrayLike,
+    tsave: ArrayLike,
+    exp_ops: ArrayLike | None = None,
```
pierreguilmin, line 59:
```suggestion
    exp_ops: list[ArrayLike] | None = None,
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -19,11 +20,43 @@
 @partial(jax.jit, static_argnames=('solver', 'gradient', 'options'))
 def mesolve(
     H: ArrayLike | TimeArray,
-    jump_ops: list[ArrayLike | TimeArray],
+    jump_ops: ArrayLike | list[ArrayLike | TimeArray],
```
pierreguilmin, line 23: Revert this chunk.

dynamiqs/mesolve/mesolve.py
```diff
@@ -19,11 +20,43 @@
...
     psi0: ArrayLike,
     tsave: ArrayLike,
     *,
-    exp_ops: list[ArrayLike] | None = None,
+    exp_ops: ArrayLike | None = None,
```
pierreguilmin, line 27: Revert this chunk.

tests/mesolve/test_batching.py
```diff
@@ -0,0 +1,72 @@
...
+    nt = 11
+    nH = 3
+    nLs = 4 if cartesian_batching else nH
+    npsi0 = 5 if cartesian_batching else nH
+    nEs = 6
+
+    options = dq.options.Options(cartesian_batching=cartesian_batching)
+
+    # create random objects
+    k1, k2, k3, k4 = jax.random.split(jax.random.PRNGKey(42), 4)
```
pierreguilmin, line 20: Can you do the same for sesolve pls?

tests/mesolve/test_batching.py
```diff
@@ -0,0 +1,72 @@
...
+    H = dq.rand.herm(k1, (nH, n, n))
+    Ls = dq.rand.herm(k2, (nLs, 1, n, n))
```
pierreguilmin, line 22:
```suggestion
    Ls = dq.rand.herm(k2, (nLs, 2, n, n))
```
to make sure the sum over `jump_ops` happens correctly.

dynamiqs/_utils.py
```diff
@@ -41,3 +42,27 @@ def bexpect(O: Array, x: Array) -> Array:
     if isket(x):
         return jnp.einsum('ij,bjk,kl->b', dag(x), O, x)  # <x|O|x>
     return jnp.einsum('bij,ji->b', O, x)  # tr(Ox)
+
+
+def compute_batching(f, cartesian_batching, other_in_axes, out_axes, *should_batch):
```
pierreguilmin, line 47: Can we call this `compute_vmap`?

tests/mesolve/test_batching.py
```diff
@@ -0,0 +1,72 @@
...
+    result = dq.mesolve(H[0], Ls[0], psi0[0], tsave, exp_ops=exp_ops, options=options)
+    assert result.ysave.shape == (nt, n, n)
+    assert result.Esave.shape == (nEs, nt)
+
+    # H batched
+    result = dq.mesolve(H, Ls[0], psi0[0], tsave, exp_ops=exp_ops, options=options)
+    assert result.ysave.shape == (nH, nt, n, n)
+    assert result.Esave.shape == (nH, nEs, nt)
+
+    # jump ops batched
```
pierreguilmin, line 37:
```suggestion
    # Ls batched
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -19,11 +20,43 @@
...
+    solver: Solver = Dopri5(),
+    gradient: Gradient | None = None,
+    options: Options = Options(),
+):
+    # === vectorize function
+
```
pierreguilmin, line 33:
```suggestion
```

dynamiqs/sesolve/sesolve.py
```diff
@@ -27,22 +28,14 @@ def sesolve(
     options: Options = Options(),
 ):
     # === vectorize function
-    f = _sesolve
```
pierreguilmin, line 31:
```suggestion
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -19,11 +20,43 @@
...
+        H.ndim > 2,
+        jump_ops.ndim > 3,
+        psi0.ndim > 2,
+    )
+
+    # === apply vectorized function
+    return f(H, jump_ops, psi0, tsave, exp_ops, solver, gradient, options)
+
+
+@partial(jax.jit, static_argnames=('solver', 'gradient', 'options'))
```
pierreguilmin, line 53:
```suggestion
```

dynamiqs/_utils.py
```diff
@@ -41,3 +42,27 @@ def bexpect(O: Array, x: Array) -> Array:
...
+    if not cartesian_batching:
+        batch_mask = [sum(batch_mask)]
+
+    for mask in reversed(batch_mask):
+        if mask == 0:  # skip the no batching case
+            continue
+        mask = tuple([0 if mask & (1 << i) else None for i in range(len(should_batch))])
+        f = jax.vmap(f, in_axes=mask + other_in_axes, out_axes=out_axes)
+
+    return f
```
pierreguilmin, line 68: I like the idea, but the code is very cryptic honestly. What about:
```python
def compute_batching(f, cartesian_batching, is_batched, out_axes):
    if any(is_batched):
        if cartesian_batching:
            idx_batched = np.where(is_batched)[0]
            for i in idx_batched[::-1]:
                in_axes = [None] * len(is_batched)
                in_axes[i] = 0
                f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
        else:
            in_axes = list(np.where(is_batched, 0, None))
            f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)

    return f
```
(with typing + comments) and e.g. for sesolve
```python
    # we vectorize over H and psi0, all other arguments are not vectorized
    is_batched = (H.ndim > 2, psi0.ndim > 2, False, False, False, False, False)
    # the result is vectorized over ysave and Esave
    out_axes = Result(None, None, None, None, 0, 0)
    f = compute_batching(_sesolve, options.cartesian_batching, is_batched, out_axes)
```

pierreguilmin (changes requested):
Nice!

dynamiqs/_utils.py
pierreguilmin: Or
```
for i, v in enumerate(is_batched[::-1]):
    if v:
        in_axes = [None] * len(is_batched)
        in_axes[i] = 0
        f = jax.vmap(f, in_axes=in_axes[::-1], out_axes=out_axes)
```
for the `cartesian_batching` option.

dynamiqs/mesolve/mesolve.py
abocquet: Because of messy rebases sorry

dynamiqs/_utils.py
abocquet: Much better indeed

pierreguilmin (dismissed):
Amazing! 🥳

dynamiqs/mesolve/mesolve.py
```diff
@@ -6,6 +6,7 @@
 from jax import numpy as jnp
 from jaxtyping import ArrayLike
 
+from dynamiqs import compute_vmap
```
pierreguilmin, line 9: Ho and you need to make this import relative.

gautierronan (approved):

pierreguilmin:
@abocquet you need to fix the import of compute_vmap in sesolve and then run isort to have the CI passing.

pierreguilmin:
Ha and also (sorry for adding and adding on the list 😅):
```
        is_batched = (
            H.ndim > 2,
>           jump_ops.ndim > 3,
            psi0.ndim > 2,
            False,
            False,
            False,
            False,
            False,
        )
E       AttributeError: 'list' object has no attribute 'ndim'
```

PR #443, opened 2024-02-09
Remove `dtype` argument

pierreguilmin:
On top of #440.

As discussed tgther, this is now fixed by a global option that we can query using `cdtype()`.

### DIFF
--- a/dynamiqs/utils/array_types.py
+++ b/dynamiqs/utils/array_types.py
@@ -15,52 +15,11 @@ def _get_default_dtype() -> jnp.float32 | jnp.float64:
     return jnp.float64 if default_dtype == jnp.float64 else jnp.float32
 
 
-def get_cdtype(
-    dtype: jnp.complex64 | jnp.complex128 | None = None,
-) -> jnp.complex64 | jnp.complex128:
-    if dtype is None:
-        # the default dtype for complex arrays is determined by the default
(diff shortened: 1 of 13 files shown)


dynamiqs/utils/array_types.py
```diff
@@ -15,52 +15,11 @@ def _get_default_dtype() -> jnp.float32 | jnp.float64:
...
-) -> jnp.complex64 | jnp.complex128:
-    if dtype == jnp.float32:
-        return jnp.complex64
-    elif dtype == jnp.float64:
-        return jnp.complex128
+def cdtype() -> jnp.complex64 | jnp.complex128:
+    # the default dtype for complex arrays is determined by the default floating point
+    # dtype (jnp.complex128 if default is jnp.float64, otherwise jnp.complex64)
+    default_dtype = _get_default_dtype()
+    return jnp.complex128 if default_dtype is jnp.float64 else jnp.complex64
```
gautierronan, line 22:
```suggestion
    if default_dtype is jnp.float64:
        return jnp.complex128
    elif default_dtype is jnp.float32:
        return jnp.complex64
    else:
        raise ValueError(f"Data type {default_dtype} is not yet supported.")
    return jnp.complex128 if default_dtype is jnp.float64 else jnp.complex64
```

gautierronan (changes requested):
nit below

gautierronan (approved):

PR #447, opened 2024-02-09
Add propagator solver for sesolve and mesolve

pierreguilmin:
If you find better ideas for some of the proposed code/fixes, you're very welcome! 😊 

Note that for constant Hamiltonian, this solver is much faster for sesolve (~x10). Of course it would be even better with the caching (for now we recompute the propagator at each time interval, even if `delta_t` is constant). I think this solver will be a good start to implement the first caching mechanisms, because (1) it's very valuable and (2) the solver is super simple.


dynamiqs/core/abstract_solver.py
```diff
@@ -21,15 +23,31 @@ def run(self) -> PyTree:
...
+
+        return Result(self.ts, self.solver, self.gradient, self.options, ysave, Esave)
+
 
 SESolver = BaseSolver
 
 
 class MESolver(BaseSolver):
-    Ls: list[TimeArray]
+    Ls: list[Array | TimeArray]
```
abocquet, line 53: Why not `ArrayLike`?

abocquet (approved):

dynamiqs/core/abstract_solver.py
pierreguilmin: This is not the API layer anymore so we type everything with fully converted types in the backend.

dynamiqs/core/propagator_solver.py
```diff
@@ -0,0 +1,57 @@
...
+        # check that Hamiltonian is time-independent
+        if not isinstance(self.H, ConstantTimeArray):
+            raise TypeError(
+                'Solver `Propagator` requires a time-independent Hamiltonian.'
+            )
+        self.H = self.H.x
+
+    def run(self) -> PyTree:
+        # === solve differential equation
+        def f(y, delta_t):
```
gautierronan, line 26:
```suggestion
        def propagate(y, delta_t):
```
You seem to like naming everything `f` these days, but it's not very explicit :p

dynamiqs/core/propagator_solver.py
```diff
@@ -0,0 +1,57 @@
...
+            # propagate forward except if delta_t is zero
+            y = jax.lax.cond(delta_t == 0, lambda: y, lambda: self.forward(delta_t, y))
+            # save result
+            res = self.save(y)
```
gautierronan, line 30: Shouldn't we save only if `delta_t == 0` ?

dynamiqs/core/propagator_solver.py
```diff
@@ -0,0 +1,57 @@
...
+            return y, res
+
+        # todo: fix next line once t0 option is implemented
+        delta_ts = jnp.diff(jnp.concatenate((jnp.zeros(1), self.ts)))
```
gautierronan, line 34: What if `self.ts[0] == 0` already?

gautierronan (changes requested):
Nice ! I like it :)

dynamiqs/core/propagator_solver.py
pierreguilmin: If `tsave = [0, 1, 2, 3]`, then `delta_ts = [0, 1, 1, 1]` (at first iteration we don't propagate and save directly, and for the next three iterations we propagate for `delta_t=1`). I guess this also answers next comment. Once we fix https://linear.app/dynamiqs/issue/DYN-17/add-option-to-evolve-from-t0-=-00 we can do:
```python
delta_ts = jnp.diff(jnp.concatenate((self.options.t0, self.ts)))
```

dynamiqs/core/propagator_solver.py
pierreguilmin: Cf. other reply.

gautierronan (approved):

PR #451, opened 2024-02-12
Fallback method for `dm_fidelity` on GPU

pierreguilmin:
All the tests pass on GPU now 🎉 I tested that the computed fidelity is the same as qutip on 30 random density matrices, for both the CPU and GPU implementation.

Closes https://linear.app/dynamiqs/issue/DYN-158.

### DIFF
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -32,3 +32,8 @@ def check_time_array(x: Array, arg_name: str, allow_empty: bool = False):
         )
     if not jnp.all(x >= 0):
         raise ValueError(f'Argument `{arg_name}` must contain positive values only.')
+
+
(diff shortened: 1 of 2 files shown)


gautierronan (approved):
Very nice. Should we check somewhere (maybe inside `dq.isdm`) that `x` and `y` are indeed hermitian positive semi-definite? To make sure we can actually use `eigvalsh` later on?

pierreguilmin:
> Should we check somewhere (maybe inside dq.isdm) that x and y are indeed hermitian positive semi-definite?

I'm actually implementing smthg for checks like this across the codebase (checking at various levels: the shape, the trace, the positive semi-definitness, etc...). Stay tuned 😎

PR #452, opened 2024-02-13
New ODE solvers: Dopri8 and Tsit5

pierreguilmin:
This PR introduces new ODE solvers from diffrax, each having its own motivation:
- **Dopri8**: to have very accurate solutions for demanding tolerances (probably for `float64`). It may also serve as a reference to compare fidelities between other solvers.
- **Tsit5**: as a replacement to Dopri5, the ODE community says it's more efficient. I tested on a few examples, it is indeed slightly better (by a few % in speedup). We should keep Dopri5 though if we want to also compare eye-to-eye with QuTiP.

Sources:
- Patrick Kidger diffrax doc > https://docs.kidger.site/diffrax/usage/how-to-choose-a-solver/
- Christopher Rackauckas blog > https://www.stochasticlifestyle.com/comparison-differential-equation-solver-suites-matlab-r-julia-python-c-fortran/

I also propose to reduce the default rtol/atol to match the default precision (float32), the current default is a bit low compared with machine precision.

We're only missing implicit solvers (for stiff problems), see https://linear.app/dynamiqs/issue/DYN-159/add-implicit-ode-solvers.

<details>
  <summary>Quick comparison of Tsit5 and Dopri5 on mesolve</summary>

Lossy oscillator:
```python
import dynamiqs as dq
from jax import numpy as jnp

for n in [32, 64, 128]:
    print(f'=== n={n}')
    a = dq.destroy(n)
    delta = 1.0 * 2 * jnp.pi
    alpha0 = 2.0
    kappa = 1.0 * 2 * jnp.pi / 4
    H = delta * dq.dag(a) @ a
    Ls = [jnp.sqrt(kappa) * a, jnp.sqrt(kappa) * a]
    y0 = dq.coherent(n, alpha0)

    dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Tsit5()).ysave.block_until_ready()
    %timeit dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Tsit5()).ysave.block_until_ready()
    dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Dopri5()).ysave.block_until_ready()
    %timeit dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Dopri5()).ysave.block_until_ready()
```
```
=== n=32
16.8 ms ± 58 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
17.6 ms ± 23.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
=== n=64
37.6 ms ± 22.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
39 ms ± 52 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
=== n=128
47.8 ms ± 30 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
52.1 ms ± 28.7 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Cat inflation:
```python
for n in [32, 64, 128]:
    print(f'=== n={n}')
    a = dq.destroy(n)
    alpha = 2.0
    k2 = 1.0
    H = dq.zero(n)
    Ls = [jnp.sqrt(k2) * (a @ a - alpha**2 * dq.eye(n))]
    y0 = dq.coherent(n, 0.0)
    tsave = jnp.linspace(0, 1 / k2, 11)
    dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Tsit5()).ysave.block_until_ready()
    %timeit dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Tsit5()).ysave.block_until_ready()
    dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Dopri5()).ysave.block_until_ready()
    %timeit dq.mesolve(H, Ls, y0, tsave, solver=dq.solver.Dopri5()).ysave.block_until_ready()
```
```
=== n=32
71.1 ms ± 38.2 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
75.6 ms ± 84.7 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
=== n=64
293 ms ± 544 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
305 ms ± 238 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
=== n=128
1.38 s ± 1.71 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.45 s ± 1.02 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

</details>

### DIFF
--- a/dynamiqs/core/diffrax_solver.py
+++ b/dynamiqs/core/diffrax_solver.py
@@ -65,10 +65,9 @@ class EulerSolver(DiffraxSolver):
         self.max_steps = 100_000  # todo: fix hard-coded max_steps
 
 
-class Dopri5Solver(DiffraxSolver):
+class AdaptiveSolver(DiffraxSolver):
     def __init__(self, *args):
         super().__init__(*args)
-        self.diffrax_solver = dx.Dopri5()
         self.stepsize_controller = dx.PIDController(
             rtol=self.solver.rtol,
             atol=self.solver.atol,
@@ -78,3 +77,15 @@ class Dopri5Solver(DiffraxSolver):
         )
         self.dt0 = None
         self.max_steps = self.solver.max_steps
+
+
+class Dopri5Solver(AdaptiveSolver):
+    diffrax_solver = dx.Dopri5()
+
+
+class Dopri8Solver(AdaptiveSolver):
+    diffrax_solver = dx.Dopri8()
+
+
+class Tsit5Solver(AdaptiveSolver):
+    diffrax_solver = dx.Tsit5()
--- a/dynamiqs/mesolve/mediffrax.py
+++ b/dynamiqs/mesolve/mediffrax.py
@@ -5,7 +5,13 @@ import jax.numpy as jnp
 from jaxtyping import PyTree, Scalar
 
 from ..core.abstract_solver import MESolver
-from ..core.diffrax_solver import DiffraxSolver, Dopri5Solver, EulerSolver
+from ..core.diffrax_solver import (
+    DiffraxSolver,
+    Dopri5Solver,
+    Dopri8Solver,
+    EulerSolver,
+    Tsit5Solver,
+)
 from ..time_array import TimeArray
 from ..utils.utils import dag
 
@@ -40,4 +46,12 @@ class MEEuler(MEDiffraxSolver, EulerSolver):
 
 
 class MEDopri5(MEDiffraxSolver, Dopri5Solver):
+    pass
+
+
+class MEDopri8(MEDiffraxSolver, Dopri8Solver):
+    pass
+
+
+class METsit5(MEDiffraxSolver, Tsit5Solver):
     pass
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -26,7 +26,7 @@ def mesolve(
     tsave: ArrayLike,
     *,
     exp_ops: list[ArrayLike] | None = None,
-    solver: Solver = Dopri5(),
+    solver: Solver = Tsit5(),
     gradient: Gradient | None = None,
     options: Options = Options(),
 ):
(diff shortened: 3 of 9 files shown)


gautierronan (approved):
Nice, not much to say on the PR, it looks good :)

>    Dopri8: to have very accurate solutions for demanding tolerances (probably for float64). It may also serve as a reference to compare fidelities between other solvers.

So I don't think `Dopri8` is more precise than `Dopri5` (or any other adaptive solver for that matter). It's rather a matter of stiffness of the problem. For very stiff problems, you want low-order solvers (or even implicit ones), that will naturally make small steps. For very non-stiff problems (smooth solution), you want high-order solvers (e.g. `Dopri8`) to make very large time steps at once.

What makes the precision of the solver is just `atol`/`rtol`, unless I'm missing something.

> I also propose to reduce the default rtol/atol to match the default precision (float32), the current default is a bit low compared with machine precision.

Yes I'm down for that ! Not too much though so we don't catch people by surprise. In the end, I think `rtol`/`atol` gives the precision at each time-step. So the final precision of the result should be somewhat a random walk of added `rtol`/`atol` errors at each time step. So final error = sqrt(num_tsteps) * atol or something like that.

> We're only missing implicit solvers (for stiff problems), see https://linear.app/dynamiqs/issue/DYN-159/add-implicit-ode-solvers.

Are there any in diffrax?

pierreguilmin:
Right, thanks for the clarifying comment. Is it correct that if you set a fairly low rtol/atol, Dopri5 will make extremely small steps and probably Dopri8 is better suited? I guess that's what I meant. From diffrax doc: « If you need accurate solutions at tight tolerances then try diffrax.Dopri8. ». 

> Are there any in diffrax?

Yes! See the issue description 😉

PR #455, opened 2024-02-14
Update Gradient namespace

gautierronan:
Following the benchmark of `dx.RecursiveCheckpointAdjoint` (see below), this PR removes the adjoint method altogether (at least for now).

There are now two available gradient methods:
 - `Autograd` which correponds to the jax autograd, available in diffrax using `dx.DirectAdjoint`
 - `CheckpointAutograd` corresponding to `dx.RecursiveCheckpointAdjoint` (not available for our custom solvers, for now)

The first one seems fastest if memory is not an issue. The second one solves the memory problem similarly to the adjoint method, but probably better all-around.

NB: Also fix the `dq.expect` bug by using `jnp.einsum` and the `jnp.diff(..., prepend=...)` bug.

<details>
  <summary>Benchmark of gradients in diffrax on a small problem</summary>

```python
import jax
import jax.numpy as jnp
import diffrax as dx
import dynamiqs as dq
import functools
import timeit
import warnings


# create a simple quantum problem on which to test the adjoints
@functools.partial(jax.jit, static_argnums=(1, 2))
def run_mesolve(delta, adjoint, N):
    # make quantum problem
    a = dq.destroy(N)
    i = dq.eye(N)
    H = dq.dag(a) @ dq.dag(a) @ a @ a
    L = 0.1 * a @ a
    tsave = jnp.linspace(0.0, 10.0, 100)
    rho0 = dq.coherent_dm(N, 2.0)

    # initialize diffrax solver
    def vector_field(t, y, args):
        out = -1j * args[0] * H @ y + 0.5 * L @ y @ dq.dag(L) - 0.5 * dq.dag(L) @ L @ y
        return out + dq.dag(out)

    term = dx.ODETerm(vector_field)
    solver = dx.Dopri5()
    saveat = dx.SaveAt(ts=tsave)
    stepsize_controller = dx.PIDController(rtol=1e-5, atol=1e-7)

    # run solver
    with warnings.catch_warnings():
        warnings.simplefilter('ignore', UserWarning)
        solution = dx.diffeqsolve(
            term,
            solver,
            t0=tsave[0],
            t1=tsave[-1],
            dt0=None,
            y0=rho0,
            saveat=saveat,
            stepsize_controller=stepsize_controller,
            adjoint=adjoint,
            args=(delta,),
        )

    return dq.expect(a, solution.ys[-1]).real


# initialize parameters and grad function
N = 20
delta = 0.2
grad_mesolve = jax.grad(run_mesolve, argnums=0)

[...]

PR #459, opened 2024-02-14
Refactor gradient parametrization, remove adjoint

pierreguilmin:
Replaces https://github.com/dynamiqs/dynamiqs/pull/455 (see description there) by @gautierronan. The idea to merge gradient and solver options is from @abocquet (a long time ago 😋).

Strings are not valid JAX types, so I chose a boolean flag `autograd=True/False` for diffrax solver as a temporary solution.


gautierronan:
If we use `autograd=True/False`, I am really not in favor of this solution. This is not a binary problem whatsoever, and it hides the fact that we use a widely different gradient for diffrax and non-diffrax solvers.

pierreguilmin:
Replaced by #460.

PR #465, opened 2024-02-17
Fix vmap together with Callable or Modulated TimeArrays

gautierronan:
Fixes a bug where you could not batch over a `CallableTimeArray` or `ModulatedTimeArray` (because the Pytree could not be reduced over the first batching dimension).

To fix this, this slightly changes the API:
- The user should now pass its batching array as an argument to his callable, and then pass it to `dq.totime(..., *args)`.
- I have (for now) removed the option to sum two `CallableTimeArrays` which is probably not so useful (easy to bypass for the user) and which I didn't figure out how to do.

Here is an example of a typical usage:
```python
import dynamiqs as dq
import jax.numpy as jnp

a = dq.destroy(10)
delta = jnp.linspace(-1.0, 1.0, 20)
H0 = dq.dag(a) @ a
psi = dq.coherent(10, 1.0)
tsave = jnp.linspace(0, 1.0, 100)

def _Ht(t, delta):
    return jnp.cos(t * jnp.expand_dims(delta, (-1, -2))) * H0
H = H0 - dq.totime(_Ht, delta).mT.conj()

print(dq.sesolve(H, psi, tsave))
# ==== Result ====
# Solver  : Tsit5
# States  : Array complex64 (20, 100, 10, 1) | 156.25 Kb
```


abocquet:
What about using `tree_map` instead ? (https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html#patterns)
Using it we would map a tree of batched tensors of dimension n to a list of trees of tensors of dimensions n-1. This way we fix the fix without restreining on the possibilities.

abocquet (approved):
We'll fix addition in another PR

pierreguilmin:
@gautierronan can you fix the PR description to reflect the changes?

PR #466, opened 2024-02-18
Fix QObjs/lists as input, and arbitrary batching over jump_ops

gautierronan:
This PR achieves two things:
- 51824b8679a9faac690e6170ba32f739d0677a2b: fixes arbitrary array-like objects as input to sesolve/mesolve
- d2ce02dea1f242e01c389108e5f061ee79de4734: fixes the arbitrary batching over jump_ops. Now we can have as many jump_ops batched as we want (or not), and the result will match the structure. 

Quick test (on top of the library tests):
```python
import dynamiqs as dq
import jax.numpy as jnp

N = 10
a = dq.destroy(10)
H = dq.dag(a) @ a
psi0 = dq.coherent(10, 1.0)

# simple batching
jump_ops = [jnp.stack([0.1 * a, 0.2 * a])]
print(result)

# double batching
jump_ops = [
    jnp.stack([0.1 * a, 0.2 * a]),
    jnp.stack([0.01 * dq.dag(a), 0.02 * dq.dag(a)]),
]
print(result)

# cartesian batching=False
jump_ops = [
    jnp.stack([0.1 * a, 0.2 * a]),
    jnp.stack([0.01 * dq.dag(a), 0.02 * dq.dag(a)]),
]
options = dq.Options(cartesian_batching=False)
result = dq.mesolve(H, jump_ops, psi0, tsave, options=options)
print(result)
```
```
==== Result ====
Solver  : Tsit5
States  : Array complex64 (2, 11, 10, 10) | 17.19 Kb
==== Result ====
Solver  : Tsit5
States  : Array complex64 (2, 2, 11, 10, 10) | 34.38 Kb
==== Result ====
Solver  : Tsit5
States  : Array complex64 (2, 11, 10, 10) | 17.19 Kb
```

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -38,11 +38,18 @@ def mesolve(
     tsave = jnp.asarray(tsave)
     exp_ops = jnp.asarray(exp_ops, dtype=cdtype()) if exp_ops is not None else None
 
+    # === check len(jump_ops) > 0
+    if len(jump_ops) == 0:
+        raise ValueError(
+            "At least one jump operator must be provided for `dq.mesolve`. If you don't"
+            ' need jump operators, consider using `dq.sesolve` instead.'
+        )
+
     # === vectorize function
     # we vectorize over H, jump_ops and rho0, all other arguments are not vectorized
     is_batched = (
         H.ndim > 2,
-        jump_ops_ndim > 3,  # todo: this is a temporary fix
+        [jump_op.ndim > 2 for jump_op in jump_ops],
         rho0.ndim > 2,
         False,
         False,
@@ -56,7 +63,16 @@ def mesolve(
     f = compute_vmap(_mesolve, options.cartesian_batching, is_batched, out_axes)
 
     # === apply vectorized function
-    return f(H, jump_ops, rho0, tsave, exp_ops, solver, gradient, options)
+    return f(
+        H,
+        jump_ops,
+        rho0,
+        tsave,
+        exp_ops,
+        solver,
+        gradient,
+        options,
+    )
 
 
 def _mesolve(
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -29,14 +29,21 @@ def mesolve(
     solver: Solver = Tsit5(),
     gradient: Gradient | None = None,
     options: Options = Options(),
-):
+) -> Result:
+    # === convert arguments
+    H = _astimearray(H)
+    jump_ops = [_astimearray(jump_op) for jump_op in jump_ops]
+    psi0 = jnp.asarray(psi0, dtype=cdtype())
+    rho0 = todm(psi0)
+    tsave = jnp.asarray(tsave)
+    exp_ops = jnp.asarray(exp_ops, dtype=cdtype()) if exp_ops is not None else None
+
     # === vectorize function
-    # we vectorize over H, jump_ops and psi0, all other arguments are not vectorized
-    jump_ops_ndim = _astimearray(jump_ops[0]).ndim + 1
+    # we vectorize over H, jump_ops and rho0, all other arguments are not vectorized
     is_batched = (
         H.ndim > 2,
         jump_ops_ndim > 3,  # todo: this is a temporary fix
-        psi0.ndim > 2,
+        rho0.ndim > 2,
         False,
         False,
         False,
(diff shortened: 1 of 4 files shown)


pierreguilmin:
As discussed tgther we also need to fix the JIT:
```python
@jax.jit
def f(x):
    x = jnp.asarray(x)
    return x @ x

x = np.array([2, 3])
f(x)  # ok
x = [1, 2, 3]
f(x)  # ok
x = qt.Qobj(x)
f(x)  # invalid JAX type
```

Dummy solution to start:
```python
def f(x):
    x = jnp.asarray(x)
    return _f(x)

@jax.jit
def _f(x):
    return x @ x

x = np.array([[1, 2], [2, 3]])
f(x)  # ok
x = qt.Qobj(x)
f(x)  # ok
```

Ideally we should make a PR to QuTiP to make `QObj ` valid JAX types, see https://linear.app/dynamiqs/issue/DYN-167/allow-jitted-function-to-take-qutip-qobj-as-parameters.

dynamiqs/mesolve/mesolve.py
```diff
@@ -49,27 +56,19 @@ def mesolve(
...
     options: Options = Options(),
 ) -> Result:
-    # === convert arguments
-    H = _astimearray(H)
-    Ls = [_astimearray(L) for L in jump_ops]
-    y0 = jnp.asarray(psi0, dtype=cdtype())
-    y0 = todm(y0)
-    ts = jnp.asarray(tsave)
-    Es = jnp.asarray(exp_ops, dtype=cdtype()) if exp_ops is not None else None
-
```
pierreguilmin, line 72: I 👍 keeping the arguments name in this section. However, I think we should keep the same API for `mesolve` and `_mesolve`. This makes it very handy if one wants to use the `_sesolve` method without the batching layer (`sesolve` is really made for JAX-beginners). For example, if one wants to batch over a PyTrees of `H` and not a list, which we don't support yet, it's done in two lines using `_sesolve`.

gautierronan: I don't get what you want exactly. We need these argument conversions before the `vmap` to catch the object dimensions.

dynamiqs/mesolve/mesolve.py
```diff
@@ -85,12 +84,33 @@ def mesolve(
...
+    H: TimeArray,
+    jump_ops: list[TimeArray],
+    rho0: ArrayLike,
+    tsave: ArrayLike,
+    exp_ops: list[ArrayLike] | None = None,
+    solver: Solver = Tsit5(),  # noqa: B008
+    gradient: Gradient | None = None,
+    options: Options = Options(),  # noqa: B008
+) -> Result:
     # === vectorize function
```
abocquet, line 109: Why do we need to declare an interemediary function here? Couldn't we merge `mesolve` and `_vmap_mesolve`?

dynamiqs/sesolve/sesolve.py
```diff
@@ -74,6 +73,25 @@ def sesolve(
...
+@partial(jax.jit, static_argnames=('solver', 'gradient', 'options'))
+def _vmap_sesolve(
+    H: TimeArray,
+    psi0: ArrayLike,
+    tsave: ArrayLike,
+    exp_ops: list[ArrayLike] | None = None,
+    solver: Solver = Tsit5(),  # noqa: B008
+    gradient: Gradient | None = None,
+    options: Options = Options(),  # noqa: B008
+) -> Result:
```
abocquet, line 94: Same question

dynamiqs/mesolve/mesolve.py
gautierronan: This is to handle QObjs as inputs: we intentionnally delay the `jit` so there is no tracing on QObjs inputs. See also the comment of @pierreguilmin above.

dynamiqs/mesolve/mesolve.py
pierreguilmin: Sorry I wasn't clear. We should keep the argument conversions in `mesolve`, before the `vmap`, we do need them indeed. I think we should also keep the argument conversions at the beginning of `_mesolve` (and the arguments typing). The main reason is that people who want to do custom stuff (e.g. me 😋) should be able to use `_mesolve` directly. So I'm in favor of repeating the code at the beginning of both functions.

pierreguilmin: It's a minor comment nevermind.

pierreguilmin:
Nice, this is very useful! I've added nit commits on top, and also reverted the change you made because of my comment which wasn't that relevant (I think the point I raised is important, but it's pointless to bother with it at the moment, sorry 🙈).

PR #467, opened 2024-02-19
Fix propagator bug

abocquet:
`ValueError: Zero-dimensional arrays cannot be concatenated.` due to the fact that `np.array(scalar)` returns an array with no shape

### DIFF
--- a/dynamiqs/core/propagator_solver.py
+++ b/dynamiqs/core/propagator_solver.py
@@ -32,7 +32,7 @@ class PropagatorSolver(BaseSolver):
 
         # we use `jnp.asarray(self.t0)` because of the bug fixed here:
         # https://github.com/google/jax/pull/19381 (fixed in jax-0.4.24)
-        delta_ts = jnp.diff(self.ts, prepend=jnp.asarray(self.t0))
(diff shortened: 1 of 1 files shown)


gautierronan (approved):

pierreguilmin (commented):
Why isn't this catched by our test? 🤔

pierreguilmin (commented):
The issue is probably that `self.ts` is not 1D in your case. This will be checked upstream once we reimplement-back `check_time_tensor`. We can close this PR if you agree.
```python
import jax.numpy as jnp

t0 = 0

ts = jnp.array([0, 1])
jnp.diff(ts, prepend=jnp.asarray(t0))  # ok

ts = jnp.array([1])
jnp.diff(ts, prepend=jnp.asarray(t0))  # ok

ts = jnp.array(1)
jnp.diff(ts, prepend=jnp.asarray(t0))  # ValueError: diff requires input that is at least one dimensional
```

pierreguilmin:
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.diff.html
> prepend (array_like, optional) – Values to prepend or append to a along axis prior to performing the difference. Scalar values are expanded to arrays with length 1 in the direction of axis and the shape of the input array in along all other axes. Otherwise the dimension and shape must match a except along axis.

PR #469, opened 2024-02-19
Custom ysave save func

abocquet:


pierreguilmin:
I'm happy to merge this if you're satisfied with the proposed (relatively minor) modifications. The few things I'm unsatisfied (that we can fix later) are:
- [ ] the name "other" is quite bad, "save" is a little too general too, we need smthg better
- [ ] we need to add an example smwh in the documentation
- [ ] the printing of result is not perfect, we don't treat all arrays the same:
```python
# define custom save function
vacuum = dq.coherent(n, 0)
save_fun = lambda y: {'purity': dq.trace(y @ y), 'vacuum_overlap': dq.overlap(y, vacuum)}

# run simulation
result = dq.mesolve(H, jump_ops, psi0, tsave, options=dq.Options(save_fun=save_fun))
print(result)
```
```text
==== Result ====
Solver  : Tsit5
States  : Array complex64 (101, 128, 128) | 12.62 Mb
Other   : {'purity': c64[101], 'vacuum_overlap': f32[101]}
```
Ideally we would print all arrays with their memory (which is super nice to have IMO).

pierreguilmin (approved):

dynamiqs/options.py
```diff
@@ -1,11 +1,21 @@
...
     save_states: bool = True
     verbose: bool = True
     cartesian_batching: bool = True
     t0: Scalar | None = None  # defaults to tsave[0]
+    save_extra: Callable[[Array], PyTree] | None = None
+
+    def __post_init__(self):
+        if self.save_extra is not None:
+            # use `jax.tree_util.Partial` to make `save_extra` a valid Pytree
+            self.save_extra = jax.tree_util.Partial(self.save_extra)
```
gautierronan, line 21: Is it required that it is a PyTree here? I would argue not.

gautierronan (approved):
Just a question below, otherwise LGTM

dynamiqs/options.py
pierreguilmin: Yes, because it's used in jitted code. I get a
```
TypeError: Value <function <lambda> at 0x1235cb5b0> with type <class 'function'> is not a valid JAX type
```
otherwise.

PR #470, opened 2024-02-19
Fix propagator bug with JAX < 0.4.24 due to `jnp.diff`

abocquet:


dynamiqs/core/propagator_solver.py
```diff
@@ -32,7 +32,9 @@ def propagate(y, delta_t):
 
         # we use `jnp.asarray(self.t0)` because of the bug fixed here:
         # https://github.com/google/jax/pull/19381 (fixed in jax-0.4.24)
-        delta_ts = jnp.diff(self.ts, prepend=jnp.asarray(self.t0))
+        # the `.reshape(-1)` covers the case where `self.t0` is a 0-dimensional array,
+        # that can occur when using `vmap`
+        delta_ts = jnp.diff(self.ts, prepend=jnp.asarray(self.t0).reshape(-1))
```
pierreguilmin, line 37:
```suggestion
        # the `.reshape(-1)` covers the case where `self.t0` is a 0-dimensional array
        delta_ts = jnp.diff(self.ts, prepend=jnp.asarray(self.t0).reshape(-1))
```
(occurs also if the user passes a `self.t0` manually.

pierreguilmin (changes requested):

PR #486, opened 2024-03-01
Revert "JIT wigners"

abocquet:
Reverts dynamiqs/dynamiqs#482 to not leave a broken code in main


abocquet:
The old version was also broken, closing

PR #487, opened 2024-03-01
Rroussea/liouvillian

RemiRousseau:
This pull request proposes to add the two funtions einkron and liouvillian to the utils of dynamiqs.
The first allows to do kronecker product of arrays of operator and the second to compute the liouvillian super-operator.


gautierronan:
Thanks for the PR @RemiRousseau ! As mentionned in private by @pierreguilmin, these functions are already in the API, under other names (`dq.tensor`, `dq.slindbladian`) so I'm closing this. Feel free to re-open if needed.

PR #489, opened 2024-03-01
Update the API of TimeArray

gautierronan:
Following discussion with @pierreguilmin, here is an update on the API of `dq.totime` to make it more granular (and easier to document+use, hopefully).

This also fixes a bug with jitting PWC and Modulated TimeArrays, but makes the `vmap` bug come back. However, I know how to fix it (it should be quite easy and clean too), but for now I'm just putting up a warning in the doc. When we've settled on the TimeArray implementation, then I will fix it.

You can review commit by commit.


pierreguilmin:
I'm very happy with this PR, it's much better 🥳 I made a few commits on top (mostly very minor fixes/nit), I'm merging this now (my first merge from a plane 🤯), and we can iterate if you want to modify smthg back.

pierreguilmin (approved):

pierreguilmin (approved):

PR #490, opened 2024-03-03
Add `dq.set_device()` and `dq.set_precision()` utilities

pierreguilmin:
- Move `cdtype()` to `dynamiqs._utils`
- Rename `dynamiqs/utils/array_types.py` to `dynamiqs/utils/jax_utils.py`
- Add `set_device()` and `set_precision()` globa switches


dynamiqs/utils/jax_utils.py
```diff
@@ -95,3 +81,45 @@ def to_qutip(x: ArrayLike, dims: tuple[int, ...] | None = None) -> Qobj | list[Q
         elif isop(x):  # [[3], [3]] or for composite systems [[3, 4], [3, 4]]
             dims = [dims, dims]
         return Qobj(x, dims=dims)
+
+
+def device(backend: Literal['cpu', 'gpu', 'tpu']):
```
gautierronan, line 86: WDYT of using a more explicit name, e.g. `dq.set_device` and `dq.set_precision`? IMO `dq.device` rather refers to an actual device, and not to the action of setting one. It won't be super used functions so I think it's ok that the name is a bit longer.

dynamiqs/utils/jax_utils.py
```diff
@@ -95,3 +81,45 @@ def to_qutip(x: ArrayLike, dims: tuple[int, ...] | None = None) -> Qobj | list[Q
...
+
+    See [JAX documentation on devices](https://jax.readthedocs.io/en/latest/faq.html#faq-data-placement).
+
+    Args:
+        backend _(string 'cpu', 'gpu', or 'tpu')_: Default device.
+    """
+    jax.config.update('jax_default_device', jax.devices(backend)[0])
+
+
+def precision(dtype: jnp.float32 | jnp.float64):
```
gautierronan, line 103: I would suggest `def set_precision(precision: Literal['simple', 'double'])` (similar API to `device`), or `def set_dtype(dtype: jnp.float32 | jnp.float64)`. Here it's somewhere in between the two.

gautierronan (changes requested):
Looks good ! Small suggestions on the API below.

gautierronan (approved):

PR #498, opened 2024-03-09
Move wigner function in doc

gautierronan:
To make this change, I needed to have `__all__ = [..., 'wigner']` inside the `.utils.utils` file (otherwise the Python API table would not render properly inside the doc). But due to a circular import, I also had to move the content of the wigner file inside `.utils.utils`. This kinda makes sense though if we document them together.

If you have another idea I'm also open to change.


pierreguilmin:
Replaced by #505.

PR #502, opened 2024-03-10
Minor fixes (documentation, type hints, useless code)

pierreguilmin:


docs/getting_started/sharp-bits.md
```diff
@@ -53,7 +53,7 @@ Array([[0.+0.j, 1.+0.j],
        [1.+0.j, 0.+0.j]], dtype=complex64)
 ```
 
-### Computing daggers
+### Computing matrix adjoint
```
gautierronan, line 56: I prefer daggers, more explicit IMO, and more of a physicist language! But if you are convinced, we can keep it like that.

gautierronan (approved):
LGTM apart from the nit below.

docs/getting_started/sharp-bits.md
gautierronan: I don't really like the in-between. We can keep `Computing matrix adjoint`, it's not super important.

PR #507, opened 2024-03-11
Move back contact infos as main section

pierreguilmin:


gautierronan (changes requested):
Need to fix also the link at the top of the Contributing section.

pierreguilmin:
Good catch!

gautierronan (approved):

PR #508, opened 2024-03-11
Fix vector_field for Lindbladian calculation

gautierronan:
Not sure whether this type of optimization is taken care of by the JIT. Better safe than sorry.


pierreguilmin (approved):
👏⚡

pierreguilmin:
@gautierronan FYI this was not JIT-optimized, comparing cat CNOT simulations with each modes truncated at 40 (timing hand-made over multiple runs):
- before: 12.8s
- after: 10.7s

Great low-hanging fruit!! 🚀

PR #511, opened 2024-03-14
Add `__str__` for solvers and gradients

bsaoptima:
Implemented feedback


pierreguilmin:
Replaces #509.

pierreguilmin (changes requested):
First contribution 🥳 As discussed together, I think we should target something fairly shorter for `__str__`. I suggest
```python
def __str__(self) -> str:
    return eqx.tree_pformat(self, indent=0).replace('\n', '').replace(',', ', ')
```
e.g. for `dq.solver.Dopri5()`:
```python
>>> import dynamiqs as dq
>>> print(dq.solver.Dopri5())
Dopri5(rtol=1e-06, atol=1e-06, safety_factor=0.9, min_factor=0.2, max_factor=5.0, max_steps=100000)
```

You can add this to the `Solver` and `Gradient` base classes then, no need to modify each class 😉

pierreguilmin:
And then we can probably tweak `Result` to use this instead of just printing the solver and gradient class name (see https://github.com/dynamiqs/dynamiqs/blob/main/dynamiqs/result.py#L75-L78).

dynamiqs/gradient.py
```diff
@@ -21,6 +21,10 @@ def __init__(self):
             [`diffrax.DirectAdjoint`](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.DirectAdjoint)
             option.
         """
+        super().__init__()
```
pierreguilmin, line 24: We don't need this.

pierreguilmin (changes requested):
Nice! Let's
- Move the implementation to a helper method in `dynamiqs/_utils.py`, e.g. called `tree_str_inline()`.
- Adress this comment > https://github.com/dynamiqs/dynamiqs/pull/511#issuecomment-1998710739.

pierreguilmin:
Could we also do the same for options.py, and also print them in results?

bsaoptima:
forgot to do what you said for `options.py` wait for next push.

dynamiqs/_utils.py
```diff
@@ -54,3 +55,7 @@ def cdtype() -> jnp.complex64 | jnp.complex128:
         return jnp.complex128
     else:
         raise ValueError(f'Data type `{dtype.dtype}` is not yet supported.')
+
+
+def tree_str_inline(pytree: PyTree) -> str:
+    return eqx.tree_pformat(pytree, indent=0).replace('\n', '').replace(',', ', ')
```
pierreguilmin, line 61:
```suggestion
def tree_str_inline(x: PyTree) -> str:
    return eqx.tree_pformat(x, indent=0).replace('\n', '').replace(',', ', ')
```

dynamiqs/gradient.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import equinox as eqx
+from _utils import tree_str_inline
```
pierreguilmin, line 4:
```suggestion
from ._utils import tree_str_inline
```

pierreguilmin: Same for the other files, to fix
```
ModuleNotFoundError: No module named '_utils'
```

dynamiqs/sparse.py
pierreguilmin, line 1: I guess you're spoiling a little bit of another PR. 😋

dynamiqs/result.py
```diff
@@ -72,16 +72,15 @@ def extra(self) -> PyTree | None:
...
+            'Solver  ': self.solver,
+            'Gradient': (self.gradient if self.gradient is not None else None),
             'States  ': array_str(self.states),
             'Expects ': array_str(self.expects) if self.expects is not None else None,
             'Extra   ': (
                 eqx.tree_pformat(self.extra) if self.extra is not None else None
             ),
             'Infos   ': self.infos if self.infos is not None else None,
+            'Options ': self.options,
         }
```
pierreguilmin, line 84: It does not really look good 😢
```python
import dynamiqs as dq
import jax.numpy as jnp

n = 16
a = dq.destroy(n)
ad = dq.dag(a)

H = 10 * ad @ a
jump_ops = [jnp.sqrt(0.1) * a]
psi0 = dq.coherent(n, 1.0)
gradient = dq.gradient.Autograd()
result = dq.mesolve(H, jump_ops, psi0, tsave, gradient=gradient)
print(result)
```

```
==== Result ====
Solver  : Tsit5(rtol=f32[], atol=f32[], safety_factor=f32[], min_factor=f32[], max_factor=f32[], max_steps=i32[])
Gradient: Autograd()
States  : Array complex64 (101, 16, 16) | 202.00 Kb
Infos   : 67 steps (67 accepted, 0 rejected)
Options : Options(save_states=bool[], verbose=bool[], cartesian_batching=bool[], t0=None, save_extra=None)
```

I think the reason is that for the JIT every int/bool is converted to a single-element JAX array 😬

For now I propose to keep the old result code, and we can think about a fix in another PR.

pierreguilmin (changes requested):

pierreguilmin (approved):
I've fixed the remaining bits. Thanks @bsaoptima!

pierreguilmin (approved):

PR #516, opened 2024-03-19
Explain choice of numerical scheme in mesolve

gaspardbb:
Solves `DYN-171`


pierreguilmin:
Brilliant 😍 Thanks also to Lev Arcady and Rémi Robin for helping on this, and big up to @gautierronan for putting the final piece.

pierreguilmin:
First contribution from @gaspardbb 🚀

PR #519, opened 2024-03-20
Add gifit function

RemiRousseau:
Added the gifit function in plots/utils.
The gifit function is a decorator to create a gif from any plot function taking states as first input.


pierreguilmin:
Very cool 😍 I think we might organise this functionality a bit differently, because for example some functions can be computed in a batched way on multiple states, for efficiency, (e.g. it's what we do for wigners), which is not possible here. I'll review/improve on top of this PR this weekend, thanks a lot for the contribution though, this is great 👌

RemiRousseau:
I totally agree with you that's it's not computationally efficient for now, but I only thought of separating plotting functions in two -> one creating the data and one doing the plotting. But I was not sure it was the way you guys wanted to go !

gautierronan (approved):
Four months later (🥵), I realized I needed this function quite often.

I've made a few commits with several improvements and nits.

I'd be down to merge this even without the proper batching @pierreguilmin (actually I wasn't even sure of what you had in mind since it's a plotting function). Are you ok with this?

pierreguilmin:
Yes, let me find time to make a quick review during the week and we can merge this. 👍 Thanks a lot @RemiRousseau for the first iteration!

pierreguilmin:
Lots of various minor modifications, follow up ticket > https://linear.app/dynamiqs/issue/DYN-230/improve-dqgifit.

gautierronan (approved):

pierreguilmin:
Huge thanks again @RemiRousseau, sorry for taking so long! 🙇‍♂️

PR #522, opened 2024-03-21
Implements cos and sin for matrices

gaspardbb:
Resolves DYN-10. 
Uses `jax.scipy.linalg.expm`. That's how [Julia does](https://github.com/JuliaLang/julia/blob/09400e4f5c3e31ba528c2dbcf8e1ccef2a66ce3d/stdlib/LinearAlgebra/src/dense.jl#L1100).
Added `sincosm`, which is useful when we need both sine and cosine and avoid twice the computation.

### DIFF
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -4,8 +4,8 @@ from .wigner import *
 __all__ = [
     'dag',
     'powm',
-    'sinm',
     'cosm',
+    'sinm',
     'tracemm',
     'trace',
     'ptrace',
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -4,6 +4,9 @@ from .wigner import *
 __all__ = [
     'dag',
     'mpow',
+    'sincosm',
+    'sinm',
+    'cosm',
     'tracemm',
     'trace',
     'ptrace',
@@ -27,7 +30,4 @@ __all__ = [
     'fidelity',
     'eigenstates',
     'wigner',
-    'sincosm',
-    'sinm',
-    'cosm',
 ]
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -6,7 +6,6 @@ __all__ = [
     'powm',
     'sinm',
     'cosm',
-    'sincosm',
     'tracemm',
     'trace',
     'ptrace',
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -4,9 +4,9 @@ from .wigner import *
 __all__ = [
     'dag',
     'mpow',
-    'sincosm',
     'sinm',
     'cosm',
+    'sincosm',
     'tracemm',
     'trace',
     'ptrace',
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -27,4 +27,7 @@ __all__ = [
     'fidelity',
     'eigenstates',
     'wigner',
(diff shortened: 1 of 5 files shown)


dynamiqs/utils/utils/__init__.py
```diff
@@ -27,4 +27,7 @@
     'fidelity',
     'eigenstates',
     'wigner',
+    'sincosm',
+    'sinm',
+    'cosm',
```
gautierronan, line 32: I would probably put `sincosm` after `cosm` instead of before. Also, I would probably put these 3 functions after `mpow` because they have similar effects.

dynamiqs/utils/utils/general.py
```diff
@@ -35,6 +35,9 @@
     'overlap',
     'fidelity',
     'eigenstates',
+    'sincosm',
+    'sinm',
+    'cosm',
```
gautierronan, line 40: Same comment here.

pierreguilmin:
Nice! I'll review this weekend.

Also note from the Julia src code:
> If `A` is symmetric or Hermitian, its eigendecomposition ([`eigen`](@ref)) is used to
compute the cosine. Otherwise, the cosine is determined by calling [`exp`](@ref).

Should we do this too?

pierreguilmin:
> Also, you need to add these new functions to `mkdocs.yml` so they appear in the documentation.

And in `dynamiqs/utils/utils/__init__.py`. Try using `task docserve` to check the documentation website locally (cf. [CONTRIBUTING.md](https://github.com/dynamiqs/dynamiqs/blob/main/CONTRIBUTING.md)).

pierreguilmin:
I've added a few (very minor) fixes, and adressed @gautierronan's comment on the go. I'm happy to merge once we adress:

> Also note from the Julia src code:

> > If A is symmetric or Hermitian, its [eigen](https://github.com/dynamiqs/dynamiqs/pull/@ref)decomposition (eigen) is used to
compute the cosine. Otherwise, the cosine is determined by calling [exp](https://github.com/dynamiqs/dynamiqs/pull/@ref).

> Should we do this too?

pierreguilmin (approved):

pierreguilmin:
I'm also thinking: in many cases the matrix is real, e.g. $cos(a+a^\dagger)$, and then $e^{-iA}$ is just the complex conjuguate of $e^{iA}$, so very cheap. We should maybe use this.

pierreguilmin:
And nit, we could add a
```python
from jax.scipy.linalg import expm
```
at the top to have more readable functions.

pierreguilmin:
Actually, if you're motivated @gaspardbb it would make sense to also open a PR to directly add these functions to `jax.scipy.linalg`.

pierreguilmin:
Maybe the `sincosm` is a bit too niche and un-needed, wdyt @gautierronan? Do you know a quantum system definition where we compute both cos and sin of the same matrix?

gautierronan:
> Maybe the `sincosm` is a bit too niche and un-needed, wdyt @gautierronan? Do you know a quantum system definition where we compute both cos and sin of the same matrix?

Agreed. It's also quite easy to rewrite if need be.

gaspardbb:
> Should we do this too?

I don't think that's necessary for the moment. I suggest to wait the moment we to implement those functions for matrices with special structure. 

> This maybe motivates returning real matrices when possible in the rest of the code (e.g. for `dq.identity`, `dq.destroy`, etc...). 🤔

Definitely!

pierreguilmin:
Cool, I've open a follow-up issue > https://linear.app/dynamiqs/issue/DYN-187/improve-dqcos-and-dqsin-performance. Let's merge this, thanks a lot @gaspardbb 🎉

pierreguilmin (approved):

pierreguilmin (approved):

PR #524, opened 2024-03-22
Fix bug with save_states=False

gautierronan:
`save_states=False` was bugged.
```python
import dynamiqs as dq
import jax.numpy as jnp

N = 16
alpha = 2.0

psi0 = dq.coherent(N, alpha)
a = dq.destroy(N)
H = dq.dag(a) @ a
options = dq.Options(save_states=False)

result = dq.sesolve(H, psi0, tsave, options=options)
```


pierreguilmin (approved):

pierreguilmin:
Thanks for that, sorry for introducing this bug.

PR #527, opened 2024-03-25
Verify the shape of `ArrayLike` arguments in public API

pierreguilmin:
Here's a proposal for the backend to systematically verify the shape of `ArrayLike` argument in our public API. Having a shared backend allows to write once a clean exception, which always has the same format when the user mistakes a shape. For example:

```python
>>> x = jnp.array([1, 2])
>>> dq.norm(x)
ValueError: Argument `x` must have shape (..., n, 1), (..., 1, n) or (..., n, n), but has shape x.shape=(2,).
```

Implementation: it is quite rudimentary but I don't think we should make a crazy super general code, this is largely good enough to start with IMO. It has the advantage to be very versatile. This works with vectorize, vmap, and the benchmarked impact is negligible (especially, it's 100% free once the code is jitted).

What needs to be done later, and is not adressed by this PR, is to check _agreement_ between shapes, e.g. for `dq.fidelity()` check that `x` and `y` have the same `n`.

I considered using [jaxtyping](https://docs.kidger.site/jaxtyping/) for this but it has several drawbacks: uncommon function signature in the doc, can't define several shapes (e.g. `(…, n, 1)` or `(…, 1, n)` or `(…, n, n)`), need to apply a decorator to all functions of the library for runtime type checking, can't customize easily the error messages (that are a bit _énervés_). I don't think it would be a good solution for us.

### DIFF
--- a/dynamiqs/_checks.py
+++ b/dynamiqs/_checks.py
@@ -59,9 +59,8 @@ def check_shape(
 
 
 def check_times(x: Array, argname: str, allow_empty: bool = False):
-    # check that a time-array is valid (it must be a 1D array sorted in strictly
+    # check that an array of time is valid (it must be a 1D array sorted in strictly
     # ascending order and containing only positive values)
-    # check_shape(x, argname, '(n,)', subs={'n': f'n{argname}'})
 
     if x.ndim != 1:
         raise ValueError(
--- a/dynamiqs/_checks.py
+++ b/dynamiqs/_checks.py
@@ -58,7 +58,7 @@ def check_shape(
     )
 
 
-def check_time_array(x: Array, argname: str, allow_empty: bool = False):
+def check_times(x: Array, argname: str, allow_empty: bool = False):
     # check that a time-array is valid (it must be a 1D array sorted in strictly
     # ascending order and containing only positive values)
     # check_shape(x, argname, '(n,)', subs={'n': f'n{argname}'})
--- a/dynamiqs/_checks.py
+++ b/dynamiqs/_checks.py
@@ -6,26 +6,28 @@ from jax import Array
 _is_perfect_square = lambda n: int(n**0.5) ** 2 == n
 
 
+_cases = {
+    '(..., n, 1)': lambda x: x.ndim >= 2 and x.shape[-1] == 1,
+    '(..., 1, n)': lambda x: x.ndim >= 2 and x.shape[-2] == 1,
+    '(..., n, n)': lambda x: x.ndim >= 2 and x.shape[-2] == x.shape[-1],
+    '(N, ..., n, n)': lambda x: x.ndim >= 3 and x.shape[-2] == x.shape[-1],
+    '(..., n)': lambda x: x.ndim >= 1,
+    '(n, 1)': lambda x: x.ndim == 2 and x.shape[-1] == 1,
+    '(1, n)': lambda x: x.ndim == 2 and x.shape[-2] == 1,
+    '(n, n)': lambda x: x.ndim == 2 and x.shape[-2] == x.shape[-1],
+    '(n,)': lambda x: x.ndim == 1,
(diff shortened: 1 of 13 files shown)


pierreguilmin:
I'll move the dict definition outside of the `has_shape` function.

dynamiqs/_checks.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import annotations
+
+import jax.numpy as jnp
+from jax import Array
+
+_is_perfect_square = lambda n: int(n**0.5) ** 2 == n
+
+
+def has_shape(x: Array, shape: str) -> bool:
+    cases = {
```
gautierronan, line 10: Missing (maybe intentional):
 - (N, 1, n) for batched bras
 - (..., n^2, n^2) for vectorized Liouvillians
 - (..., 1, n^2) for daggered vectorized density matrices

dynamiqs/_checks.py
```diff
@@ -0,0 +1,71 @@
...
+            return
+
+    if len(shapes) == 1:
+        shapes_str = shapes[0]
+    else:
+        shapes_str = ', '.join(shapes[:-1]) + ' or ' + shapes[-1]
+
+    if subs is not None:
+        for k, v in subs.items():
+            shapes_str = shapes_str.replace(k, v)
```
gautierronan, line 47: I didn't catch what this `subs` does. It's substitions for the shape?

dynamiqs/_checks.py
```diff
@@ -0,0 +1,71 @@
...
+
+    raise ValueError(
+        f'Argument `{argname}` must have shape {shapes_str}, but has shape'
+        f' {argname}.shape={x.shape}.'
+    )
+
+
+def check_time_array(x: Array, argname: str, allow_empty: bool = False):
```
gautierronan, line 55: Now that we have an object called TimeArray, this is properly not a well suited name. Maybe `check_times` instead ? Or `check_array_of_times` ?

gautierronan (changes requested):
This is nice and very useful ! A few minor comments below, otherwise we can go forward with it :) 

Regarding checking common shapes, I guess a simple local check with an `if` statement should be enough. No need to make some big artillery for that.

dynamiqs/_checks.py
pierreguilmin: Yes it was intentional, I just added checks that we actually use in the codebase. We can add new checks as we go along when we need them.

dynamiqs/_checks.py
pierreguilmin: This is needed because we have a specification for shapes which does not work for arbitrary symbols, but sometimes the doc uses more complex symbols, and the error message should as well. For example in `dq.expect()`:
```python
# O _(array_like of shape (nO?, n, n))_: Arbitrary operator or list of _nO_ operators.
check_shape(O, 'O', '(?, n, n)', subs={'?': 'nO?'})
```
This says « verify the shape of O with specification '(?, n, n)', but in the error message replace '?' by 'nO?' ».

It's just used in `dq.expect()` for now, but will be used a lot by `sesolve`/`mesolve`, e.g.:
```python
# rho0 (array-like of shape (nrho0?, n, 1) or (nrho0?, n, n)) – Initial state.
check_shape(rho0, 'rho0', '(?, n, 1)', '(?, n, n)', subs={'?': 'nrho0?'})
```

I'll add a comment to make this more clear.

dynamiqs/_checks.py
pierreguilmin: Haha yes true, I like `check_times` 👍

pierreguilmin:
> Regarding checking common shapes, I guess a simple local check with an `if` statement should be enough. No need to make some big artillery for that.

What do you mean?

gautierronan (approved):
LGTM

Issue #528, opened 2024-03-27
Wigner function with arbitrary xvec, yvec


HarshBabla99:
Hello, 
Firstly, thanks so much for creating this wonderful package and making it available for everyone. It is so well written and a pleasure to use!
Could you consider a slight edit to the `dq.wigner(...)` method? Currently the method takes in two values `xmax` & `ymax`, and creates a symmetric wigner function - extending equally far vertically and horizontally. Would you be okay with allowing for arbitrary `xvec` and `yvec` values as inputs to the method? Going through the source code, it should be a straightforward change. 

Here are a few reasons to motivate the change:
1. **Generality**: It's nice to have more flexibility and plot Wigner functions for arbitrary regions in phase space. This is especially useful for zooming into interesting regions. 
2. **Backwards compatibility with QuTiP** 
3. **Getting the Wigner values for specific values**: For many Bosonic codes key properties such as the expectation values of Pauli operators eg. $\langle X \rangle$, $\langle Z \rangle$, $\langle XX \rangle$, etc. can be minimally extracted by evaluating the Wigner function at specific displacements in phase space. (See: Pg 7 onwards in the Appendix of [https://arxiv.org/pdf/2311.04423.pdf](url)). This would be very inconvenient to do with the current dynamiqs' implementation. 

Of course, all these problems would be solved if one directly calls `_wigner_clenshaw()` or `_wigner_fft`. But this is quite a clumsy hack, and it is perhaps unsavory to use private methods. Moreover, the nice jit and auto-vec optimization that you have implemented on the wrapper would be lost. 

I hope you consider implementing this change. If you decide to implement it, I would be happy to contribute.

pierreguilmin:
Relates to https://linear.app/dynamiqs/issue/DYN-150/change-wigner-parameters-from-scalars-to-tensors.

abocquet:
Hello @HarshBabla99 
I had the code around so I pushed to #529. Can you test the branch `abocquet/wigner-arbitrary` and tell me if this is what you had in mind?

pierreguilmin:
[...]

PR #529, opened 2024-03-29
Improve `dq.wigner()` (remove fft method, support arbitrary `xvec` and `yvec`)

abocquet:
Closes https://github.com/dynamiqs/dynamiqs/issues/528.

Closes DYN-150.

I chose to set `xvec` and `yvec` as static arguments because I assume they will not change much and this avoids one conditional branching

### DIFF
--- a/dynamiqs/plots/plots_wigner.py
+++ b/dynamiqs/plots/plots_wigner.py
@@ -364,10 +364,10 @@ def plot_wigner_gif(
 def check_state_device_type(state: ArrayLike) -> ArrayLike:
     if next(iter(state.devices())).platform == 'gpu' and state.dtype == jnp.complex128:
         logging.warning(
-            'Wigner function is not supported yet for f64 on GPU. '
-            'Dynamiqs will copy the state to the CPU to compute the wigner function. '
-            'Performance penalty is expected. If this is a problem for you, open an '
-            'issue at https://github.com/dynamiqs/dynamiqs/issues'
+            'Wigner-related functions are not yet supported for double precision (f64) '
+            'on GPU. The `state` array will be copied to the CPU to compute the wigner '
+            'function. Performance penalty is expected. If this is a problem for you, '
+            'open an issue at https://github.com/dynamiqs/dynamiqs/issues'
         )
         state = jax.device_put(state, jax.devices(backend='cpu')[0])
     return state
--- a/dynamiqs/plots/plots_wigner.py
+++ b/dynamiqs/plots/plots_wigner.py
@@ -362,12 +362,12 @@ def plot_wigner_gif(
 
 
 def check_state_device_type(state: ArrayLike) -> ArrayLike:
-    if next(iter(state.devices())).platform == "gpu" and state.dtype == jnp.complex128:
+    if next(iter(state.devices())).platform == 'gpu' and state.dtype == jnp.complex128:
         logging.warning(
-            "Wigner function is not supported yet for f64 on GPU. "
-            "Dynamiqs will copy the state to the CPU to compute the wigner function. "
-            "Performance penalty is expected. If this is a problem for you, open an "
-            "issue at https://github.com/dynamiqs/dynamiqs/issues"
+            'Wigner function is not supported yet for f64 on GPU. '
+            'Dynamiqs will copy the state to the CPU to compute the wigner function. '
+            'Performance penalty is expected. If this is a problem for you, open an '
+            'issue at https://github.com/dynamiqs/dynamiqs/issues'
         )
-        state = jax.device_put(state, jax.devices(backend="cpu")[0])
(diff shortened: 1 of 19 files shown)


gautierronan (changes requested):
LGTM. I guess we should propagate these changes to related `plot_...` functions as well.

pierreguilmin:
Great, I'll review this evening or tmo!

> LGTM. I guess we should propagate these changes to related plot_... functions as well.
I don't think it makes sense, `xvec` and `yvec` can be arbitrary here (not necessary evenly-spaced, nor ordered).

pierreguilmin (changes requested):
We need to
- add `xvec`, `yvec` to the documentation
- fix the typing: `ArrayLike | None`
- check their shape if they're not `None`

@gautierronan are both methods (clenshaw and fft) working whatever `xvec` and `yvec`? (not evenly spaced, not ordered, repeated elements, ...) If not we should document in a warning what's possible for `xvec`/`yvec`.

gautierronan:
> @gautierronan are both methods (clenshaw and fft) working whatever xvec and yvec? (not evenly spaced, not ordered, repeated elements, ...) If not we should document what's suitable for xvec/yvec.

I think we should remove the `fft` method altogether. It makes sense in qutip because it's much faster, but here clenshaw is jitted so it's really comparable in runtime, and much more flexible. I don't see any reason to use `fft`, and it's annoying to maintain.

pierreguilmin (changes requested):
A few more things to fix:
- Remove unused `_fock_to_position`
- Check shape for xvec and yvec
- Fix return documentation
- Fix CI (merge with main, correct line length)

A few extras:
- @gautierronan we discussed this a while ago, I think it was related to the fft method, can we set the default xmax/ymax to e.g. 6.0 instead of this weird factor?
- Should we take in a single array of complex displacements rather than `xvec`, `yvec`? This is even more general, because currently we force to compute on a grid (that can be irregular). @gautierronan would the clenshaw algorithm work in the case of arbitrary displacements?
- Ideally we should support arbitrary batching on `xvec`/`yvec` or on the displacement array (we can do this in another PR).

pierreguilmin:
Also (for another PR), the JIT is probably disallowing the function to work correctly with QuTiP QObj, we should either fix this generally https://linear.app/dynamiqs/issue/DYN-167/allow-jitted-function-to-take-qutip-qobj-as-parameters or replay the trick we used for mesolve/sesolve (separate arrays conversion from jitted backend).

dynamiqs/plots/plots_wigner.py
```diff
@@ -354,3 +359,15 @@ def plot_wigner_gif(
...
             shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+def check_state_device_type(state: ArrayLike) -> ArrayLike:
+    if next(iter(state.devices())).platform == 'gpu' and state.dtype == jnp.complex128:
+        logging.warning(
+            'Wigner function is not supported yet for f64 on GPU. '
+            'Dynamiqs will copy the state to the CPU to compute the wigner function. '
+            'Performance penalty is expected. If this is a problem for you, open an '
+            'issue at https://github.com/dynamiqs/dynamiqs/issues'
```
gautierronan, line 370:
```suggestion
            'Wigner-related functions are not yet supported for double precision (f64) on GPU. '
            'The `state` array will be copied to CPU to compute the wigner function. '
            'Performance penalty is expected. If this is a problem for you, open an '
            'issue at https://github.com/dynamiqs/dynamiqs/issues'
```

gautierronan (approved):
LGTM.


> @gautierronan we discussed this a while ago, I think it was related to the fft method, can we set the default xmax/ymax to e.g. 6.0 instead of this weird factor?

Yes we can if we remove `fft` indeed! BTW, this wierd factor is just 2pi :D 

> Should we take in a single array of complex displacements rather than xvec, yvec? This is even more general, because currently we force to compute on a grid (that can be irregular). @gautierronan would the clenshaw algorithm work in the case of arbitrary displacements?

In principle it should work, but it requires a bit of refactoring and rethinking. I'd be in favor of keeping it like that for now, until someone actually asks for it. Especially since people can always compute the expectation value of the displaced parity if needed.

> Ideally we should support arbitrary batching on xvec/yvec or on the displacement array (we can do this in another PR).

Again, a bit niche I think, but shouldn't be too hard to do.

pierreguilmin:
I have a few modifications to propose, but I won't have time before this weekend sorry.

abocquet:
Unless something is really critical let's merge and change later. Or tell me what bothers you and I'll change

pierreguilmin:
Main thing is why did you move the `states = check_state_device_type(states)` check to plot functions?

pierreguilmin (approved):

PR #532, opened 2024-03-29
Fix JIT sesolve/mesolve broken by runtime check

pierreguilmin:
The check for `tsave` being in strictly ascending order was breaking the JIT. This PR uses [`equinox.error_if`](https://docs.kidger.site/equinox/api/errors/) to implement a JIT-compatible check.

This is exactly what's done in Diffrax too, see e.g. https://github.com/patrick-kidger/diffrax/blob/d97ba2006426836b9b57dfed8d2c24c7373567e0/diffrax/_integrate.py#L796.

MWE:
```python
import dynamiqs as dq
import jax

n = 16
a = dq.destroy(n)
ad = dq.dag(a)

H = 10 * ad @ a
psi0 = dq.coherent(n, 1.0)
tsave = [0, 1, -1]

@jax.jit
def foo():
    return dq.sesolve(H, psi0, tsave)

foo()
```
```
XlaRuntimeError: INTERNAL: Generated function failed: CpuCallback error: EqxRuntimeError: Argument tsave must be sorted in strictly ascending order.
```

### DIFF
--- a/dynamiqs/_checks.py
+++ b/dynamiqs/_checks.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import jax.numpy as jnp
+import equinox as eqx
 from jax import Array
 
 _is_perfect_square = lambda n: int(n**0.5) ** 2 == n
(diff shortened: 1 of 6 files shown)


pierreguilmin:
CI is broken bc of #531.

gautierronan (approved):
Nice catch! Solution looks good.

PR #533, opened 2024-03-29
Fix batching for PWC, Modulated and Summed TimeArrays

gautierronan:
This PR fixes batching for TimeArrays all around. We still cannot handle weird batchings (e.g. multiple batching within the same TimeArray) but it shouldn't be very hard to add: it's rather a question of whether we want it.

I needed it so I decided to implement it without waiting for the `args` refactorization of TimeArrays. I hope it's fine, and a similar structure might be needed anyway. 

Nits but can remove:
 - 72fe473c5778fe67792bc4d7aa2677eeced8b7b8 removes comments in args checks (thought it was self-explanatory already)
 - da7f2e6db1468aef0c89bb20ad7f1e6951886b17 renames `self.x` to `self.array` in `ConstantTimeArray`. I think it's more aligned with other notations inside TimeArrays this way.

Closes DYN-192.

### DIFF
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -6,7 +6,14 @@ from jaxtyping import ArrayLike
 
 from .._utils import cdtype, obj_type_str
 from ..solver import Solver
-from ..time_array import ConstantTimeArray, TimeArray
+from ..time_array import (
+    CallableTimeArray,
+    ConstantTimeArray,
+    ModulatedTimeArray,
+    PWCTimeArray,
+    SummedTimeArray,
+    TimeArray,
+)
 from .abstract_solver import AbstractSolver
 
 
@@ -67,10 +74,30 @@ def compute_vmap(
                     in_axes = jax.tree_util.tree_map(lambda _: None, leaves)
                     in_axes[n - 1 - i] = 0
                     in_axes = jax.tree_util.tree_unflatten(treedef, in_axes)
-                    f = jax.vmap(f, in_axes=in_axes, out_axes=0)
+                    f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
         else:
             # map over all batched dimensions at once
             in_axes = jax.tree_util.tree_map(lambda x: 0 if x else None, is_batched)
             f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
 
     return f
+
+
+def compute_timearray_batching(tarray: TimeArray) -> TimeArray:
+    # This function finds all batched arrays within a given TimeArray.
+    # To do so, it goes down the PyTree and identifies batched fields depending
+    # on the type of TimeArray.
+    if isinstance(tarray, SummedTimeArray):
+        return SummedTimeArray(
+            [compute_timearray_batching(arr) for arr in tarray.timearrays]
+        )
+    elif isinstance(tarray, ConstantTimeArray):
+        return ConstantTimeArray(tarray.x.ndim > 2)
+    elif isinstance(tarray, PWCTimeArray):
+        return PWCTimeArray(False, tarray.values.ndim > 1, False)
+    elif isinstance(tarray, ModulatedTimeArray):
+        return ModulatedTimeArray(False, False, (arg.ndim > 0 for arg in tarray.args))
+    elif isinstance(tarray, CallableTimeArray):
+        return CallableTimeArray(False, (arg.ndim > 0 for arg in tarray.args))
+    else:
+        raise TypeError(f'Unsupported TimeArray type: {type(tarray).__name__}')
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -94,8 +94,10 @@ def is_timearray_batched(tarray: TimeArray) -> TimeArray:
     elif isinstance(tarray, PWCTimeArray):
         return PWCTimeArray(False, tarray.values.ndim > 1, False)
     elif isinstance(tarray, ModulatedTimeArray):
-        return ModulatedTimeArray(False, False, (arg.ndim > 0 for arg in tarray.args))
+        return ModulatedTimeArray(
+            False, False, tuple(arg.ndim > 0 for arg in tarray.args)
+        )
     elif isinstance(tarray, CallableTimeArray):
-        return CallableTimeArray(False, (arg.ndim > 0 for arg in tarray.args))
+        return CallableTimeArray(False, tuple(arg.ndim > 0 for arg in tarray.args))
     else:
         raise TypeError(f'Unsupported TimeArray type: {type(tarray).__name__}')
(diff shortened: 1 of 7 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -182,25 +189,18 @@ def _check_mesolve_args(
...
-    # === check H shape
     check_shape(H, 'H', '(?, n, n)', subs={'?': 'nH?'})
+    check_shape(rho0, 'rho0', '(?, n, 1)', '(?, n, n)', subs={'?': 'nrho0?'})
+    check_times(tsave, 'tsave')
+
+    if exp_ops is not None:
+        check_shape(exp_ops, 'exp_ops', '(N, n, n)', subs={'N': 'nE'})
+
+    for i, L in enumerate(jump_ops):
+        check_shape(L, f'jump_ops[{i}]', '(?, n, n)', subs={'?': 'nL?'})
```
pierreguilmin, line 200: Nit: it'd be better to order this function as before, by arguments order as in the function signature (for code readability + if the user made multiple mistakes, he'll see them in order and fix the function call left to right).

dynamiqs/core/_utils.py
```diff
@@ -67,10 +74,30 @@ def compute_vmap(
                     in_axes = jax.tree_util.tree_map(lambda _: None, leaves)
                     in_axes[n - 1 - i] = 0
                     in_axes = jax.tree_util.tree_unflatten(treedef, in_axes)
-                    f = jax.vmap(f, in_axes=in_axes, out_axes=0)
+                    f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
```
pierreguilmin, line 77: Is this correct? 🤔

tests/test_time_array.py
```diff
@@ -70,6 +74,23 @@ def test_radd(self):
...
+        # constant time array
+        H_cte = jnp.stack([H0, 2 * H0])
+
+        # test sesolve output shape
+        result = sesolve(H_cte, psi0, times)
+        assert result.states.shape == (2, 11, 4, 1)
+
+        result = sesolve(H0 + H_cte, psi0, times)
+        assert result.states.shape == (2, 11, 4, 1)
+
```
pierreguilmin, line 93: We should not use `sesolve` to test `TimeArray`, the unit tests of a file should not use other very high level function (every time we will break sesolve, we will also break these tests). If you want to add tests for this PR we should add them smwh in sesolve tests.

pierreguilmin (changes requested):
@gautierronan, is this PR to support **sum** of batched time arrays? Why not just doing as before and having the `ndim` of `SummedTimeArray` returning the maximum dimension?

dynamiqs/sesolve/sesolve.py
```diff
@@ -106,9 +111,20 @@ def _vmap_sesolve(
 ) -> SEResult:
     # === vectorize function
     # we vectorize over H and psi0, all other arguments are not vectorized
-    is_batched = (H.ndim > 2, psi0.ndim > 2, False, False, False, False, False)
+    is_batched = (
+        is_timearray_batched(H),
```
pierreguilmin, line 115: If we choose to do this we need to fix the type of `is_batched: list[bool | list[bool]]` in `compute_vmap`. But let's maybe wait to see if there's another solution.

pierreguilmin:
I'm confused by the new function, what's the resulting batching if
1. I sum time arrays with incompatible broadcast shape: `(n, n)` + `(a, n, n)` + `(b, n, n)`
2. I sum time arrays with compatible but different broadcast shape: `(n, n)` + `(1, a, n, n)` + `(b, 1, n, n)`

gautierronan:
> @gautierronan, is this PR to support sum of batched time arrays? Why not just doing as before and having the ndim of SummedTimeArray returning the maximum dimension?

No we need more than just the overall dimension, we need to know where the batching happens (e.g. within the `PWCTimeArray.values`) otherwise `vmap` will not know itself and will raise an error.

So this PR adds support for batching for PWC, Modulated and Sums of TimeArrays, as stated, even if there is a single batching dimension over the full TimeArray.

> I'm confused by the new function, what's the resulting batching if
> I sum time arrays with incompatible broadcast shape: (n, n) + (a, n, n) + (b, n, n)

No these two arrays are not broadcastable, so even trying to compute the time array shape will fail.

> I sum time arrays with compatible but different broadcast shape: (n, n) + (1, a, n, n) + (b, 1, n, n)

I didn't try it, but I think this should work with this PR.

dynamiqs/core/_utils.py
gautierronan: Yes. Before it was batching the result over all batchable output arrays. So even `result.times` was batched, which doesn't make sense. We can change `out_axes` if we want to batch over different outputs, but the syntax is correct.

tests/test_time_array.py
gautierronan: Right, but `sesolve` tests call `closed_system` stuff right now, and this wouldn't fit in. How should I write that?

dynamiqs/sesolve/sesolve.py
gautierronan: What's wrong with this solution? :D

dynamiqs/mesolve/mesolve.py
```diff
@@ -124,18 +129,20 @@ def _vmap_mesolve(
...
+        [is_timearray_batched(jump_op) for jump_op in jump_ops],
         rho0.ndim > 2,
         False,
         False,
         False,
         False,
         False,
     )
+
     # the result is vectorized over `saved`
```
pierreguilmin, line 142:
```suggestion
    # the result is vectorized over `_saved` and `infos`
```
(same in sesolve)

tests/test_time_array.py
pierreguilmin: For example in `tests/sesolve/test_sesolve_batching.py`. But it's a bit had-oc, basically this means that our vectorizing strategy is too tight with our solvers API, we should be able to test them independently. Doesn't matter much for now, just move this test smwh in `tests/sesolve`. 😉

dynamiqs/core/_utils.py
pierreguilmin: Right, so this fixes a notable bug 🙏

dynamiqs/sesolve/sesolve.py
pierreguilmin: Mmmm it's just that how things are batched becomes quite hard to understand, we're in between supporting lists or supporting full PyTrees of H/Ls/rhos, we're "kind of supporting more advanced batching than just 1d per argument but with this, that and this constraint". I think the two extremes (1d list only, full PyTrees) are easier to write, easier to read, and easier to explain. Cf https://linear.app/dynamiqs/issue/DYN-157/support-batching-on-pytrees. That's why I'm not 100% satisfied with the solution, but I agree it's good enough for now. 👍

pierreguilmin:
> > I sum time arrays with compatible but different broadcast shape: (n, n) + (1, a, n, n) + (b, 1, n, n)

> I didn't try it, but I think this should work with this PR.


What would work? Would it return a state `(a, b, nt, n, n)`? Or `(a * b, nt, n, n)`?

pierreguilmin (approved):
I'm OK to merge this if you need it, but I think we need to refactor properly both TimeArray and batching someday.

pierreguilmin:
Ho, and verry sorry for the late reply. I wanted to look at this more in depth to have ideas of what's the future ahead of us (what would be another solution), but couldn't find the time. Thanks a lot for the fix.

pierreguilmin (approved):

PR #367, opened 2023-12-03
Von-Neumann entropy

baptistehoyau:
*I wrote the Von-Neumann entropy function following what was done on qutip

Closes DYN-129.

PR #534, opened 2024-03-31
Add Von Neumann entropy `dq.entropy_vn()`

pierreguilmin:
Replaces #367. Thanks @baptistehoyau for starting this PR back in the PyTorch days!

Closes DYN-129.

### DIFF
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -27,6 +27,7 @@ __all__ = [
     'braket',
     'overlap',
     'fidelity',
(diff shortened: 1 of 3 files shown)


pierreguilmin:
> Also, I may prefer the name dq.entropy instead of dq.entropy_vn. It's a bit closer to the other function names that we already have. What do you think?

@gautierronan concerning this comment, I'm in favour of keeping smthg that refers to the Von Neumann entropy, as there are lots of different definitions.

gautierronan (approved):

PR #535, opened 2024-04-02
Support rectangular matrices in `dq.dag()`

abocquet:
Is is legitimate to compute the hermitian conjugate of arbitrary shaped matrices, so we should remove the check


gautierronan (approved):

pierreguilmin (approved):

pierreguilmin:
QuTiP has a hidden dependency on `scipy < 1.13`. Let me hot fix this.

pierreguilmin (approved):

PR #540, opened 2024-04-05
Remove `dq.eigenstates()` (for now)

pierreguilmin:
As discussed with @gautierronan and @RemiRousseau this functions has a bunch of issues for now, we'll work on a better version

- Somewhat ill-defined for super-operators (only returns right eigenvectors)
- Documentation is too loose
- Not GPU-compatible
- Bug in the current implementation (due to mismatch btw `jnp.linalg.eig()` and `np.linalg.eig()`)

### DIFF
--- a/dynamiqs/utils/utils/__init__.py
+++ b/dynamiqs/utils/utils/__init__.py
@@ -28,6 +28,5 @@ __all__ = [
     'overlap',
     'fidelity',
     'entropy_vn',
-    'eigenstates',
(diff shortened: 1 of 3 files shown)


pierreguilmin:
Re-implementing this is tracked in https://linear.app/dynamiqs/issue/DYN-194/implement-an-eigenstates-like-function. We could start with Hermitian matrices first, it's much easier to have smthg rigorous.

PR #544, opened 2024-04-05
Add Bosh3 solver from diffrax

gautierronan:
Just adding a low-order Runge Kutta solver (`Bosh3`) from diffrax to dynamiqs. Wanted to try it out on a specific problem (no RWA, fast dynamics) to see if it would improve the solving. We can merge already, but can also wait to see if tests are concluant.


pierreguilmin:
Let's see if we find some use case where it's useful before adding it 😉

gautierronan:
> Let's see if we find some use case where it's useful before adding it 😉

Turns out it was not useful, so I'm closing. Feel free to reopen if needed.

PR #548, opened 2024-04-11
Add `dq.smesolve()`

pierreguilmin:
Closes DYN-164.

First version April 11, 2024, updated September 2, 2024. Changed target branch to `dev-smesolve`.

This is quite a big one, I advise reading commit by commit. I did my best to split the logic between saving, solver logic and API. This solver has not reached it's final version, consider this is a first working POC on which we can iterate.

Screenshot 2024-04-11 at 22 59 48

## Example API usage
```python
H = dq.zero(2)
psi0 = dq.unit(dq.basis(2, 0) + dq.basis(2, 1))
jump_ops = [1.0 * dq.sigmaz()]
etas = [1.0]
tsave = np.linspace(0, 1.0, 101)
key = jax.random.PRNGKey(42)
ntrajs = 6
keys = jax.random.split(key, ntrajs)

result = dq.smesolve(H, jump_ops, etas, psi0, tsave, keys)

print(result.states.shape)  # (ntrajs, ntsave, n, n)
print(result.measurements.shape)  # (ntrajs, nLm, ntsave-1)
```

## Example 1: monitored qubit
First example to convince the SME-practionner that the solver is correct. Qubit starting from $\ket+$ with $H=0$ and monitored $L=\sigma_z$ (we expect diffusive projection onto one of $\sigma_z$ eigenvalues).

qubit-trajectories

<details>
  <summary>Code</summary>

  ```python
  # imports
  import jax
  import jax.numpy as jnp
  import numpy as np
  
  import dynamiqs as dq
  from dynamiqs.plots.utils import mplstyle
  
  mplstyle()
  
  # simulation
  H = dq.zero(2)
  psi0 = dq.unit(dq.basis(2, 0) + dq.basis(2, 1))
  jump_ops = [1.0 * dq.sigmaz()]
  etas = [1.0]
  tsave = np.linspace(0, 1.0, 101)
  key = jax.random.PRNGKey(42)
  ntrajs = 6
  keys = jax.random.split(key, ntrajs)
  
  result = dq.smesolve(H, jump_ops, etas, psi0, tsave, keys)
  
  print(result.states.shape)  # (ntrajs, ntsave, n, n)
  print(result.measurements.shape)  # (ntrajs, nLm, ntsave-1)
  
  # plot
  fig, axs = dq.gridplot(3, 3, w=6, h=3, sharex=True)
  ax0, ax1, ax2 = list(axs)
  
  for states in result.states:
      ax0.plot(tsave, dq.expect(dq.sigmaz(), states).real, lw=1.5)
      ax0.axhline(-1, ls='--', lw=1.0, color='gray')
      ax0.axhline(1, ls='--', lw=1.0, color='gray')
      ax0.set(
          title=r'Time evolution of $\sigma_z$ expectation value',
          xlabel=r'$t$',
          ylabel=r'$\mathrm{Tr}[\sigma_z\rho_t]$',
          ylim=(-1.1, 1.1),
      )
  
  for It in result.measurements[:, 0]:
      integrated_signal = jnp.cumsum(It)
      ax1.axhline(0, ls='--', lw=1.0, color='gray')
      ax1.plot(tsave[1:], integrated_signal, lw=1.5)
      ax1.set(
          title=r'Time integrated signal',
          xlabel=r'$t$',
          ylabel=r'$\int_0^t \mathrm{d}Y_u$',
          ylim=(-4.5, 4.5),
      )
  
      ax2.plot(tsave[:-1], It, lw=1.5)
      delta_t = tsave[1] - tsave[0]
      ax2.set(
          title=rf'Instantaneous signal (binned on $\Delta t={delta_t}$)',
          xlabel=r'$t$',
          ylabel=r'$I_t$',
          ylim=(-0.4, 0.4),
      )
  
  fig.savefig('qubit-sme.png', dpi=300, bbox_inches='tight')
  ```

</details>

## Example 2: monitored cavity

Cavity starting from $\ket\alpha$ with $H=\Delta a^\dagger a$ and monitored $L_X = \sqrt\kappa a$ and $L_P = \sqrt\kappa (-ia)$ (we expect no measurement backaction and average signal trajectories following the state expectation values)
cavity-sme

<details>
  <summary>Code</summary>

```python
# imports
import jax
import jax.numpy as jnp
import numpy as np

import dynamiqs as dq
from dynamiqs.plots.utils import mplstyle

mplstyle()

# simulation
n = 16
a = dq.destroy(n)
H = 10 * dq.dag(a) @ a
psi0 = dq.coherent(n, 2.0)
jump_ops = [a, -1j * a]
etas = [1.0, 1.0]
tsave = np.linspace(0, 1.0, 101)
key = jax.random.PRNGKey(42)
ntrajs = 100
keys = jax.random.split(key, ntrajs)

result = dq.smesolve(H, jump_ops, etas, psi0, tsave, keys)

# plot
fig, axs = dq.gridplot(6, 3, w=6, h=3, sharex=True)
ax0, ax1, ax2, ax3, ax4, ax5 = list(axs)
delta_t = tsave[1] - tsave[0]

for states in result.states:
    ax0.plot(tsave, dq.expect(dq.position(n), states).real, lw=1.5)
    ax0.set(
        title=r'Time evolution of $X$ expectation value',
        xlabel=r'$t$',
        ylabel=r'$\mathrm{Tr}[X\rho_t]$',
        ylim=(-2.5, 2.5),
    )

    ax1.plot(tsave, dq.expect(dq.momentum(n), states).real, lw=1.5)
    ax1.set(
        title=r'Time evolution of $P$ expectation value',
        xlabel=r'$t$',
        ylabel=r'$\mathrm{Tr}[P\rho_t]$',
        ylim=(-2.5, 2.5),
    )

ItXs = result.measurements[:, 0] / delta_t / 2
ax2.plot(tsave[1:], jnp.mean(ItXs, axis=0))
ax2.set(
    title=r'$X$ quadrature signal averaged over all trajectories',
    xlabel=r'$t$',
    ylabel=r'$\mathbb{E}[I_t^X]$',
    ylim=(-2.5, 2.5),
)

ItPs = result.measurements[:, 1] / delta_t / 2
ax3.plot(tsave[1:], jnp.mean(ItPs, axis=0))
ax3.set(
    title=r'$P$ quadrature signal averaged over all trajectories',
    xlabel=r'$t$',
    ylabel=r'$\mathbb{E}[I_t^P]$',
    ylim=(-2.5, 2.5),
)

for It in ItXs[:3]:
    ax4.plot(tsave[:-1], It, lw=1.5)
    ax4.set(
        title=rf'Instantaneous $X$ quadrature signal for 3 trajectories (binned on $\Delta t={delta_t}$)',
        xlabel=r'$t$',
        ylabel=r'$I_t^X$',
        ylim=(-15, 15),
    )

for It in ItPs[:3]:
    ax5.plot(tsave[:-1], It, lw=1.5)
    ax5.set(
        title=rf'Instantaneous $P$ quadrature signal for 3 trajectories (binned on $\Delta t={delta_t}$)',
        xlabel=r'$t$',
        ylabel=r'$I_t^P$',
        ylim=(-15, 15),
    )

fig.savefig('cavity-sme.png', dpi=300, bbox_inches='tight')
```
</details>

## Miscellaneous
I manually tested `H` and `rho0` batching, that `exp_ops` keep working, and that `tmeas` does its job.

What's missing:
- (easy to solve) the same key is used by each batch instance
- (easy to solve) result.wiener is not correctly batched
- (tedious but easy) write tests
- (meeeh) the outer smesolve function is not JIT-compatible due to the split between measured and dissipative operators
- (meeeeh) performance is not very good

### DIFF
--- a/dynamiqs/result.py
+++ b/dynamiqs/result.py
@@ -2,13 +2,19 @@ from __future__ import annotations
 
 import equinox as eqx
 from jax import Array
-from jaxtyping import PyTree
+from jaxtyping import PRNGKeyArray, PyTree
 
 from .gradient import Gradient
 from .options import Options
 from .solver import Solver
 
-__all__ = ['SESolveResult', 'MESolveResult', 'SEPropagatorResult', 'MEPropagatorResult']
+__all__ = [
+    'SESolveResult',
+    'MESolveResult',
+    'SMESolveResult',
+    'SEPropagatorResult',
+    'MEPropagatorResult',
+]
 
 
 def memory_bytes(x: Array) -> int:
@@ -37,6 +43,10 @@ class Saved(eqx.Module):
 class SolveSaved(Saved):
     Esave: Array | None
     extra: PyTree | None
+
+
+class SMESolveSaved(SolveSaved):
+    Jsave: Array
 
 
 class PropagatorSaved(Saved):
@@ -228,6 +238,78 @@ class MESolveResult(SolveResult):
         [Batching simulations](../../documentation/basics/batching-simulations.md)
         tutorial for more details.
     """
+
+
+class SMESolveResult(SolveResult):
+    r"""Result of the diffusive SME integration.
+
+    For the shape indications we name `ntrajs` the number of trajectories
+    (`ntrajs = len(keys)`) and `nLm` the number of loss channels for which the
+    measurement efficiency is not null.
+
+    Attributes:
+        states _(array of shape (..., ntrajs, ntsave, n, n))_: Saved states.
+        final_state _(array of shape (..., ntrajs, n, n))_: Saved final state.
+        measurements _(array of shape (..., ntrajs, nLm, ntmeas-1))_: Saved
+            measurements.
+        expects _(array of shape (..., ntrajs, len(exp_ops), ntsave) or None)_: Saved
+            expectation values, if specified by `exp_ops`.
+        extra _(PyTree or None)_: Extra data saved with `save_extra()` if
+            specified in `options` (see [`dq.Options`][dynamiqs.Options]).
+        infos _(PyTree or None)_: Solver-dependent information on the resolution.
+        tsave _(array of shape (ntsave,))_: Times for which results were saved.
+        tmeas _(array of shape (ntmeas,))_: Times for which measurement signals are
+            averaged and saved.
+        keys _(PRNG key array of shape (ntrajs,))_: PRNG keys used to sample the Wiener
+            processes.
+        solver _(Solver)_: Solver used.
+        gradient _(Gradient)_: Gradient used.
+        options _(Options)_: Options used.
+
+    Note-: Result of running multiple simulations concurrently
+        The resulting states, measurements and expectation values are batched according
+        to the leading dimensions of the Hamiltonian `H` and initial state `rho0`. The
+        behaviour depends on the value of the `cartesian_batching` option
+
+        === "If `cartesian_batching = True` (default value)"
+            The results leading dimensions are
+            ```
+            ... = ...H, ...rho0
+            ```
+            For example if:
+
+            - `H` has shape _(2, 3, n, n)_,
+            - `rho0` has shape _(4, n, n)_,
+
+            then `states` has shape _(2, 3, 4, ntrajs, ntsave, n, n)_.
+
+        === "If `cartesian_batching = False`"
+            The results leading dimensions are
+            ```
+            ... = ...H = ...rho0  # (once broadcasted)
+            ```
+            For example if:
+
+            - `H` has shape _(2, 3, n, n)_,
+            - `rho0` has shape _(3, n, n)_,
+
+            then `states` has shape _(2, 3, ntrajs, ntsave, n, n)_.
+
+        See the
+        [Batching simulations](../../documentation/basics/batching-simulations.md)
+        tutorial for more details.
+    """
+
+    tmeas: Array
+    keys: PRNGKeyArray
+
+    @property
+    def measurements(self) -> Array:
+        return self._saved.Jsave
+
+    def _str_parts(self) -> dict[str, str]:
+        d = super()._str_parts()
+        return d | {'Measurements': array_str(self.measurements)}
 
 
 class SEPropagatorResult(PropagatorResult):
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -4,3 +4,5 @@
 *[Array-like objects]: e.g. Python lists, NumPy and JAX arrays or QuTiP Qobjs
 *[PWC]: piecewise constant
 *[SME]: stochastic master equation
+*[SMEs]: stochastic master equations
+*[PRNG]: pseudorandom number generator
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -43,6 +43,7 @@ The **dynamiqs** Python API features two main types of functions: solvers of dif
         - Rouchon1
         - Rouchon2
         - Expm
+        - Milstein
 
 ### Gradients (dq.gradient)
 
--- a/dynamiqs/integrators/apis/smesolve.py
+++ b/dynamiqs/integrators/apis/smesolve.py
@@ -2,13 +2,33 @@
 
 from __future__ import annotations
 
-from jaxtyping import ArrayLike
+import logging
+from functools import partial
 
+import jax
+import jax.numpy as jnp
+from jax import Array
+from jaxtyping import ArrayLike, PRNGKeyArray
+
+from ..._checks import check_shape, check_times
+from ..._utils import cdtype
 from ...gradient import Gradient
 from ...options import Options
-from ...result import Result
-from ...solver import Solver
-from ...time_array import TimeArray
+from ...result import SMESolveResult
+from ...solver import Euler, Milstein, Solver
+from ...time_array import Shape, TimeArray
+from ...utils.quantum_utils.general import todm
+from .._utils import (
+    _astimearray,
+    _cartesian_vectorize,
+    _flat_vectorize,
+    catch_xla_runtime_error,
+    get_integrator_class,
+)
+from ..smesolve.diffrax_integrator import (
+    SMESolveEulerIntegrator,
+    SMESolveMilsteinIntegrator,
+)
 
 
 def smesolve(
@@ -17,19 +37,20 @@ def smesolve(
     etas: ArrayLike,
     rho0: ArrayLike,
     tsave: ArrayLike,
+    keys: list[PRNGKeyArray],
     *,
     tmeas: ArrayLike | None = None,
-    ntrajs: int = 10,
     exp_ops: list[ArrayLike] | None = None,
-    solver: Solver | None = None,
+    solver: Solver = Milstein(),  # noqa: B008
     gradient: Gradient | None = None,
     options: Options = Options(),  # noqa: B008
-) -> Result:
+) -> SMESolveResult:
     r"""Solve the diffusive stochastic master equation (SME).
 
     Warning:
-        This function has not been ported to JAX yet. The following documentation is
-        a draft API, copied from the old PyTorch version of the library.
+        This function is still under test and development. Please
+        [open an issue on GitHub](https://github.com/dynamiqs/dynamiqs/issues/new) if
+        you encounter any bug.
 
     This function computes the evolution of the density matrix $\rho(t)$ at time $t$,
     starting from an initial state $\rho_0$, according to the diffusive SME in Itô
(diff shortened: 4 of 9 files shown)


pierreguilmin:
Follow-up issue: https://linear.app/dynamiqs/issue/DYN-196/improve-dqsmesolve.

dynamiqs/result.py
```diff
@@ -120,3 +120,48 @@ class MEResult(Result):
...
+class SMESaved(Saved):
+    Isave: Array
+    wiener: PyTree
+
+
+class SMEResult(Result):
+    """Result of the SME integration.
+
+    Attributes:
+        states _(array of shape (nH?, nrho0?, ntrajs, ntsave, n, n))_: Saved states.
```
gautierronan, line 134: It's the same in `MEResult`, but I guess we could also batch over jump operators. In any case, we'll have to properly document this at some point: it's very much related to the discussion on TimeArrays.

I'm thinking a notation like `(nH..., nLs..., nrho0..., ntrajs, ntsave, n, n)` could be more explicit than what we currently have.

dynamiqs/smesolve/smediffrax.py
```diff
@@ -0,0 +1,117 @@
...
+
+# state for the diffrax solver for SMEs
+class Y(eqx.Module):
+    rho: Array
+    dYt: Array
+
+
+class MeasurementTerm(dx.ControlTerm):
+    def prod(self, vf: dx.VF, control: dx.Control) -> dx.Y:
+        dW = control
```
gautierronan, line 23: Why not just name it `dW` in the arguments? Would be more explicit IMO. This looks like there's additional logic.

dynamiqs/smesolve/smediffrax.py
```diff
@@ -0,0 +1,117 @@
...
+
+            return Y(rho, dYt)
+
+        lindblad_term = dx.ODETerm(vector_field)
+
+        # === define stochastic term
+        # This contains everything before the "dWt" in the SME:
+        # - measurement backaction term for drho
+        # - simply dWt for dYt
+        def vector_field(t, y, _):  # noqa: ANN001, ANN202
```
gautierronan, line 76: Can we name this in a different way? We have two `def vector_field` in the same scope: this doesn't seem very good practice (although I agree it's fine here).

dynamiqs/smesolve/smediffrax.py
```diff
@@ -0,0 +1,117 @@
...
+            tmp = Lms_rho + dag(Lms_rho)
+            tr = trace(tmp).real
+
+            # state
+            etas = self.etas[:, None, None]  # (nLm, n, n)
+            tr = tr[:, None, None]  # (nLm, n, n)
+            rho = jnp.sqrt(etas) * (tmp - tr * y.rho)  # (nLm, n, n)
+
+            # signal
+            dYt = jnp.empty(len(self.etas))
```
gautierronan, line 89: Should this not be `jnp.zeros` instead? `empty` will just allocate memory without initializing, which might be added to the actual result later on. Maybe I'm missing something.

dynamiqs/smesolve/smesolve.py
```diff
@@ -19,19 +35,19 @@ def smesolve(
...
     options: Options = Options(),  # noqa: B008
-) -> Result:
+) -> SMEResult:
     r"""Solve the diffusive stochastic master equation (SME).
 
     Warning:
-        This function has not been ported to JAX yet. The following documentation is
-        a draft API, copied from the old PyTorch version of the library.
+        This function is still under test and development, it is not as stable as
+        [`dq.sesolve()`][dynamiqs.sesolve] and [`dq.mesolve()`][dynamiqs.mesolve].
```
gautierronan, line 50:
```suggestion
Please raise any bug you may encounter on github.com/dynamiqs/dynamiqs.
```

dynamiqs/smesolve/smesolve.py
```diff
@@ -63,20 +79,20 @@ def smesolve(
...
 
-    The signals $I_k(t)$ are singular quantities, the solver returns the averaged signals
-    $J_k(t)$ defined for a time interval $[t_0, t_1)$ by:
+    The signals $I_k(t)$ are singular quantities, the solver returns the averaged
+    signals $J_k(t)$ defined for a time interval $[t_0, t_1[$ by:
     $$
-        J_k([t_0, t_1)) = \frac{1}{t_1-t_0}\int_{t_0}^{t_1} I_k(t) \dt
-        = \frac{1}{t_1-t_0}\int_{t_0}^{t_1} \dd y_k(t).
+        J_k([t_0, t_1[) = \frac{1}{t_1-t_0}\int_{t_0}^{t_1} I_k(t)\ \dt
+        = \frac{1}{t_1-t_0}\int_{t_0}^{t_1} \dd Y_k(t).
```
gautierronan, line 95: I probably missed it, but I thought we were actually returning the `dYt` instead, with the way you wrote it (with the Pytree state). Anyway, just want to make sure :)

dynamiqs/smesolve/smesolve.py
```diff
@@ -91,31 +107,240 @@ def smesolve(
...
+        tutorial for more details
+
+    Warning:
+        Batching on `jump_ops` and `etas is not yet supported, if this is needed don't
+        hesitate to
+        [open an issue on GitHub](https://github.com/dynamiqs/dynamiqs/issues/new).
+
+    Warning:
+        For now the same key is used for every batched simulation over `H` and `rho0`.
+        This will be probably change in the future.
```
gautierronan, line 122: Isn't this just a matter of splitting the user keys before vmapping?

gautierronan (changes requested):
This looks really great. Very nice work. I have several comments or questions below, but mostly nit. As you stated, there's still some follow-up work to make, but I think it's a great start!

dynamiqs/result.py
pierreguilmin: Yes, it's tracked in https://linear.app/dynamiqs/issue/DYN-73/support-generalized-batching.

> I'm thinking a notation like `(nH..., nLs..., nrho0..., ntrajs, ntsave, n, n)` could be more explicit than what we currently have.
Once we've implented the ND batching yes, I like `H: (...H, n, n)` where I read `...H` as "all the other dimensions of H`.

Note that documenting the batching over `jump_ops` is a bit tricky because technically it is `(...L1, ...L2, ..., ...L_nL)`.

dynamiqs/smesolve/smediffrax.py
pierreguilmin: I like using the notations from diffrax API/source code in the overriden function signatures, I think it's neat to understand how our problem maps to the diffrax API. When I read
```python
def prod(self, vf: dx.VF, control: dx.Control) -> dx.Y:
```
I know the `prod` method of the diffrax `ControlTerm` is  reimplemented, while e.g.
```python
def prod(self, rho: dx.VF, dW: dx.Control) -> dx.Y:
```
is more obscure IMO.

In any case, when overriding a base class method it's best practice not to rename the arguments.

dynamiqs/smesolve/smediffrax.py
pierreguilmin: Do you prefer `def vector_field_deterministic()` and `def vector_field_stochastic()`?

dynamiqs/smesolve/smediffrax.py
pierreguilmin: This is correct. In `prod` we just throw this to the bin, and return `Y(rho, dW)` (there is no signal-dependent term in the stochastic part, it's just `dYt = deterministic + dWt`. I think using `jnp.empty` is a good way to express that this is unused, I can maybe put a comment to make it more clear.

dynamiqs/smesolve/smesolve.py
pierreguilmin: We indeed accumulate the dYt. The diffrax state at time $t$ is basically

$$\left(\sum_{k=0}^n \mathrm{d}\rho_{k\mathrm{d}t}, \sum_{k=0}^n \mathrm{d}Y_{k\mathrm{d}t}\right)$$

I can add a comment to explain this better too.

But then we do (L104-106 of smediffrax.py):
```python
        # Diffrax integrates the state from t0 to t1. In this case, the state is
        # (rho, dYt). So we recover the signal by simply diffing the resulting array.
        Isave = jnp.diff(integrated_dYt, axis=0)
```
to recover the integrated signal.

dynamiqs/smesolve/smediffrax.py
pierreguilmin: Another option (to explain) would be to do `dYt = jnp.ones(len(self.etas))` and in `prod`:
```
    def prod(self, vf: dx.VF, control: dx.Control) -> dx.Y:
        dW = control
        rho = (vf.rho * dW[:, None, None]).sum(0)  # (n, n)
        dY = vf.dY * dW
        return Y(rho, dY)
```

dynamiqs/smesolve/smesolve.py
pierreguilmin: Actually I think the notations are not very rigorous in the code, I will fix this. Can you read this Notion page and tell me if it's clearer? > https://pierreguilmin.notion.site/SME-diffrax-176db6e9e8cc4bdfba4fb30121642625?pvs=4.

gautierronan: The notations are good, but I'm still bothered by something. We simulate the continuous SME for which there is an infinitesimal backaction term because we assume a continuous measurement of the quantum system. But we return the integrated signal, which would correspond to a different SME, with a backaction term that only accounts for binned information about the signal.

Am I missing something?

In other words, I feel like there are two equations:

1. the continuous time SME with infinitesimal backaction and continuous signal,
2. the "binned" SME with binned backaction and discrete signals.

I'm not sure about this second equation: does it even exist/make sense?

dynamiqs/smesolve/smesolve.py
gautierronan: How about both?

By default, the user passes a single key, which we split and distribute among batches (with a different key per batch: seems the most natural to me and what users will expect). And we also allow for passing an array of keys of the same shape as the batch, in which case we don't do the internal splitting and directly pass each key to its corresponding batch.

This second part can come in a separate PR, later on.

dynamiqs/smesolve/smesolve.py
pierreguilmin: Let me try to clarify. The discrete-time formulation converges (in law) to the continuous SME as the chosen time step $\mathrm{d}t\to0$. This means that you will get the same statistics for the state and signal as the continuous SME if you take a sufficiently small $\mathrm{dt}$. We _could_ return $\mathrm{d}Y_t/\mathrm{d}t$ for each time step $\mathrm{d}t$ we simulate, but (i) this can easily become super costly in memory if there are lots of timestep, (ii) this quantity should be understood with care for adaptive time-stepping and (iii) it can anyway be reconstructed _if needed_ from the Virtual Brownian Tree holding the sampled Wiener process.

I think having the integrated signal in between user-defined times allows for an easier interpretation and rigorous work. But we could have an option to return the full signal too, and all the time steps, that'd be great actually!

gautierronan (approved):
I forgot what we concluded exactly regarding the API of `smesolve`, but I think this PR looks great as is. I think we should merge and iterate on top with smaller PRs, and we're probably not so far from merging on main too (once we fix the first three bullet points of your to-do list at least, the rest can wait).

PR #550, opened 2024-04-15
Document `dq.gridplot()` and `dq.mplstyle()`

pierreguilmin:
Closes DYN-177.

These APIs are used in the doc.

<details>
  <summary>dq.gridplot() doc screenshot</summary>

gridplot - dynamiqs

</details>

<details>
  <summary> dq.mplstyle() doc screenshot</summary>

mplstyle - dynamiqs

</details>

### DIFF
--- a/dynamiqs/plots/utils.py
+++ b/dynamiqs/plots/utils.py
@@ -16,7 +16,7 @@ from matplotlib.figure import Figure
 from matplotlib.ticker import FixedLocator, MultipleLocator, NullLocator
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
-__all__ = ['gridplot']
+__all__ = ['gridplot', 'mplstyle']
 # __all__ = [
 #     'linmap',
 #     'figax',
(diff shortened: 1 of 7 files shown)


dynamiqs/plots/utils.py
```diff
@@ -77,36 +77,61 @@ def wrapper(  # noqa: ANN202
...
+    n: int,
+    nrows: int = 1,
+    *,
+    w: float = 3.0,
+    h: float | None = None,
+    sharexy: bool = False,
+    **kwargs,
 ) -> tuple[Figure, Iterable[Axes]]:
-    """Return an iterator on `Axes` objects organised in a grid fashion.
+    """Returns a figure and a list of subplots organised in a grid.
```
gautierronan, line 88: Is it not an iterator still?

gautierronan (dismissed):
I have to admit I find a bit strange that dynamiqs provides this `gridplot` function. It seems to me that this should be within a separate library. But it's fine for now, and indeed very useful.

Just a nit below, otherwise looks good.

pierreguilmin:
> I have to admit I find a bit strange that dynamiqs provides this gridplot function. It seems to me that this should be within a separate library. But it's fine for now, and indeed very useful.

I 100% agree! IMO all these plotting functions should move to a different library at some point. We use `dq.gridplot` in the documentation though, so we have to document it publicly.

PR #552, opened 2024-04-16
Add new batching API for time arrays and support for ND batching

### LINKED ISSUE #547: Allow Hamiltonian arguments to be multidimensional without batching
I'd like for some of the arguments I pass to my Hamiltonian functions to be multidimensional. A simple example is the following where I have multiple drive Hamiltonians and want to specify the amplitudes as an array.

```python
import dynamiqs as dq
import jax.numpy as jnp
from dynamiqs import timecallable
from dynamiqs.solver import Tsit5

omega = 2.0 * jnp.pi * 1.0
H0 = -0.5 * omega * dq.sigmaz()
H1s = [dq.sigmax(), dq.sigmay()]

def H_func(t, amps):
    H = H0 + jnp.einsum("i,ijk->jk", jnp.cos(omega * t) * amps, H1s)
    return H

H_tc = timecallable(H_func, args=(jnp.array([0.01, 0.02]),))
tsave = jnp.linspace(0, 1.0, 21)
initial_states = [dq.basis(2, 1), ]
result = dq.sesolve(
    H_tc,
    initial_states,
    tsave,
    solver=Tsit5(),
)

``` 
This code raises an error `ValueError: Einstein sum subscript 'i' does not contain the correct number of indices for operand 0.
` as the code attempts to batch over `amps` which is not my intent. Of course I could just pass them in as individual arguments `amp_x` and `amp_y` but I have more complex examples in mind where this wouldn't work as easily. I'm wondering if there is an alternative to checking argument dimensions for batching? Might it make more sense to make it similar to how batching over e.g. initial states works, and stack `timecallable` instances?

abocquet:
This PR:
i) removes the `args` from `TimeArray`s
ii) allows for vmapping on time arrays
iii) allows for vmapping on ND time arrays

Note that support for batched times in pwc time arrays is not far, but I keep it for a next PR

Closes #547.

Related to DYN-197.

### DIFF
--- a/docs/tutorials/time-dependent-operators.md
+++ b/docs/tutorials/time-dependent-operators.md
@@ -155,7 +155,7 @@ Let's define the modulated operator $H=\cos(2\pi t)\sigma_x$:
 >>> f = lambda t: jnp.cos(2.0 * jnp.pi * t)
 >>> H = dq.modulated(f, dq.sigmax())
 >>> type(H)
-<class 'dynamiqs.time_array.CallableTimeArray'>
+<class 'dynamiqs.time_array.ModulatedTimeArray'>
 >>> H.shape
 (2, 2)
 ```
--- a/docs/tutorials/time-dependent-operators.md
+++ b/docs/tutorials/time-dependent-operators.md
@@ -155,7 +155,7 @@ Let's define the modulated operator $H=\cos(2\pi t)\sigma_x$:
 >>> f = lambda t: jnp.cos(2.0 * jnp.pi * t)
 >>> H = dq.modulated(f, dq.sigmax())
 >>> type(H)
-<class 'dynamiqs.time_array.ModulatedTimeArray'>
+<class 'dynamiqs.time_array.CallableTimeArray'>
 >>> H.shape
 (2, 2)
 ```
@@ -173,9 +173,9 @@ Array([[0.+0.j, 1.+0.j],
 ??? Note "Function with additional arguments"
     To define a modulated time-array with additional arguments, you can use the optional `args` parameter of [`dq.modulated()`][dynamiqs.modulated]:
     ```python
-    f = lambda t, omega: jnp.cos(omega * t)
+    f = lambda t: jnp.cos(omegas * t)
     omega = 1.0
-    H = dq.modulated(f, dq.sigmax(), args=(omega,))
+    H = dq.modulated(f, dq.sigmax())
     ```
 
 ### Arbitrary time-dependent operators
@@ -218,9 +218,9 @@ Array([[1., 0.],
 ??? Note "Function with additional arguments"
     To define a callable time-array with additional arguments, you can use the optional `args` parameter of [`dq.timecallable()`][dynamiqs.timecallable]:
     ```python
-    f = lambda t, x: x * jnp.array([[t, 0], [0, 1 - t]])
     x = 1.0
-    H = dq.timecallable(f, args=(x,))
+    f = lambda t: x * jnp.array([[t, 0], [0, 1 - t]])
+    H = dq.timecallable(f)
     ```
 
 ## Batching and differentiation with time-arrays
@@ -229,7 +229,7 @@ For modulated and arbitrary time-dependent operators, any array that is batched 
 
 For example to define a modulated Hamiltonian $H=\cos(\omega t)\sigma_x$ batched or differentiated over the parameter $\omega$:
 ```python
-f = lambda t, omega: jnp.cos(omega * t)
 omegas = jnp.linspace(0.0, 2.0, 10)
-H = dq.modulated(f, dq.sigmax(), args=(omegas,))
+f = lambda t: jnp.cos(omegas * t)
+H = dq.modulated(f, dq.sigmax())
 ```
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -9,7 +9,6 @@ from ..solver import Solver
 from ..time_array import (
     CallableTimeArray,
     ConstantTimeArray,
-    ModulatedTimeArray,
     PWCTimeArray,
     SummedTimeArray,
     TimeArray,
@@ -27,7 +26,7 @@ def _astimearray(x: ArrayLike | TimeArray) -> TimeArray:
             return ConstantTimeArray(array)
         except (TypeError, ValueError) as e:
             raise TypeError(
-                f'Argument must be an array-like or a time-array object, but has type'
+                'Argument must be an array-like or a time-array object, but has type'
                 f' {obj_type_str(x)}.'
             ) from e
 
@@ -93,10 +92,6 @@ def is_timearray_batched(tarray: TimeArray) -> TimeArray:
         return ConstantTimeArray(tarray.array.ndim > 2)
     elif isinstance(tarray, PWCTimeArray):
         return PWCTimeArray(False, tarray.values.ndim > 1, False)
-    elif isinstance(tarray, ModulatedTimeArray):
-        return ModulatedTimeArray(
-            False, False, tuple(arg.ndim > 0 for arg in tarray.args)
-        )
     elif isinstance(tarray, CallableTimeArray):
         return CallableTimeArray(False, tuple(arg.ndim > 0 for arg in tarray.args))
     else:
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import equinox as eqx
 import jax
 import jax.numpy as jnp
+import jax.tree_util as jtu
 from jaxtyping import Array, ArrayLike, PyTree
 
 from .._utils import cdtype, obj_type_str
@@ -77,34 +78,27 @@ def compute_vmap(
     # - If `cartesian_batching` is False, we directly map f over all batched arguments
     #   and apply vmap once.
 
-    if any(n > 0 for n in n_batch):
+    # we use tree_utils to handle nested batching such as `jump_ops`
+    leaves, treedef = jtu.tree_flatten(n_batch)
+    if any(leaf > 0 for leaf in leaves):
         if cartesian_batching:
             # map over each batched dimension separately
             # note: we apply the successive vmaps in reverse order, so the output
             # batched dimensions are in the correct order
-            for i, c_n_batch in reversed(list(enumerate(n_batch))):
-                if c_n_batch > 0:
+            for i, leaf in reversed(list(enumerate(leaves))):
+                if leaf > 0:
                     # build the `in_axes` argument with the same structure as
                     # `is_batched`, but with 0 at the `leaf` position
-                    in_axes = [None] * len(n_batch)
+                    in_axes = jtu.tree_map(lambda _: None, leaves)
                     in_axes[i] = 0
-                    for _ in range(c_n_batch):
+                    in_axes = jtu.tree_unflatten(treedef, in_axes)
+                    print(in_axes)
+                    for _ in range(leaf):
                         f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
         else:
             # map over all batched dimensions at once
-            n = None
-            for cn in n_batch:
-                if cn == 0:
-                    pass
-                elif n is None:
-                    n = cn
-                elif cn != n:
-                    raise ValueError(
-                        f"All objects must have the same "
-                        f"batching dimension but got {cn} and {n}"
-                    )
-
-            in_axes = [0 if x > 0 else None for x in n_batch]
+            in_axes = jtu.tree_map(lambda x: 0 if x > 0 else None, n_batch)
+            n = jtu.tree_reduce(max, n_batch)
             for _ in range(n):
                 f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
 
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -7,13 +7,7 @@ from jaxtyping import ArrayLike, PyTree
 
 from .._utils import cdtype, obj_type_str
 from ..solver import Solver
-from ..time_array import (
-    CallableTimeArray,
-    ConstantTimeArray,
-    PWCTimeArray,
-    SummedTimeArray,
-    TimeArray,
-)
+from ..time_array import ConstantTimeArray, TimeArray
 from .abstract_solver import AbstractSolver
 
 
@@ -45,10 +39,14 @@ def get_solver_class(
 
 
 def compute_vmap(
-    f: callable, cartesian_batching: bool, n_batch: [int], out_axes: PyTree[int | None]
+    f: callable,
+    cartesian_batching: bool,
+    n_batch: PyTree[int],
+    out_axes: PyTree[int | None],
 ) -> callable:
     # This function vectorizes `f` by applying jax.vmap over batched dimensions. The
-    # argument `is_batched` indicates for each argument of `f` whether it is batched.
+    # argument `n_batch` indicates for each argument of `f` the number of batching
+    # dimensions it has.
     # There are two possible strategies to vectorize `f`:
     # - If `cartesian_batching` is True, we want to apply `f` to every possible
     #   combination of the arguments batched dimension (the cartesian product). To do
@@ -81,19 +79,3 @@ def compute_vmap(
                 f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
 
     return f
-
-
-def is_timearray_batched(tarray: TimeArray) -> TimeArray:
-    # This function finds all batched arrays within a given TimeArray.
-    # To do so, it goes down the PyTree and identifies batched fields depending
-    # on the type of TimeArray.
-    if isinstance(tarray, SummedTimeArray):
-        return SummedTimeArray([is_timearray_batched(arr) for arr in tarray.timearrays])
-    elif isinstance(tarray, ConstantTimeArray):
-        return ConstantTimeArray(tarray.array.ndim > 2)
-    elif isinstance(tarray, PWCTimeArray):
-        return PWCTimeArray(False, tarray.values.ndim > 1, False)
-    elif isinstance(tarray, CallableTimeArray):
-        return CallableTimeArray(False, tuple(arg.ndim > 0 for arg in tarray.args))
-    else:
-        raise TypeError(f'Unsupported TimeArray type: {type(tarray).__name__}')
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
+import equinox as eqx
 import jax
 import jax.numpy as jnp
-from jaxtyping import ArrayLike, PyTree
+from jaxtyping import Array, ArrayLike, PyTree
 
 from .._utils import cdtype, obj_type_str
 from ..solver import Solver
@@ -31,6 +32,23 @@ def _astimearray(x: ArrayLike | TimeArray) -> TimeArray:
             ) from e
 
 
+class BatchedCallable(eqx.Module):
+    f: callable[[float], Array]
+    indices: list[Array]
+
+    def __init__(self, f: callable[[float], Array]):
+        shape = jax.eval_shape(f, 0.0).shape
+        self.f = f
+        (*self.indices,) = jnp.indices(shape[:-2])
+
+    def __call__(self, t: float) -> Array:
+        return self.f(t)[*self.indices]
+
+    @property
+    def ndim(self) -> int:
+        return jax.eval_shape(self.f, 0.0).ndim
+
+
 def get_solver_class(
     solvers: dict[Solver, AbstractSolver], solver: Solver
 ) -> AbstractSolver:
@@ -46,7 +64,7 @@ def get_solver_class(
 def compute_vmap(
     f: callable,
     cartesian_batching: bool,
-    is_batched: PyTree[bool],
+    n_batch: [int],
     out_axes: PyTree[int | None],
 ) -> callable:
     # This function vectorizes `f` by applying jax.vmap over batched dimensions. The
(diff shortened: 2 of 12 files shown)


pierreguilmin (commented):
Looks good overall, very cool! 🥳

I'll do a pass on this later to fix a few things here and there. In the meantime, can you add a description to the PR to explain the two things you do here: (i) support ND batching and (ii) fix time-arrays API? Ideally we should split the PR (first (i) then (ii)), but it's OK if that's too tedious.

Note to myself: I also need to fix documentation in another PR on top.

abocquet:
Thanks for your feedback Pierre. I don't really feel like splitting this one but if it feels hairy we can pair up so as to speed up the reviews and make them less of a headache for you (and @gautierronan too)

dynamiqs/core/_utils.py
```diff
@@ -93,10 +93,6 @@ def is_timearray_batched(tarray: TimeArray) -> TimeArray:
         return ConstantTimeArray(tarray.array.ndim > 2)
     elif isinstance(tarray, PWCTimeArray):
         return PWCTimeArray(False, tarray.values.ndim > 1, False)
-    elif isinstance(tarray, ModulatedTimeArray):
-        return ModulatedTimeArray(
-            False, False, tuple(arg.ndim > 0 for arg in tarray.args)
-        )
```
pierreguilmin, line 99: I guess we should have this back.

dynamiqs/core/_utils.py
```diff
@@ -45,10 +45,7 @@ def get_solver_class(
 
 
 def compute_vmap(
-    f: callable,
-    cartesian_batching: bool,
-    is_batched: PyTree[bool],
-    out_axes: PyTree[int | None],
+    f: callable, cartesian_batching: bool, n_batch: [int], out_axes: PyTree[int | None]
 ) -> callable:
     # This function vectorizes `f` by applying jax.vmap over batched dimensions. The
```
pierreguilmin, line 50: This inline comments should be updated throughout the function.

dynamiqs/core/_utils.py
```diff
@@ -45,10 +45,7 @@ def get_solver_class(
 
 
 def compute_vmap(
-    f: callable,
-    cartesian_batching: bool,
-    is_batched: PyTree[bool],
-    out_axes: PyTree[int | None],
+    f: callable, cartesian_batching: bool, n_batch: [int], out_axes: PyTree[int | None]
```
pierreguilmin, line 48: `n_batch` typing is incorrect.

dynamiqs/mesolve/mesolve.py
```diff
@@ -200,7 +195,7 @@ def _check_mesolve_args(
         )
 
     # === check rho0 shape
-    check_shape(rho0, 'rho0', '(?, n, 1)', '(?, n, n)', subs={'?': 'nrho0?'})
+    check_shape(rho0, 'rho0', '(..., n, 1)', '(?, n, n)', subs={'?': 'nrho0?'})
```
pierreguilmin, line 198:
```suggestion
    check_shape(rho0, 'rho0', '(..., n, 1)', '(..., n, n)', subs={'...': '...rho0'})
```

dynamiqs/sesolve/sesolve.py
```diff
@@ -167,7 +157,7 @@ def _sesolve(
 def _check_sesolve_args(H: TimeArray, psi0: Array, exp_ops: Array | None):
     # === check H shape
     check_shape(H, 'H', '(..., n, n)')
-    check_shape(psi0, 'psi0', '(?, n, 1)', subs={'?': 'npsi0?'})
+    check_shape(psi0, 'psi0', '(..., n, 1)', subs={'?': 'npsi0?'})
```
pierreguilmin, line 160:
```suggestion
    check_shape(psi0, 'psi0', '(..., n, 1)', subs={'...': '...psi0'})
```

tests/sesolve/test_batching.py
```diff
@@ -73,18 +58,16 @@ def test_timearray_batching():
...
     deltas = jnp.linspace(0.0, 1.0, 4)
-    H_mod = dq.modulated(lambda t, delta: jnp.cos(t * delta), H0, args=(deltas,))
+    H_mod = dq.modulated(lambda t: jnp.cos(t * deltas), H0)
 
     result = dq.sesolve(H_mod, psi0, times)
     assert result.states.shape == (4, 11, 4, 1)
-    result = dq.sesolve(H0 + H_mod, psi0, times)
-    assert result.states.shape == (4, 11, 4, 1)
+    # result = dq.sesolve(H0 + H_mod, psi0, times)
+    # assert result.states.shape == (4, 11, 4, 1)
```
pierreguilmin, line 66: Is time-array addition broken?

tests/test_time_array.py
```diff
@@ -75,7 +75,7 @@ class TestCallableTimeArray:
     @pytest.fixture(autouse=True)
     def _setup(self):
         f = lambda t: t * jnp.array([1, 2])
-        self.x = CallableTimeArray(f, ())
+        self.x = timecallable(f)
```
pierreguilmin, line 78: Can you also replace calls to `ConstantTimeArray` with the `constant()` constructer?

dynamiqs/time_array.py
```diff
@@ -69,35 +70,37 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
...
 
     A modulated time-array is defined by $O(t) = f(t) O_0$ where $f(t)$ is a
     time-dependent scalar. The function $f$ is defined by passing a Python function
-    with signature `f(t: float, *args: PyTree) -> Array` that returns an array of shape
+    with signature `f(t: float) -> Array` that returns an array of shape
     _(...)_ for any time $t$.
 
     Args:
         f _(function returning array of shape (...))_: Function with signature
             `f(t: float, *args: PyTree) -> Array` that returns the modulating factor
```
pierreguilmin, line 101:
```suggestion
            `f(t: float) -> Array` that returns the modulating factor
```

dynamiqs/time_array.py
```diff
@@ -389,49 +402,43 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
             return NotImplemented
 
 
-class ModulatedTimeArray(TimeArray):
```
pierreguilmin, line 392: I guess you've inverted the order of `ModulatedTimeArray` and `CallableTimeArray` in the file and it's a bit too much for git to track changes properly. Can you re-order with `class ModulatedTimeArray` first and then `class CallableTimeArray` (to see the actual changes to these classes, and to match the `__all__` and documentation order).

dynamiqs/time_array.py
```diff
@@ -443,45 +450,47 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
...
 
     def reshape(self, *new_shape: int) -> TimeArray:
-        f = Partial(lambda t, *args: self.f(t, *args).reshape(*new_shape))
-        return CallableTimeArray(f, self.args)
+        f_shape = jax.eval_shape(self.f, 0.0).shape
+        new_f_shape, new_array_shape = _split_shape(
+            new_shape, f_shape, self.array.shape
+        )
+        f = Partial(lambda t, *args: self.f(t, *args).reshape(*new_f_shape))
+        return ModulatedTimeArray(f, self.array.reshape(*new_array_shape))
```
pierreguilmin, line 476: I guess we don't need the `*args`.

dynamiqs/time_array.py
```diff
@@ -443,45 +450,47 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
...
 
     def conj(self) -> TimeArray:
         f = Partial(lambda t, *args: self.f(t, *args).conj())
```
pierreguilmin, line 479: Same, we don't need the `*args`.

dynamiqs/time_array.py
```diff
@@ -245,6 +241,14 @@ def conj(self) -> TimeArray:
             New time-array object with element-wise complex conjuguated values.
         """
 
+    @abstractmethod
+    def as_batched_callable(self) -> callable:
```
pierreguilmin, line 245:
```suggestion
    def as_batched_callable(self) -> callable[[float], Array]:
```

dynamiqs/time_array.py
```diff
@@ -69,35 +70,37 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
...
+    #     times = jnp.broadcast_to(times, values_batching + times.shape)
+    #     array = jnp.broadcast_to(array, values_batching + array.shape)
+
     return PWCTimeArray(times, values, array)
 
 
-def modulated(
-    f: callable[[float, ...], Array], array: ArrayLike, *, args: tuple[PyTree] = ()
-) -> ModulatedTimeArray:
+def modulated(f: callable[[float, ...], Array], array: ArrayLike) -> CallableTimeArray:
```
pierreguilmin, line 91:
```suggestion
def modulated(f: callable[[float], Array], array: ArrayLike) -> CallableTimeArray:
```

dynamiqs/time_array.py
```diff
@@ -113,28 +116,23 @@ def modulated(
...
     f = Partial(f)
 
-    return ModulatedTimeArray(f, array, args)
+    return ModulatedTimeArray(f, array)
 
 
-def timecallable(
-    f: callable[[float, ...], Array], *, args: tuple[PyTree] = ()
-) -> CallableTimeArray:
+def timecallable(f: callable[[float, ...], Array]) -> CallableTimeArray:
```
pierreguilmin, line 125:
```suggestion
def timecallable(f: callable[[float], Array]) -> CallableTimeArray:
```

dynamiqs/time_array.py
```diff
@@ -329,36 +336,22 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
 
 
 class PWCTimeArray(TimeArray):
-    times: Array  # (nv+1,)
+    times: Array  # (..., nv+1,)
     values: Array  # (..., nv)
-    array: Array  # (n, n)
+    array: Array  # (..., n, n)
```
pierreguilmin, line 341: Can we start with the batching only carried by `values` first? Supporting batching on arrays here is a new feature (that needs added documentation, tests, etc...), I'd rather add this in another PR if possible, to limit the amount of change in this PR.

dynamiqs/time_array.py
```diff
@@ -302,15 +306,18 @@ def shape(self) -> tuple[int, ...]:
     def mT(self) -> TimeArray:
         return ConstantTimeArray(self.array.mT)
 
-    def __call__(self, t: ScalarLike) -> Array:  # noqa: ARG002
-        return self.array
-
```
pierreguilmin, line 307: Could you re-order the method `__call__` as before in all subclasses? (to see the changes in git + to match base class ordering).

pierreguilmin (changes requested):
I'm very happy with this PR!! Beautiful small tricks here and there (the reverse enumerate 🚶‍♂️, the parametrize for the batching test 🤌), congrats! A few minor comments only.

I haven't looked in depth at the `time_array.py` modifications yet bc the changes are all mixed up due to reorderings (see comments). It'd be helpful if you could re-order as before 🙏

So what's left to do in the backend? Just the ND batching of `PWCTimeArray`?

dynamiqs/time_array.py
abocquet: It's essentially free to implement, I'll do it in next PR

dynamiqs/time_array.py
abocquet: The base class has `__call__` after all non `__` methods, hence the reordering

tests/sesolve/test_batching.py
abocquet: No, it's a reminiscence from darker times

tests/test_time_array.py
abocquet: No, because the shape of the constant array is not legal (size `(n,)` instead of `(n,n)`), so it breaks tests.

dynamiqs/core/_utils.py
abocquet: I'm not sure what you mean but I updated the comment.

tests/test_time_array.py
pierreguilmin: Ha ok. Maybe we should make these time arrays more general some day, but 👌 for now.

docs/tutorials/time-dependent-operators.md
```diff
@@ -173,9 +173,9 @@ Array([[0.+0.j, 1.+0.j],
 ??? Note "Function with additional arguments"
     To define a modulated time-array with additional arguments, you can use the optional `args` parameter of [`dq.modulated()`][dynamiqs.modulated]:
     ```python
-    f = lambda t, omega: jnp.cos(omega * t)
+    f = lambda t: jnp.cos(omegas * t)
     omega = 1.0
-    H = dq.modulated(f, dq.sigmax(), args=(omega,))
+    H = dq.modulated(f, dq.sigmax())
```
gautierronan, line 178: The first line is now wrong. I guess we can explain how to use a `functools.partial` for optional arguments in this short paragraph.

docs/tutorials/time-dependent-operators.md
```diff
@@ -218,9 +218,9 @@ Array([[1., 0.],
 ??? Note "Function with additional arguments"
     To define a callable time-array with additional arguments, you can use the optional `args` parameter of [`dq.timecallable()`][dynamiqs.timecallable]:
     ```python
-    f = lambda t, x: x * jnp.array([[t, 0], [0, 1 - t]])
     x = 1.0
-    H = dq.timecallable(f, args=(x,))
+    f = lambda t: x * jnp.array([[t, 0], [0, 1 - t]])
+    H = dq.timecallable(f)
```
gautierronan, line 223: Same comment here.

dynamiqs/time_array.py
```diff
@@ -245,6 +241,14 @@ def conj(self) -> TimeArray:
             New time-array object with element-wise complex conjuguated values.
         """
 
+    @abstractmethod
+    def as_batched_callable(self) -> callable[[float], Array]:
+        """Returns the time array under the form of a callable that is vmappable.
```
gautierronan, line 246: Should the function be called `as_vmappable_callable` instead? Maybe more explicit seeing the one-liner below.

dynamiqs/mesolve/mesolve.py
```diff
@@ -129,22 +124,22 @@ def _vmap_mesolve(
...
+        0,
+        0,
     )
 
     # the result is vectorized over `_saved` and `infos`
     out_axes = MEResult(None, None, None, None, 0, 0)
 
     # compute vectorized function with given batching strategy
-    f = compute_vmap(_mesolve, options.cartesian_batching, is_batched, out_axes)
+    f = compute_vmap(_mesolve, options.cartesian_batching, n_batch, out_axes)
```
gautierronan, line 142: I guess we're missing a call to `as_batched_callable` for `H` and `jump_ops` somewhere before this? This is what's done for `sesolve`.

dynamiqs/time_array.py
```diff
@@ -494,7 +503,10 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
 
 
 class SummedTimeArray(TimeArray):
-    timearrays: list[TimeArray]
+    timearrays: list[TimeArray] = eqx.field(static=True)
```
gautierronan, line 506: Is this required? This does not make all the underlying time arrays also static?

dynamiqs/time_array.py
```diff
@@ -535,3 +548,29 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
...
             return NotImplemented
+
+
+def eval_timearrays_sum(timearrays: [TimeArray], t: float) -> TimeArray:
+    return jtu.tree_reduce(
+        jnp.add,
+        jtu.tree_map(
+            lambda x: x(t), timearrays, is_leaf=lambda x: isinstance(x, TimeArray)
+        ),
+    )
```
gautierronan, line 559: Why not move this function directly within the `__call__` method? Is it used somewhere else?

gautierronan (changes requested):
Sorry for the late review, this is great work. I have a few comments below, and also one higher-level comment.

In the PR, modulated and callable time arrays are converted to `BatchedCallable` when entering `mesolve`/`sesolve`, while constant and pwc time arrays stay as `TimeArray`. Should we instead integrate the logic of this `BatchedCallable` class directly within `ModulatedTimeArray` and `CallableTimeArray`? This would also allow for users to directly `vmap` over arbitrary time arrays without having to bother calling the `as_batched_callable` method.

I believe a simple way to achieve this would be to make `self.f` into a `BatchedCallable` within the `__init__` of `ModulatedTimeArray` and `CallableTimeArray`. However, this will also require the `ModulatedTimeArray.array` to be expanded to the same number of dimensions as `self.f`.

docs/tutorials/time-dependent-operators.md
pierreguilmin: I'll do it in #555.

dynamiqs/time_array.py
abocquet: It used to but not anymore, I'll fix it

dynamiqs/time_array.py
abocquet: Yes it's a vestige from the past, I shoud get rid of it

dynamiqs/sesolve/sesolve.py
```diff
@@ -112,21 +107,16 @@ def _vmap_sesolve(
...
-        False,
-    )
+
+    n_batch = (H.ndim - 2, psi0.ndim - 2, 0, 0, 0, 0, 0)
 
     # the result is vectorized over `_saved` and `infos`
     out_axes = SEResult(None, None, None, None, 0, 0)
 
+    if H.ndim > 2:
+        H = H.as_batched_callable()
```
pierreguilmin, line 117: Why do we need this here but not in `mesolve`?

pierreguilmin (commented):
Still need to review time_array.py

pierreguilmin:
See #557 (on which we can work independently, it's really an addition to this PR).

pierreguilmin:
I pushed a few more nits. I hope this doesn't conflict too much with your work @abocquet. 🙇‍♂️ I'll wait for answers to Ronan's comment to finish my review (I reviewed everything up to the line `class SummedTimeArray(TimeArray)` until the end of the file).

pierreguilmin:
Tell me if I'm wrong, but I think with the current code we loose the information on the type of time array within the solvers? That's an issue.

abocquet:
> I pushed a few more nits. I hope this doesn't conflict too much with your work @abocquet.
I'd prefer if you made PRs like https://github.com/dynamiqs/dynamiqs/pull/557 next time 🙏 

> Tell me if I'm wrong, but I think with the current code we loose the information on the type of time array within the solvers? That's an issue.
We can have `{Constant,PWC,Modulated,Callable}BatchedArray` classes to identify them, but let's do this later, there is no rush

pierreguilmin:
> I'd prefer if you made PRs like https://github.com/dynamiqs/dynamiqs/pull/557 next time 🙏
Yep, and I didn't see but I broke your PR, do you want me to
a) anyway keep it like this
b) revert all commits
c) remove all commits (rebase, force push)

> We can have `{Constant,PWC,Modulated,Callable}BatchedArray` classes to identify them, but let's do this later, there is no rush

I'm confused why the `Propagator` solver tests don't fail.

abocquet:
No worries, next time is fine 👍

pierreguilmin:
So I guess my commit 7a160e48d3652f0979778a27f613035d5f196b07 introduced a bug right?
Screenshot 2024-04-25 at 16 20 43

It's a bit sad that we can't vmap on `values` and have to broadcast `times` and `array` as well, even if it doesn't really change anything perf-wise.

abocquet:
Yes, please explain the rational behind https://github.com/dynamiqs/dynamiqs/pull/552/commits/7a160e48d3652f0979778a27f613035d5f196b07

pierreguilmin:
> Yes, please explain the rational behind https://github.com/dynamiqs/dynamiqs/commit/7a160e48d3652f0979778a27f613035d5f196b07

I just wanted to remove changes related to having arbitrary batching, and mistakenly changed the `as_batched_callable` by reading too fast.

pierreguilmin:
But maybe there's a better way than this `BatchedArray` logic, and having to make everything vmap-compatible in the returned PyTree, and changing the object type. See proposition in https://linear.app/dynamiqs/issue/DYN-197/fix-new-nd-batching-feature comment
> To make it work with time arrays (which is a bit tricky for summed arrays and pwc/modulated), I think the simplest would be that they have a « compute_in_axes » like function that returns the PyTree we need to pass to vmap.

Would that work @abocquet?

abocquet:
> I just wanted to remove changes related to having arbitrary batching for PWCTimeArray

Yes, but why?

abocquet:
> To make it work with time arrays (which is a bit tricky for summed arrays and pwc/modulated), I think the simplest would be that they have a « compute_in_axes » like function that returns the PyTree we need to pass to vmap.

This Pytree is what is returned but `as_vmappable_callable`

pierreguilmin:
> This Pytree is what is returned but as_vmappable_callable

Sorry, I meant the PyTree to use as argument for `in_axes` of vmap.

pierreguilmin:
> Sorry, I meant the PyTree to use as argument for in_axes of vmap.

So every `TimeArray` would natively be a `BatchedArray`, and we would inform vmap of the structure of the batching for each `TimeArray`, by specifying the correct `in_axes` argument.

abocquet:
> So every TimeArray would natively be a BatchedArray, and we would inform vmap of the structure of the batching for each TimeArray, by specifying the correct in_axes argument.

That would require nested `in_axes`. I don't think it is possible. I'll check

gautierronan:
> > Tell me if I'm wrong, but I think with the current code we loose the information on the type of time array within the solvers? That's an issue.
> 
> We can have `{Constant,PWC,Modulated,Callable}BatchedArray` classes to identify them, but let's do this later, there is no rush

Or we could transform only self.f into a BatchedCallable? This way we keep the initial TimeArray PyTree structure, while having all attributes vmappable.

On the subject of the in_axes, I agree with Pierre's solution. This is somewhat similar to what we had before with the compute_in_axes function.

abocquet:
>Or we could transform only self.f into a BatchedCallable? This way we keep the initial TimeArray PyTree structure, while having all attributes vmappable.

`self.f` does not exist for PWC time arrays

abocquet:
I'll revamp the PR, please don't add comments anymore for the moment

pierreguilmin:
My bad for https://github.com/dynamiqs/dynamiqs/commit/7a160e48d3652f0979778a27f613035d5f196b07, the whole commit should stay bc of the current structure (i.e. we need to create a fully vmap-pable structure, thus have arbitrary shape `times` and `array`, which we just be broadcasted to `values` in the end when building the batched array). I thought these bits were rests from a pre-matured implementation of DYN-204 et DYN-205. So yep, ideally specifying the `in_axes` would allow not doing this. 🤞

gautierronan:
> >Or we could transform only self.f into a BatchedCallable? This way we keep the initial TimeArray PyTree structure, while having all attributes vmappable.
> 
> `self.f` does not exist for PWC time arrays

Only for modulated and callable. For the rest it should not be required to use this self.indices trick.

PR #554, opened 2024-04-19
Add matmul_precision option for dq.set_precision

gautierronan:


dynamiqs/utils/jax_utils.py
```diff
@@ -132,3 +140,15 @@ def set_precision(precision: Literal['simple', 'double']):
             f"Argument `x` should be a string 'simple' or 'double', but is"
             f" '{precision}'."
         )
+
+    if matmul_precision == 'low':
+        jax.config.update('jax_default_matmul_precision', 'bfloat16')
+    elif matmul_precision == 'high':
+        jax.config.update('jax_default_matmul_precision', 'bfloat16_3x')
+    elif matmul_precision == 'highest':
+        jax.config.update('jax_default_matmul_precision', 'float32')
```
abocquet, line 149: It seems strange that we cannot go to float64 for matrix product no? What is the point of f64 then?

gautierronan: We can, it's just that this API is very poorly written from Jax. Float32 actually means float32 or float64 depending on the global precision. Figure out! 🤷

pierreguilmin (changes requested):
Great addition!!

I think it'd be better to have two different functions `set_precision` and `matmul_precision`. It will make it easier to change these parameters independently (e.g. in a loop). `matmul_precision` is also more niche than `set_precision` which would be used pretty often, it's good to keep it simple.

Another option would be to have a `dq.set(device=?, precision=?, matmul_precision=?)` with only optional arguments, but the doc would get a bit crazy, so I'd prefer the current version (with one function per option).

dynamiqs/utils/jax_utils.py
abocquet: We should comment that somewhere

pierreguilmin (approved):
Very nice!! I've added commits with a few suggestions.

dynamiqs/utils/jax_utils.py
pierreguilmin: See https://jax.readthedocs.io/en/latest/jax.lax.html#jax.lax.Precision, I've switched the key to `'highest'` which is not ambiguous.

PR #555, opened 2024-04-24
Document new ND batching feature and new time-array API

pierreguilmin:
Closes DYN-73 and DYN-197.

On top of #552. I recommend reading commit by commit.

Here are a few screenshots of the main bits for convenience:

**MEResult documentation**

The note is folded by default.
Screenshot 2024-04-24 at 14 46 08

**Time-dependent operators documentation**
The note is folded by default.
Screenshot 2024-04-24 at 14 47 07

I propose to revamp the "Batching simulations" tutorial in another PR (there's a bit of work, we need to explain cartesian vs non-cartesian batching, nd batching and broadcasting) > https://linear.app/dynamiqs/issue/DYN-206/rework-batching-simulations-tutorial.

### DIFF
--- a/docs/tutorials/time-dependent-operators.md
+++ b/docs/tutorials/time-dependent-operators.md
@@ -213,7 +213,6 @@ Array([[1., 0.],
 ```
 
 !!! Warning "The function `f` must return a JAX array (not an array-like object!)"
-    An error is raised if the function `f` does not return a JAX array. This error includes other array-like objects. This is enforced to avoid costly conversions at every time step of the numerical integration.
 
 ??? Note "Function with additional arguments"
     To define a callable time-array with additional arguments, you can use the optional `args` parameter of [`dq.timecallable()`][dynamiqs.timecallable]:
@@ -221,6 +220,7 @@ Array([[1., 0.],
     x = 1.0
     f = lambda t: x * jnp.array([[t, 0], [0, 1 - t]])
     H = dq.timecallable(f)
+    An error is raised if the function `f` does not return a JAX array. This error concerns any other array-like objects. This is enforced to avoid costly conversions at every time step of the numerical integration.
     ```
 
 ## Batching and differentiation with time-arrays
--- a/docs/tutorials/time-dependent-operators.md
+++ b/docs/tutorials/time-dependent-operators.md
@@ -145,7 +145,7 @@ where $f(t)$ is an time-dependent scalar.
 
 In dynamiqs, modulated operators are defined by:
 
-- `f`: a Python function with signature `f(t: float, *args: PyTree) -> Array` that returns the modulating factor $f(t)$ for any time $t$, as an array of shape _(...)_,
+- `f`: a Python function with signature `f(t: float) -> Array` that returns the modulating factor $f(t)$ for any time $t$, as an array of shape _(...)_,
 - `array`: the array defining the constant operator $O_0$, of shape _(n, n)_.
 
 To construct a modulated operator, pass these two arguments to the [`dq.modulated()`][dynamiqs.modulated] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. This object then returns an array with shape _(..., n, n)_ when called at any time $t$.
@@ -188,7 +188,7 @@ where $f(t)$ is a time-dependent operator.
 
 In dynamiqs, arbitrary time-dependent operators are defined by:
 
-- `f`: a Python function with signature `f(t: float, *args: PyTree) -> Array` that returns the operator $f(t)$ for any time $t$, as an array of shape _(..., n, n)_.
+- `f`: a Python function with signature `f(t: float) -> Array` that returns the operator $f(t)$ for any time $t$, as an array of shape _(..., n, n)_.
 
 To construct an arbitrary time-dependent operator, pass this argument to the [`dq.timecallable()`][dynamiqs.timecallable] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. This object then returns an array with shape _(..., n, n)_ when called at any time $t$.
 
--- a/docs/tutorials/time-dependent-operators.md
+++ b/docs/tutorials/time-dependent-operators.md
@@ -96,7 +96,7 @@ In dynamiqs, PWC operators are defined by three array-like objects:
 - `values`: the constant values $(c_0, \ldots, c_{N-1})$ for each time interval, of shape _(..., N)_,
 - `array`: the array defining the constant operator $O_0$, of shape _(n, n)_.
 
-To construct a PWC operator, pass these three arguments to the [`dq.pwc()`][dynamiqs.pwc] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object.
+To construct a PWC operator, pass these three arguments to the [`dq.pwc()`][dynamiqs.pwc] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. This object then returns an array with shape _(..., n, n)_ when called at any time $t$.
 
 !!! Notes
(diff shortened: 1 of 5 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -182,11 +182,11 @@ def _check_mesolve_args(
     H: TimeArray, jump_ops: list[TimeArray], rho0: Array, exp_ops: Array | None
 ):
     # === check H shape
-    check_shape(H, 'H', '(..., n, n)')
+    check_shape(H, 'H', '(..., n, n)', subs={'...': '...H'})
```
abocquet, line 185: "H must be of shape `...H`" sounds a bit cryptic, even though I see what you mean. I would replace with `(..., n, n)`. (PS: I don't have the "suggesion option anymore in GH interface)

dynamiqs/mesolve/mesolve.py
```diff
@@ -182,11 +182,11 @@ def _check_mesolve_args(
...
 
-    # === check jump_ops
+    # === check jump_ops shape
     for i, L in enumerate(jump_ops):
-        check_shape(L, f'jump_ops[{i}]', '(..., n, n)')
+        check_shape(L, f'jump_ops[{i}]', '(..., n, n)', subs={'...': f'...L{i}'})
```
abocquet, line 189: Same comment

dynamiqs/mesolve/mesolve.py
```diff
@@ -195,7 +195,7 @@ def _check_mesolve_args(
         )
 
     # === check rho0 shape
-    check_shape(rho0, 'rho0', '(..., n, 1)', '(..., n, n)', subs={'...': '...nrho0'})
+    check_shape(rho0, 'rho0', '(..., n, 1)', '(..., n, n)', subs={'...': '...rho0'})
```
abocquet, line 198: Same

dynamiqs/sesolve/sesolve.py
```diff
@@ -156,8 +156,10 @@ def _sesolve(
 
 def _check_sesolve_args(H: TimeArray, psi0: Array, exp_ops: Array | None):
     # === check H shape
-    check_shape(H, 'H', '(..., n, n)')
-    check_shape(psi0, 'psi0', '(..., n, 1)', subs={'...': '...npsi0?'})
+    check_shape(H, 'H', '(..., n, n)', subs={'...': '...H'})
```
abocquet, line 159: Same

dynamiqs/sesolve/sesolve.py
```diff
@@ -156,8 +156,10 @@ def _sesolve(
...
+
+    # === check psi0 shape
+    check_shape(psi0, 'psi0', '(..., n, 1)', subs={'...': '...psi0'})
```
abocquet, line 162: Same

abocquet (changes requested):
Thanks for these nice additions 👍 I would be down to explicitely indicate the shape we are expecting for all objects (see comments) but aside from that LGTM 💯

dynamiqs/mesolve/mesolve.py
gautierronan: I quite like `...H` on my end! I think it's quite explicit in the sense that I can't find another interpretation for it, so people should understand.

gautierronan: Could be useful to add a short sentence about what `...H` and `...psi0` mean in the "Running multiple simulations concurrently" box above?

docs/tutorials/time-dependent-operators.md
```diff
@@ -213,23 +227,14 @@ Array([[1., 0.],
...
-??? Note "Function with additional arguments"
-    To define a callable time-array with additional arguments, you can use the optional `args` parameter of [`dq.timecallable()`][dynamiqs.timecallable]:
-    ```python
-    x = 1.0
-    f = lambda t: x * jnp.array([[t, 0], [0, 1 - t]])
-    H = dq.timecallable(f)
+    An error is raised if the function `f` does not return a JAX array. This error concerns any other array-like objects. This is enforced to avoid costly conversions at every time step of the numerical integration.
+
+??? Notes "Batching"
+    The batching of the returned time-array is specified by the array returned by `f`. For example, to define am arbitrary time-dependent operator batched over a parameter $\theta$:
```
gautierronan, line 233:
```suggestion
    The batching of the returned time-array is specified by the array returned by `f`. For example, to define an arbitrary time-dependent operator batched over a parameter $\theta$:
```
typo

gautierronan (changes requested):
Looks good. A few nits below.

Otherwise we'll indeed need to rework the "Batching simulations" tutorial, in particular to explain the `jnp.vectorize` tricks and provide an overall workflow of how to work with batching on a day-to-day basis.

dynamiqs/mesolve/mesolve.py
pierreguilmin: @abocquet the `subs` argument only replaces the dictionnary keys substring, not the full shape specification. The resulting error message would read
```python
ValueError: Argument `H` must have shape (...H, n, n), but has shape H.shape=(2,).
```

You can test with e.g.
```python
from dynamiqs._checks import check_shape
import jax.numpy as jnp
H = jnp.array([1, 2])
check_shape(H, 'H', '(..., n, n)', subs={'...': '...H'})
```

pierreguilmin: > Could be useful to add a short sentence about what ...H and ...psi0 mean in the "Running multiple simulations concurrently" box above?

I'm assuming that people know what we mean by this notation (we use it everywhere e.g. in the utils). I think if we want to document it, we should ideally do it in a single place, not in every functions. But IMO it's pretty self-explanatory.

dynamiqs/result.py
```diff
@@ -89,34 +89,97 @@ def to_numpy(self) -> Result:
...
+            if specified by `exp_ops`.
         extra _(PyTree or None)_: Extra data saved with `save_extra()` if
             specified in `options`.
         infos _(PyTree or None)_: Solver-dependent information on the resolution.
         tsave _(array of shape (ntsave,))_: Times for which results were saved.
         solver _(Solver)_: Solver used.
         gradient _(Gradient)_: Gradient used.
         options _(Options)_: Options used.
+
+    Notes-: Result of running multiple simulations concurrently
```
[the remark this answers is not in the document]
pierreguilmin: I don't have a strong opinion either, but I'm slightly in favour of keeping it folded. I consider batching a bit advanced, and this makes the result page quite scary at first look for beginners. 🤷‍♂️

docs/tutorials/time-dependent-operators.md
```diff
@@ -192,6 +192,17 @@ Array([[0.+0.j, 1.+0.j],
...
     ```
 
+??? Notes "Function with additional arguments"
+    To define a modulated operator with a function that takes arguments other than time (extra `*args` and `**kwargs`), you can use [`functools.partial()`](https://docs.python.org/3/library/functools.html#functools.partial). For example:
+    ```pycon
+    >>> import functools
+    >>> def pulse(t, omega, amplitude=1.0):
+    ...     return amplitude * jnp.cos(omega * t)
+    >>> # create function with correct signature (t: float) -> Array
+    >>> f = functools.partial(pulse, 1.0, amplitude=5.0)
```
gautierronan, line 202:
```suggestion
    >>> f = functools.partial(pulse, omega=1.0, amplitude=5.0)
```
otherwise the partial goes over time.

gautierronan (approved):

docs/tutorials/time-dependent-operators.md
pierreguilmin: Wouuups, good catch 😅

PR #557, opened 2024-04-25
Split carthesian and non carthesian batching

pierreguilmin:
One functionality is missing with the current code: broadcasting in the case of non-cartesian batching. I propose a refactoring that
1. Clearly splits both cartesian and non-cartesian batching,
2. Replace all names/comments related to "vmap" by "vectorize" (because it's what we do, really),
3. Document clearly the functionalities.

Practically, the logic hasn't changed at all, `compute_vmap` has just be splitted upon `_flat_vectorize` and `_cartesian_vectorize`. I expect this part of the code to make many people curious, I think it's worth spending time making the logic crystal clear.

I didn't fix the broadcasting, but it should not be too hard, @abocquet do you want to commit on top of this PR (i.e. fix `_flat_vectorize()`? We also need to catch this in our tests, because today we only test
```python
npsi0 = npsi0 if cartesian_batching else nH
```

Example code that breaks today:
```python
import dynamiqs as dq
import jax.numpy as jnp

n = 2
H = jnp.ones((3, 4, n, n))
psi0 = jnp.ones((4, n, 1))
result = dq.sesolve(H, psi0, tsave, options=dq.Options(cartesian_batching=False))
```
IMO this code should work for any broadcastable `psi0` shape: `(n, n)`, `(4, n, n)`, `(3, 1, n, n)`, `(1, 4, n, n)` and any `(...psi0, 1, 1, n, n)`.

### DIFF
--- a/dynamiqs/core/_utils.py
+++ b/dynamiqs/core/_utils.py
@@ -38,44 +38,97 @@ def get_solver_class(
     return solvers[type(solver)]
 
 
-def compute_vmap(
-    f: callable,
-    cartesian_batching: bool,
-    n_batch: PyTree[int],
-    out_axes: PyTree[int | None],
+def _flat_vectorize(
+    f: callable, n_batch: PyTree[int], out_axes: PyTree[int | None]
 ) -> callable:
-    # This function vectorizes `f` by applying jax.vmap over batched dimensions. The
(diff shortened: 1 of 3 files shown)


abocquet:
@gautierronan since this is refactoring and documentation and it goes into another PR, I'll merge now but we can edit things later if needed

PR #558, opened 2024-04-29
Add SparseDIA Class

bsaoptima:
Hi there,

This PR is for the addition of the custom SparseDIA class ot the dynamiqs codebase. This format is fit for the sparse nature of the matrices used by our users. It is coded in JAX and handle the Just-In-Time compilation built-in system.

We only care about the various diagonals and associated offsets of a matrix in the SparseDIA format, which allows us to perform computations in an efficient manner.

Here is an overview of the methods I have created in this class:
- Multiplication & Addition between a SparseDIA matrix and a "dense" format matrix (both directions).
- Multiplication between two matrices that are both in SparseDIA format.
- Turning a SparseDIA matrix into a "dense" format.
- Turning a "dense" format matrix into the SparseDIA format.


dynamiqs/gradient.py
```diff
@@ -21,6 +21,10 @@ def __init__(self):
             [`diffrax.DirectAdjoint`](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.DirectAdjoint)
             option.
         """
+        super().__init__()
+
+    def __str__(self) -> str:
+        return eqx.tree_pformat(self, indent=0).replace('\n', '').replace(',', ', ')
```
gautierronan, line 27: I guess this is from another PR?

dynamiqs/solver.py
```diff
@@ -27,6 +27,9 @@ def assert_supports_gradient(cls, gradient: Gradient | None) -> None:  # noqa: A
                 f' {support_str}).'
             )
 
+    def __str__(self) -> str:
+        return eqx.tree_pformat(self, indent=0).replace('\n', '').replace(',', ', ')
+
```
gautierronan, line 32: Same here.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+        # there are 2N-1 offsets in a NxN matrix
+        offset_range = 2 * matrix.shape[0] - 1
+        offset_center = offset_range // 2
+
+        for offset in range(-offset_center, offset_center + 1):
+            diagonal = jnp.diagonal(matrix, offset=offset)
+            if jnp.any(diagonal != 0):
+                diagonals_with_offsets[offset] = diagonal
+
+        return diagonals_with_offsets
```
gautierronan: This function is a class creator, so it should be independent from the class. In other words, the `self` object is never used.

gautierronan: It should return an instance of `SparseDIA`.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+    @functools.partial(jax.jit, static_argnums=(0,1))
+    def _matmul(
```
gautierronan, line 25: This is a matmul against dense, so maybe we can rename it `_matmul_dense` to make it clear.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+                    diag[start:end, None] * matrix[start:end, :]
+                )
+            else:
+                out = out.at[:, start:end].add(
+                    jnp.transpose(diag[start:end, None] * jnp.transpose(matrix[:, top:bottom]))
+                )
+        return out
+    
+    @functools.partial(jax.jit, static_argnums=(0,1))
+    def _diamul(self, matrix) -> Array:
```
gautierronan, line 49: Same here, I guess we can rename `_matmul_dia`.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+
+            mask = jnp.any(diags != 0, axis=1)
+            diags = diags[mask]
+            offsets = offsets[mask]
+
+            unique_offsets, indices = jnp.unique(offsets, return_inverse=True)
+            result = jnp.zeros((len(unique_offsets), diags.shape[1]))
+
+            for i, offset in enumerate(unique_offsets):
+                result = result.at[i].set(jnp.sum(diags[indices == i], axis=0))
```
gautierronan, line 139: For code clarity, this should be moved in a separate function. This way `self.__matmul__` is just about dispatching to appropriate methods.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+                offset_list.append(new_offset)
+        
+        out = jnp.vstack(vector_list)
+        
+        return out, offset_list 
+
+        
+    
+
+    def _dag(self):
```
gautierronan, line 99: I think this should rather be `def dag(self) -> SparseDIA` and return an instance of the hermitian instance of the class.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+        """Returns the hermitian conjugate, call to_dense() to visualize."""
+        self.offsets = tuple(-1 * jnp.array(self.offsets))
+        self.diags = jnp.conjugate(self.diags)
+
+    @functools.partial(jax.jit, static_argnums=(0,))
+    def _add(self, matrix: Array) -> Array:
```
gautierronan, line 105: Same comment as before about the naming.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+
+        
+    
+
+    def _dag(self):
+        """Returns the hermitian conjugate, call to_dense() to visualize."""
+        self.offsets = tuple(-1 * jnp.array(self.offsets))
+        self.diags = jnp.conjugate(self.diags)
+
+    @functools.partial(jax.jit, static_argnums=(0,))
```
gautierronan, line 104: I think we don't need to add `jax.jit` calls inside the class. But the user will wrap his code inside `jax.jit`, so all functions should be compatible.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+            start = max(0, abs(offset))
+            end = min(N, N + abs(offset))
+            i = jnp.arange(end - start) if offset > 0 else jnp.arange(start, end)
+            j = jnp.arange(start, end) if offset > 0 else jnp.arange(end - start)
+
+            matrix = matrix.at[i, j].add(diag[start:end])
+
+        return matrix
+
+    def __matmul__(self, matrix):
```
gautierronan, line 119: A `else` is required at the end, to raise an error in case the type of `matrix` is not recognized.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+
+            
+            return SparseDIA(result, tuple([offset.item() for offset in unique_offsets]))
+        
+    def __rmatmul__(self, matrix):
+        if isinstance(matrix, Array):
+            return self._matmul(direction="right", matrix=matrix)
+        
+    def __getitem__(self, index):
+        dense = self.to_dense()
```
gautierronan, line 149: No need to convert to dense here. We can just fetch the appropriate value by looking at the offset diagonal where `index` is.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+                result = result.at[i].set(jnp.sum(diags[indices == i], axis=0))
+
+            
+            return SparseDIA(result, tuple([offset.item() for offset in unique_offsets]))
+        
+    def __rmatmul__(self, matrix):
+        if isinstance(matrix, Array):
+            return self._matmul(direction="right", matrix=matrix)
+        
+    def __getitem__(self, index):
```
gautierronan, line 148: `__getitem__` should support slicing. So we should be able to write these kinds of things:
 - `sparsemat[3:7, 2]`
 - `sparsemat[3:7, 2:4]`
 - `sparsemat[3:7, :]`
 - `sparsemat[..., 2:4]`
 - `sparsemat[None, :, :]`

Maybe the last one is a bit harder though.

Note that all of these might end being quite complicated to implement. Maybe we should remove `__getitem__` for now and implement it properly in a subsequent PR.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,149 @@
...
+        dense = self.to_dense()
+        return dense[index]
+    
+    def __add__(self, matrix):
```
gautierronan, line 152: Similarly as before, needs a `else`.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,180 @@
...
+            start = max(0, abs(offset))
+            end = min(N, N + abs(offset))
+            i = jnp.arange(end - start) if offset > 0 else jnp.arange(start, end)
+            j = jnp.arange(start, end) if offset > 0 else jnp.arange(end - start)
+
+            matrix = matrix.at[i, j].add(diag[start:end])
+
+        return matrix
+
+    """ DUNDER METHODS """
```
gautierronan, line 118: This kind of comment can be removed. It's not needed to understand the code.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,180 @@
...
+        # there are 2N-1 offsets in a NxN matrix
+        offset_range = 2 * matrix.shape[0] - 1
+        offset_center = offset_range // 2
+
+        for offset in range(-offset_center, offset_center + 1):
+            diagonal = jnp.diagonal(matrix, offset=offset)
+            if jnp.any(diagonal != 0):
+                diagonals_with_offsets[offset] = diagonal
+
+        return diagonals_with_offsets
```
gautierronan, line 180: This should return an instance of `SparseDIA`.

gautierronan (changes requested):
Nice! Here's a set of initial comments on the overall class structure. I think we're mainly missing a set of tests for core operations. Once we have that, I'll review the actual operations logic.

dynamiqs/sparse.py
bsaoptima: I'm not sure this is 100% feasible, some functions (like the cleanup one after a matmul between 2 SparseDIA matrices) are not `jax.jit` compatible. Furthermore, `jax.jit` needs to know the size and dtype of the input array, thus if the input is a SparseDIA class, we should probably have a think on how to make this possible.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,202 @@
...
+            j = jnp.arange(s, e) if offset > 0 else jnp.arange(e - s)
+
+            matrix = matrix.at[i, j].add(diag[start:end])
+
+        return matrix
+    
+
+    def _cleanup(self, diags, offsets):
+        diags = jnp.array(diags)
+        offsets = jnp.array(offsets)
```
gautierronan, line 122:
```suggestion
        diags = jnp.asarray(diags)
        offsets = jnp.asarray(offsets)
```

dynamiqs/sparse.py
```diff
@@ -0,0 +1,202 @@
...
+            out += jnp.diag(diag[start:end], k=offset)
+        return out
+
+    @functools.partial(jax.jit, static_argnums=(0,1))
+    def _matmul_dense(
+        self,
+        direction: str,
+        matrix: Array,  # If matrix is a vector need to input as column vector.
+    ) -> Array:
+        """Sparse @ Dense only using information about diagonals & offsets."""
```
gautierronan, line 30: No need to document private functions. If there's something non-trivial to be said about the function, or the way it works, then you can add it through in-line comment (`# some comment`).

tests/test_sparse.py
```diff
@@ -0,0 +1,45 @@
+"""
+Test Functions for the SparseDIA class
+"""
```
gautierronan, line 3:
```suggestion
```

tests/test_sparse.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class TestSparseDIA:
+
+    def test_matmul(self, matrixA, matrixB):
+        sparseA = to_sparse(matrixA)
+        sparseB = to_sparse(matrixB)
+
+        out_dia_dia = sparseA @ sparseB
+        out_dia_dense = sparseA @ matrixB
```
gautierronan, line 18: Missing right-hand-side multiplication, `matrixA @ sparseB`.

tests/test_sparse.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            print("Results not matching")
+
+    def test_add(self, matrixA, matrixB):
+
+        sparseA = to_sparse(matrixA)
+        sparseB = to_sparse(matrixB)
+
+        out_dia_dia = sparseA + sparseB
+        out_dia_dense = sparseA + matrixB
```
gautierronan, line 32: Same here.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,202 @@
...
+
+            i = jnp.arange(e - s) if offset > 0 else jnp.arange(s, e)
+            j = jnp.arange(s, e) if offset > 0 else jnp.arange(e - s)
+
+            matrix = matrix.at[i, j].add(diag[start:end])
+
+        return matrix
+    
+
+    def _cleanup(self, diags, offsets):
```
gautierronan, line 120: I think we should also add a test for this function, since it has non-straightforward logic.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,202 @@
...
+from jaxtyping import Array
+
+
+class SparseDIA:
+    def __init__(self, diags: Array, offsets: tuple[int]):
+        self.offsets = offsets
+        self.diags = diags
+
+
+    def to_dense(self) -> Array:
```
gautierronan, line 14: I think we want to duplicate this function to have a `dq.to_dense` in the public API (similarly to `dq.to_sparse`). So it's essentially the same as this one, but not as a class method. Both should have the same content, so you can simply redirect `SparseDIA.to_dense` to `dq.to_dense`.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,202 @@
...
+                out = out.at[top:bottom, :].add(
+                    diag[start:end, None] * matrix[start:end, :]
+                )
+            else:
+                out = out.at[:, start:end].add(
+                    jnp.transpose(diag[start:end, None] * jnp.transpose(matrix[:, top:bottom]))
+                )
+        return out
+
+    @functools.partial(jax.jit, static_argnums=(0,1))
```
gautierronan, line 48: I believe the proper way to handle the JIT is to have the class `SparseDIA` inherit from `eqx.Module` (see [Equinox](https://docs.kidger.site/equinox/)) and to make `self.offsets` a static field with `offsets = eqx.field(static=True)`.

This way, both `self` and `matrix` will be dynamic argnums (not static). So you can fully remove the decorator here.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,202 @@
...
+        for offset, diag in zip(self.offsets, self.diags):
+            start = max(0, offset)
+            end = min(N, N + offset)
+            out += jnp.diag(diag[start:end], k=offset)
+        return out
+
+    @functools.partial(jax.jit, static_argnums=(0,1))
+    def _matmul_dense(
+        self,
+        direction: str,
```
gautierronan, line 27: This is binary, so it should be a boolean. Something like:
```suggestion
        left_matmul: bool,
```

gautierronan (changes requested):
Second small round of review. I haven't looked at the actual logic of functions yet, only at overall formatting.

Apart from the comments below, I think we're missing:
 - `self.__mul__` and `self.__rmul__`, 
 - function typing all around,
 - formatting (simply run `ruff format` locally I think, or `task all`),
 - tests for operations under JIT.

dynamiqs/gradient.py
bsaoptima: Yes, I removed it!

dynamiqs/solver.py
bsaoptima: Removed!

tests/test_sparse.py
```diff
@@ -0,0 +1,45 @@
...
+
+from dynamiqs.sparse import *
+import jax
+import jax.numpy as jnp
+
+
+
+class TestSparseDIA:
+
+    def test_matmul(self, matrixA, matrixB):
```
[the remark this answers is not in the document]
bsaoptima: Done!

tests/test_sparse.py
```diff
@@ -0,0 +1,45 @@
...
+            print("Dia Addition works!")
+        else:
+            print("Results not matching!")
+
+    def test_transform(self, matrix):
+
+        if jnp.allclose(matrix, to_sparse(matrix).to_dense()):
+            print("Transform works!")
+        else:
+            print("Transform is not working!")
```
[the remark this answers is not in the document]
bsaoptima: Added some `assert` statements, awaiting feedback to resolve the comment.

gautierronan:
Closing in favor of https://github.com/dynamiqs/dynamiqs/pull/577

PR #560, opened 2024-05-02
Improve `dq.tracemm` documentation

gautierronan:
Small improvements to `tracemm` documentation, to make it compliant with the rest of the docs, and more readable.
 - Moves the function after `dq.trace` (because it's a specialisation)
 - Small improvements to the doc, in particular the one-liner is more compliant with other functions
 - Add negative spacing after the trace operator rendering


pierreguilmin (approved):
Neat! 🤩

PR #561, opened 2024-05-02
Batch dq.fock and dq.coherent over argument number/alpha

gautierronan:
This PR proposes to change the behavior of `dq.fock` and `dq.coherent` to accept batching over the `number` and `alpha` parameters, respectively. 

Before, we would accept tuple inputs to create multi-mode fock and coherent states. This is no longer the case, in favor of single-mode batching. Contrary to what I was advocating before, I believe this is actually way more useful on a day-to-day basis, and also more natural and compliant with the rest of the library. At least this is what I am seeing from my personal usage of the library.

If you agree with the behavior, I suggest reviewing commit by commit.

Examples:
```python
import dynamiqs as dq
import jax.numpy as jnp

print(dq.fock(3, 1))
# [[0.+0.j]
#  [1.+0.j]
#  [0.+0.j]]

print(dq.fock(3, (1, 0)))
# [[[0.+0.j]
#   [1.+0.j]
#   [0.+0.j]]
# 
#  [[1.+0.j]
#   [0.+0.j]
#   [0.+0.j]]]

alphas = jnp.linspace(0.0, 3.0, 5)
print(dq.coherent(4, alphas).shape)
# (5, 4, 1)

```

### DIFF
--- a/dynamiqs/utils/states.py
+++ b/dynamiqs/utils/states.py
@@ -15,28 +15,29 @@ __all__ = ['fock', 'fock_dm', 'basis', 'basis_dm', 'coherent', 'coherent_dm']
 
 
 def fock(dim: int, number: ArrayLike) -> Array:
-    r"""Returns the ket of a Fock state or the ket of a tensor product of Fock states.
+    r"""Returns the ket of a Fock state.
 
     Args:
-        dim _(int or tuple of ints)_: Dimension of the Hilbert space of each mode.
-        number _(int or tuple of ints)_: Fock state number of each mode.
+        dim: Dimension of the Hilbert space.
+        number _(array-like of integers of shape (...))_: Fock state number.
 
     Returns:
-        _(array of shape (n, 1))_ Ket of the Fock state or tensor product of Fock
-            states.
+        _(array of shape (..., dim, 1))_ Ket of the Fock state, or array of kets of
+            Fock states if batched.
 
     Examples:
         >>> dq.fock(3, 1)
         Array([[0.+0.j],
                [1.+0.j],
                [0.+0.j]], dtype=complex64)
-        >>> dq.fock((3, 2), (1, 0))
-        Array([[0.+0.j],
-               [0.+0.j],
-               [1.+0.j],
-               [0.+0.j],
-               [0.+0.j],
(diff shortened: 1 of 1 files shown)


abocquet (changes requested):
This is great work but I have to say that I use the multi modes style all the time so I'm not very eager to switch (until we implement `&` operator on QObj anyway 😋 )

gautierronan:
> This is great work but I have to say that I use the multi modes style all the time so I'm not very eager to switch (until we implement & operator on QObj anyway 😋 )

Ah! Didn't know, should have asked before then. Funny, I was the one pushing for the multi-mode feature initially, but I ended up needing the batching feature more haha! We can keep it for later then. We could also make a Frankenstein version that accepts both formats.

pierreguilmin:
Related to https://linear.app/dynamiqs/issue/DYN-128/change-all-operators-api-to-handle-multiple-dimensions.

Let's support both! It would unify the APIs of fock / coherent and displace / number / parity / etc. (see DYN-128). I use both all the time. 😇

```diff
- def fock(dim: int | tuple[int, ...], number: int | tuple[int, ...]) -> Array:
+ def fock(dim: int | tuple[int, ...], number: ArrayLike) -> Array:

- def displace(dim: int, alpha: ArrayLike) -> Array:
+ def displace(dim: int | tuple[int, ...], alpha: ArrayLike) -> Array:
```

where
- if `dim` is `int`, `number/alpha` must have shape `(...)`
- if `dim` is `tuple[int]` with `nd` elements, `number/alpha` must have shape `(..., nd)`  (or `(nd, ...)`?)

Ex use cases:
```python
dq.fock(na, 0)  # (na, 1)
dq.fock(na, [0, 1, 2, 3])  # (3, na, 1)
dq.fock((na, nb), (0, 1))  # (na x nb, 1)
ket00, ket01, ket10, ket11 = dq.fock((na, nb), ((0, 0), (0, 1), (1, 0), (0, 0))  # (4, na x nb, 1)
```

pierreguilmin:
Closing this for now, issue opened here > https://linear.app/dynamiqs/issue/DYN-208/handle-both-multiple-dimensions-and-batched-arguments-in.

PR #562, opened 2024-05-02
 Add new batching API for time arrays and support for ND batching

abocquet:
This PR:
i) removes the args from TimeArrays
ii) allows for vmapping on time arrays
iii) allows for vmapping on ND time arrays
iv) adds a bunch of comments (from Pierre's previous PR)
v) cleans some code (linked to the PR)

Note that support for batched times in pwc time arrays is not far, but I keep it for a next PR

Closes https://github.com/dynamiqs/dynamiqs/issues/547.
Closes and replaces https://github.com/dynamiqs/dynamiqs/pull/552.
Closes and includes https://github.com/dynamiqs/dynamiqs/pull/555.

Related to DYN-197.

Follow ups:
- [ ] Optimize `_flat_vectorize_` to avoid broadcasting


dynamiqs/sesolve/sesolve.py
```diff
@@ -38,52 +38,52 @@ def sesolve(
...
-        the [Time-dependent operators](../../tutorials/time-dependent-operators.md)
-        tutorial for more details.
-
-    Quote: Running multiple simulations concurrently
-        Both the Hamiltonian `H` and the initial state `psi0` can be batched to
-        solve multiple Schrödinger equations concurrently. All other arguments are
-        common to every batch. See the
-        [Batching simulations](../../tutorials/batching-simulations.md) tutorial for
-        more details.
+     This function computes the evolution of the state vector $\ket{\psi(t)}$ at time
```
abocquet, line 41: This is probably a problem with the formatted, I'll fix it

dynamiqs/time_array.py
```diff
@@ -227,24 +213,42 @@ def ndim(self) -> int:
...
     @abstractmethod
     def conj(self) -> TimeArray:
-        """Returns the element-wise complex conjugate of the time-array.
+        """Computes element-wise complex conjugate of the time-array.
 
         Returns:
             New time-array object with element-wise complex conjuguated values.
         """
 
+    @abstractmethod
```
gautierronan, line 246: Nit: maybe this should also be a property?

dynamiqs/time_array.py
```diff
@@ -390,48 +406,46 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
...
+        return ModulatedTimeArray(self.f, self.array.mT)
 
     def reshape(self, *new_shape: int) -> TimeArray:
-        f_shape = jax.eval_shape(self._call_f, 0.0).shape
-        new_f_shape, new_array_shape = _split_shape(
-            new_shape, f_shape, self.array.shape
-        )
-        f = Partial(lambda t, *args: self.f(t, *args).reshape(*new_f_shape))
-        return ModulatedTimeArray(f, self.array.reshape(*new_array_shape), self.args)
+        f = jtu.Partial(lambda t: self.f(t).reshape(*new_shape[:-2]))
```
gautierronan, line 426: f is no longer a BatchedCallable from here onward. I'm guessing this will cause problems when vmapping later on. I think we should thus reshape the function inside the BatchedCallable?

dynamiqs/time_array.py
```diff
@@ -390,48 +406,46 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
 
 
 class ModulatedTimeArray(TimeArray):
-    f: callable[[float, ...], Array]  # (...)
+    f: callable[[float], Array]  # (...)
```
gautierronan, line 409: Should this be a BatchedCallable instead then?

dynamiqs/time_array.py
```diff
@@ -390,48 +406,46 @@ def __add__(self, other: ArrayLike | TimeArray) -> TimeArray:
...
-        new_f_shape, new_array_shape = _split_shape(
-            new_shape, f_shape, self.array.shape
-        )
-        f = Partial(lambda t, *args: self.f(t, *args).reshape(*new_f_shape))
-        return ModulatedTimeArray(f, self.array.reshape(*new_array_shape), self.args)
+        f = jtu.Partial(lambda t: self.f(t).reshape(*new_shape[:-2]))
+        return ModulatedTimeArray(f, self.array)
+
+    def broadcast_to(self, *new_shape: int) -> TimeArray:
+        f = jtu.Partial(lambda t: jnp.broadcast_to(self.f(t), *new_shape[:-2]))
```
gautierronan, line 430: Same comment as before here, and throughout the code later on.

dynamiqs/time_array.py
```diff
@@ -508,19 +524,37 @@ def shape(self) -> tuple[int, ...]:
...
+                lambda x: x.broadcast_to(*new_shape),
+                self.timearrays,
+                is_leaf=lambda x: isinstance(x, TimeArray),
+            )
+        )
+
     def conj(self) -> TimeArray:
         return SummedTimeArray([tarray.conj() for tarray in self.timearrays])
 
+    def in_axes(self) -> PWCTimeArray:
```
gautierronan, line 544: As already discussed privately, this is bugged for common batching over several timearrays summed together.

gautierronan (changes requested):
This is honestly really nice work. Great job, I think we're almost ready to merge. Apart from the small bug I raised and the small comments below, I think it's all good for me! 🚀

dynamiqs/time_array.py
abocquet: I don't think it'll be a problem because everything is a pytree here, so the vmapping will propagate. But it's worth checking

gautierronan: There's a lambda in the middle so I'm not 100% confident, but maybe!

docs/tutorials/time-dependent-operators.md
```diff
@@ -135,6 +135,18 @@ Array([[ 0.+0.j,  0.+0.j],
        [ 0.+0.j, -0.+0.j]], dtype=complex64)
 ```
 
+??? Notes "Batching"
+The batching of the returned time-array is specified by `values`. For example, to define a PWC operator batched over a parameter $\theta$:
```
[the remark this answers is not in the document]
pierreguilmin: I'll fix it.

pierreguilmin:
I'm reverting d7fb54d7cd29c74a5f92a98ae8c68e39cb6352fe for now (so the CI passes), I'll fix this and a few other things in a PR on top, that we can merge in this one after.

dynamiqs/time_array.py
abocquet: To be clear, the BatchedCallable is still here, only wrapped in a Partial

pierreguilmin:
The logic of `_flat_vectorize` is incorrect. Example bug:
```python
import dynamiqs as dq
import jax.numpy as jnp

n = 8
H = jnp.ones((3, 2, n, n))
rho0 = jnp.ones((2, n, n))
result = dq.mesolve(H, [], rho0, [0.0], options=dq.Options(cartesian_batching=False))
```
```
ValueError: vmap got inconsistent sizes for array axes to be mapped:
  * one axis had size 3: axis 0 of argument H.array of type complex64[3,2,8,8];
  * one axis had size 2: axis 0 of argument rho0 of type complex64[2,8,8]
```

Moreover, I think `_flat_vectorize` is actually not easy to implement, because broadcasting is not supported natively. Consider for example this MWE:
```python
import jax
import jax.numpy as jnp

def func(x, y):
    return x.T @ y.T

n = 2
x = jnp.ones((4, n, n))
y = jnp.ones((1, n, n))

jax.vmap(func, in_axes=(0, 0))(x, y).shape
```
```
ValueError: vmap got inconsistent sizes for array axes to be mapped:
  * one axis had size 4: axis 0 of argument x of type float32[4,2,2];
  * one axis had size 1: axis 0 of argument y of type float32[1,2,2]
```
The first option would be to broadcast all arrays before vectorizing, but this is annoying performance wise. There must be a solution, as `jnp.vectorize` does exactly what we want.

pierreguilmin:
I really think we should think of the current code in two blocks:
- first design a neat and working API for `_flat_vectorize` and `_cartesian_vectorize` (i.e. that works for any PyTree),
- then have time-arrays giving the corresponding `in_axes`/`nbatch`/whatever is needed by the `_flat_vectorize` and `_cartesian_vectorize` APIs.

So `_flat_vectorize` and `_cartesian_vectorize` should not refer to anything from `time_array.py`, they are just functions working on any PyTrees.

pierreguilmin:
We could separate the tests for clarity, and here's a proposal for a more complete test suite for flat batching:

```python
@pytest.mark.parametrize('nH', [(), (3,), (3, 4)])
@pytest.mark.parametrize('npsi0', [(), (5,), (5, 6)])
def test_cartesian_batching(nH, npsi0):
    # ...

# with H: (3, 4, n, n), no need to parametrize IMO
@pytest.mark.parametrize('npsi0', [(), (1,), (1, 1), (4), (3, 1), (1, 4), (3, 4), (5, 1, 1), (5, 3, 4)])
def test_flat_batching(npsi0):
    # ...
```

abocquet:
> The logic of _flat_vectorize is incorrect. Example bug:

Thank you for raising this issue. As you might have seen this usecase was covered in `sesolve` but I forgot to port it to `mesolve`. Fix fixed this overlook. Your example also revealed a hole in the testbed. I hadn't tested the case were one term was had no batching dimension at all. This is also fixed now. 

---

I am afraid I cannot make your MWE work, even without batching on `main`:
```python
import dynamiqs as dq
import jax.numpy as jnp

n = 8
H = jnp.ones((n, n))
rho0 = jnp.ones((n, n))
result = dq.mesolve(H, [], rho0, [0.0], options=dq.Options(cartesian_batching=False))
```
---

> The first option would be to broadcast all arrays before vectorizing, but this is annoying performance wise. There must be a solution, as jnp.vectorize does exactly what we want.

Finally, performance is of utmost importance of course.  I ran two versions of the same `sesolve` code with a batched version and a control manually vmapped version. Both codes have simliar performances. This is explained by the fact that JAX makes clever memory use for broadcasting, and returns a view, not a new array. This greenlights the current approach.

image


```python
import dynamiqs as dq
import jax.numpy as jnp
import jax

n = 8
H = jnp.ones((n, n))

vec_times = []
batched_times = []

for b in range(12):
    rho0 = jnp.ones((2 ** b, 2 ** b, n, 1))
    times = jnp.linspace(0, 1, 10)
    %timeit dq.sesolve(H, rho0, [0.0], options=dq.Options(cartesian_batching=False))

    def vectorized_version(crho0):
        return dq.sesolve(H, crho0, [0.0], options=dq.Options(cartesian_batching=False))

    vectorized_version = jax.vmap(vectorized_version)
    vectorized_version = jax.vmap(vectorized_version)
    vectorized_version = jax.jit(vectorized_version)

    res = %timeit -o vectorized_version(rho0).states.block_until_ready()
    vec_times.append(res.average)

    def batched_version(crho0):
        return dq.sesolve(H, rho0, [0.0], options=dq.Options(cartesian_batching=False))

    batched_version = jax.jit(batched_version)
    res = %timeit -o batched_version(rho0).states.block_until_ready()
    batched_times.append(res.average)
    
vec_times = jnp.array(vec_times) * 1e3
batched_times = jnp.array(batched_times) * 1e3

batching_size = 4 ** jnp.arange(12) / 1e6 
plt.plot(batching_size, vec_times, label="Manually vmapped version")
plt.plot(batching_size, batched_times, label="Batched version")
plt.grid()
plt.yscale("log")
plt.xlabel("Batching number (x10^6)")
plt.ylabel("Run time [ms]")
```

A quick look at the vectorize code shows they do something a bit more clever that broadcasting everything. But this is an optimization that we can keep for a next PR.

pierreguilmin:
Thanks for the detailed answer!

I trust vmap vs batching to be efficient (what's tested by your benchmark), what I'm questionning is vectorizing vs broadcasting + vmap:
```python
import jax
import jax.numpy as jnp

def f(A, B):
    return A.T @ B.T

A = jnp.ones((7, 8, 9, 64, 64))
B = jnp.ones((64, 64))

# === JAX vectorize
f1 = jax.jit(jnp.vectorize(f, signature='(n,n),(n,n)->(n,n)'))
f1(A, B)
%timeit f1(A, B).block_until_ready()

# === homemade vectorize
@jax.jit
def f2(A, B):
    B = jnp.broadcast_to(B, A.shape)
    return jax.vmap(jax.vmap(jax.vmap(f)))(A, B)

f2(A, B)
%timeit f2(A, B).block_until_ready()
```
```
3.28 ms ± 847 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
7.31 ms ± 1.15 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I'm fine with that we work on this in another PR, but I'm not sure there is an efficient way to adapt `jnp.vectorize` to our use case actually 😕 I'd like to make sure we know how to fix this performance issue later.

pierreguilmin (commented):
I've reviewed almost everything, very clean construction for the time arrays! 👏 I propose the gazillion nits in a PR on top, nothing structural. The only remaining concern is about the way we perform flat vectorization, as explained in the previous comment.

dynamiqs/time_array.py
pierreguilmin: Fixed in #565.

pierreguilmin:
(we can merge both my PRs later, as you prefer)

abocquet:
>I'm fine with that we work on this in another PR, but I'm not sure there is an efficient way to adapt jnp.vectorize to our use case actually 😕 I'd like to make sure we know how to fix this performance issue later.

I think it won't be that hard. I will make a follow up PR.

abocquet:
I merged everything, can we merge this one?

pierreguilmin (approved):
Yep, congratulations on all the work! Even though I was grumpy about the big PR, it's really amazing to have vectorized solvers, and there are some really non-trivial cool tricks in here. Curious to see the solution for efficient fast vectorization.

PR #568, opened 2024-05-07
Remove broadcasting before ND-batching

abocquet:
Who needs broadcasting anyway ? Remove broadcasting before ND-batching.
Also removes `expand_as_broadcastable`, that isn't used anymore.

Follows https://github.com/dynamiqs/dynamiqs/pull/562

Related to DYN-197.

### DIFF
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -16,6 +16,7 @@ from ..core._utils import (
     _flat_vectorize,
     catch_xla_runtime_error,
     get_solver_class,
+    squeeze_ones,
 )
 from ..gradient import Gradient
 from ..options import Options
@@ -133,23 +134,17 @@ def _vectorized_mesolve(
 ) -> MEResult:
     # === vectorize function
     # we vectorize over H, jump_ops and rho0, all other arguments are not vectorized
+    if not options.cartesian_batching:
+        broadcast_shape = jnp.broadcast_shapes(
+            H.shape[:-2], rho0.shape[:-2], *[jump_op.shape[:-2] for jump_op in jump_ops]
+        )
+        rho0 = jnp.broadcast_to(rho0, broadcast_shape + rho0.shape[-2:])
+
     # `n_batch` is a pytree. Each leaf of this pytree gives the number of times
     # this leaf should be vmapped on.
 
     # the result is vectorized over `_saved` and `infos`
     out_axes = MEResult(None, None, None, None, 0, 0)
-
-    if not options.cartesian_batching:
-        broadcast_shape = jnp.broadcast_shapes(
-            H.shape[:-2], rho0.shape[:-2], *[jump_op.shape[:-2] for jump_op in jump_ops]
-        )
-
-        def broadcast(x: TimeArray) -> TimeArray:
-            return x.broadcast_to(*(broadcast_shape + x.shape[-2:]))
-
-        H = broadcast(H)
-        jump_ops = list(map(broadcast, jump_ops))
-        rho0 = jnp.broadcast_to(rho0, broadcast_shape + rho0.shape[-2:])
 
     n_batch = (
         H.in_axes,
@@ -167,6 +162,8 @@ def _vectorized_mesolve(
         f = _cartesian_vectorize(_mesolve, n_batch, out_axes)
     else:
         f = _flat_vectorize(_mesolve, n_batch, out_axes)
+        H = squeeze_ones(H)
+        jump_ops = list(map(squeeze_ones, jump_ops))
 
     # === apply vectorized function
     return f(H, jump_ops, rho0, tsave, exp_ops, solver, gradient, options)
--- a/dynamiqs/mesolve/mesolve.py
+++ b/dynamiqs/mesolve/mesolve.py
@@ -16,7 +16,6 @@ from ..core._utils import (
     _flat_vectorize,
     catch_xla_runtime_error,
     get_solver_class,
-    squeeze_ones,
 )
 from ..gradient import Gradient
 from ..options import Options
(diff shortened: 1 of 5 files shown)


dynamiqs/mesolve/mesolve.py
```diff
@@ -133,23 +134,17 @@ def _vectorized_mesolve(
...
-    # this leaf should be vmapped on.
-
-    # the result is vectorized over `_saved` and `infos`
-    out_axes = MEResult(None, None, None, None, 0, 0)
-
     if not options.cartesian_batching:
         broadcast_shape = jnp.broadcast_shapes(
             H.shape[:-2], rho0.shape[:-2], *[jump_op.shape[:-2] for jump_op in jump_ops]
         )
+        rho0 = jnp.broadcast_to(rho0, broadcast_shape + rho0.shape[-2:])
```
pierreguilmin, line 141:
```suggestion
        broadcast_shape = jnp.broadcast_shapes(H.shape, rho0.shape, *[L.shape for L in jump_ops])
        rho0 = jnp.broadcast_to(rho0, broadcast_shape)
```
Same for sesolve.

pierreguilmin: Why do we need to broadcast the initial state?

dynamiqs/mesolve/mesolve.py
```diff
@@ -167,6 +162,8 @@ def broadcast(x: TimeArray) -> TimeArray:
         f = _cartesian_vectorize(_mesolve, n_batch, out_axes)
     else:
         f = _flat_vectorize(_mesolve, n_batch, out_axes)
+        H = squeeze_ones(H)
+        jump_ops = list(map(squeeze_ones, jump_ops))
```
pierreguilmin, line 166: If the passed Hamiltonian is of shape `(1, n, n)` or `(2, 1, n, n)` we don't want to squeeze the one. IMO `flat_vectorize` the magic you propose should be done by `_flat_vectorize`, it has enough information with the shape of the objects.

dynamiqs/time_array.py
```diff
@@ -222,6 +222,21 @@ def reshape(self, *new_shape: int) -> TimeArray:
...
         """
 
+    def squeeze(self, axis: int) -> TimeArray:
+        """Squeeze a time-array.
+
+        Args:
+            axis: Axis to squeeze.
+
+        Returns:
+        New time-array object with squeezed_shape
```
pierreguilmin, line 232:
```suggestion
        Returns:
            New time-array object with squeezed shape.
```
If possible it'd be great to render the doc when you modify it, to check that it looks good 😊

pierreguilmin (changes requested):
Nice!

dynamiqs/mesolve/mesolve.py
abocquet: If your Hamiltonian is (k,n,n) for instance, your state must be (k,n,n) too. Otherwise you cannot store the information of all situations. You can extend this reasoning for additional dimensions.

dynamiqs/time_array.py
abocquet: Right, sorry I will

dynamiqs/mesolve/mesolve.py
abocquet: You have to squeeze for emulating the broadcasting. I think it works as is but I'll add a test for your edge case.

dynamiqs/mesolve/mesolve.py
pierreguilmin: But it's automatic with broadcasting, i.e. the first time you do H @ rho you'll get a result with shape (k, n, n).

pierreguilmin: Ha but then maybe JAX/diffrax is happy that rho keeps the same shape all along?

dynamiqs/mesolve/mesolve.py
pierreguilmin: But you're also going to squeeze the original dimensions equal to one no?

abocquet: Rho is not squeezed so shapes may be preserved

dynamiqs/mesolve/mesolve.py
abocquet: After second though your comment won't hold for sesolve because it would broadcast a bra (n,1) into a dm (n,n)

pierreguilmin: Because broadcsasting is from the left, so in the case of sesolve it would be (k, n, 1).

pierreguilmin: No, it would be `(k, n, 1)`:
```python
H = jnp.ones((5, 2, 2))
rho = jnp.ones((2, 2))
print((H @ rho).shape)  # (5, 2, 2)
psi = jnp.ones((2, 1))
print((H @ psi).shape)  # (5, 2, 1)
```

gautierronan (changes requested):

dynamiqs/mesolve/mesolve.py
abocquet: While I do agree in theory, in practise tests break when I remove these lines.

gautierronan: I don't have time to investigate, and it's pretty minor, so I'd be tempted to keep it like that and to simply track it in the Linear. Unless @pierreguilmin you have an idea?

dynamiqs/mesolve/mesolve.py
```diff
@@ -167,6 +161,8 @@ def broadcast(x: TimeArray) -> TimeArray:
         f = _cartesian_vectorize(_mesolve, n_batch, out_axes)
     else:
         f = _flat_vectorize(_mesolve, n_batch, out_axes)
+        H = H.squeeze()
+        jump_ops = [x.squeeze() for x in jump_ops]
```
pierreguilmin, line 165: I don't see how this could work. The output _shape_ will be correct, but the computation will be the same for H `(2, 1, n, n)` and `(1, 2, n, n)` (it will always be squeeze to `(2, n, n)`.

See previous comment
> If the passed Hamiltonian is of shape (1, n, n) or (2, 1, n, n) we don't want to squeeze the one. IMO the magic you propose should be done by _flat_vectorize, it has enough information with the shape of the objects.

pierreguilmin (changes requested):
I think squeezing is not the correct approach, see comment.

PR #572, opened 2024-05-14
Improve documentation styling

gautierronan:
This PR introduces several styling modifications to the documentation. The overall goal is to make it look more clean.

- Smaller fonts overall
- Removes orange left border on code snippets (but they stay on function signatures in the API)
- Increase contrast in dark mode
- Improve page spacing
- Change admonition styling (only left border now)
- Remove "quotes" admonition altogether and associated CSS
- Add titles to all notes
- Fold all notes by default

For the latter, the reasoning is that if it's a note, it's not that important and does not need to appear right away. Otherwise it should be probably be in the main text or inside a warning.

To review, I advise moving around the documentation tutorials and API reference with a `task docserve`, on top of reading the code diff.

### DIFF
--- a/docs/tutorials/time-dependent-operators.md
+++ b/docs/tutorials/time-dependent-operators.md
@@ -57,7 +57,7 @@ for any time $t$, where $O_0$ is a constant operator.
 
 In dynamiqs, constant operators can either be defined with array-like objects or as [`TimeArray`][dynamiqs.TimeArray] objects (using the [`dq.constant()`][dynamiqs.constant] function).
 
-!!! Notes
+??? Notes "Common operators"
     Common operators are available as utility functions, see the list of available operators in the [Python API](../python_api/index.md#operators).
 
 For instance, to define the Pauli $Z$ operator $H = \sigma_z$, you can use any of the following syntaxes:
@@ -102,7 +102,7 @@ In dynamiqs, PWC operators are defined by three array-like objects:
 
 To construct a PWC operator, pass these three arguments to the [`dq.pwc()`][dynamiqs.pwc] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. This object then returns an array with shape _(..., n, n)_ when called at any time $t$.
 
-!!! Notes
+??? Notes "Sorting of `times`"
     The argument `times` must be sorted in ascending order, but does not need to be evenly spaced. When calling the resulting time-array object at time $t$, the returned array is the operator $c_k\ O_0$ corresponding to the interval $[t_k, t_{k+1}[$ in which the time $t$ falls. If $t$ does not belong to any time intervals, the returned array is null.
 
 Let's define a PWC operator:
@@ -137,7 +137,7 @@ Array([[ 0.+0.j,  0.+0.j],
        [ 0.+0.j, -0.+0.j]], dtype=complex64)
 ```
 
-??? Notes "Batching"
+??? Notes "Batching PWC operators"
     The batching of the returned time-array is specified by `values`. For example, to define a PWC operator batched over a parameter $\theta$:
     ```pycon
     >>> thetas = jnp.linspace(0, 1.0, 11)  # (11,)
@@ -182,7 +182,7 @@ Array([[0.+0.j, 1.+0.j],
        [1.+0.j, 0.+0.j]], dtype=complex64)
 ```
 
-??? Notes "Batching"
+??? Notes "Batching modulated operators"
     The batching of the returned time-array is specified by the array returned by `f`. For example, to define a modulated Hamiltonian $H(t)=\cos(\omega t)\sigma_x$ batched over the parameter $\omega$:
     ```pycon
     >>> omegas = jnp.linspace(0.0, 1.0, 11)  # (11,)
@@ -238,7 +238,7 @@ Array([[1., 0.],
 !!! Warning "The function `f` must return a JAX array (not an array-like object!)"
     An error is raised if the function `f` does not return a JAX array. This error concerns any other array-like objects. This is enforced to avoid costly conversions at every time step of the numerical integration.
 
-??? Notes "Batching"
+??? Notes "Batching callable time-arrays"
     The batching of the returned time-array is specified by the array returned by `f`. For example, to define an arbitrary time-dependent operator batched over a parameter $\theta$:
     ```pycon
     >>> thetas = jnp.linspace(0, 1.0, 11)  # (11,)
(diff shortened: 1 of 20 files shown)


abocquet (approved):

docs/tutorials/time-dependent-operators.md
```diff
@@ -238,7 +238,7 @@ Array([[1., 0.],
 !!! Warning "The function `f` must return a JAX array (not an array-like object!)"
     An error is raised if the function `f` does not return a JAX array. This error concerns any other array-like objects. This is enforced to avoid costly conversions at every time step of the numerical integration.
 
-??? Notes "Batching"
+??? Notes "Batching callable time-arrays"
```
pierreguilmin, line 241:
```suggestion
??? Notes "Batching arbitrary time-dependent operators"
```

dynamiqs/mesolve/mesolve.py
```diff
@@ -56,15 +56,15 @@ def mesolve(
     where $H(t)$ is the system's Hamiltonian at time $t$ and $\{L_k(t)\}$ is a
     collection of jump operators at time $t$.
 
-    Quote: Time-dependent Hamiltonian or jump operators
+    Notes-: Defining a time-dependent Hamiltonian or jump operator
```
pierreguilmin, line 59:
```suggestion
    Notes-: Defining a time-dependent Hamiltonian or jump operators
```

dynamiqs/plots/plots_wigner.py
```diff
@@ -264,7 +264,7 @@ def plot_wigner_gif(
         This function creates files in the current working directory under
         `.tmp/dynamiqs`.
 
-    Note:
+    Notes-: Display in Jupyter
```
pierreguilmin, line 267:
```suggestion
    Notes-: Display in Jupyter notebooks
```

dynamiqs/result.py
```diff
@@ -158,7 +158,7 @@ class MEResult(Result):
         - If the option `cartesian_batching = True` (default value), the results leading
           dimensions are
           ```
-          ... = ...H, ...L0, ...L1, (...), ...rho0
+          ... = ...H, ...L0, ...L1, ...Ln, ...rho0
```
pierreguilmin, line 161: I would revert this, it's incorrect.

dynamiqs/time_array.py
```diff
@@ -46,24 +46,24 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
...
         times _(array_like of shape (N+1,))_: Time points $t_k$ defining the boundaries
             of the time intervals, where _N_ is the number of time intervals.
         values _(array_like of shape (..., N))_: Constant values $c_k$ for each time
             interval.
         array _(array_like of shape (n, n))_: Constant array $O_0$.
 
+    Notes-: `times` sorting
+        The argument `times` argument must be sorted in ascending order, but does not
+        need to be evenly spaced.
+
```
pierreguilmin, line 59: Nothing should be in-between Args and Returns. Typically we put all notes on top of args, and I think it's better not to mix the function description and the args/returns section.

dynamiqs/time_array.py
```diff
@@ -46,24 +46,24 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
...
-
     Args:
         times _(array_like of shape (N+1,))_: Time points $t_k$ defining the boundaries
             of the time intervals, where _N_ is the number of time intervals.
         values _(array_like of shape (..., N))_: Constant values $c_k$ for each time
             interval.
         array _(array_like of shape (n, n))_: Constant array $O_0$.
 
+    Notes-: `times` sorting
+        The argument `times` argument must be sorted in ascending order, but does not
```
pierreguilmin, line 57:
```suggestion
        The argument `times` must be sorted in ascending order, but does not
```

pierreguilmin (changes requested):
Very nice!

> Smaller fonts overall

In think we should revert this, I already found the font to be borderline small, now it's getting hard to read. Maybe I just need glasses 👓

> Add titles to all notes
> Fold all notes by default
> For the latter, the reasoning is that if it's a note, it's not that important and does not need to appear right away. Otherwise it should be probably be in the main text or inside a warning.

It's a bit annoying to click and open all notes one by one. To me notes != "not that important", it's just a way to add colours to the doc, and separate bits of informations, unrelated to the importance or not of the information. For example in https://www.dynamiqs.org/tutorials/open-systems.html examples have a clear separated styling, and https://www.dynamiqs.org/python_api/utils/utils/lindbladian.html the note is just an extra information that pops out as such. I think folding notes actually _forces_ the curious user to open them (to know what's inside), while he can quickly skim through otherwise. It's annoying to have to click to reveal one line of text. IMO we should only fold very long or very specific notes.

Moreover, when the note is not folded the title is often redundant with the content, and I like keeping "Note" then, it unifies a bit all these blue things throughout the documentation. So I think most of the newly added titles are just redundant informations.

dynamiqs/result.py
pierreguilmin: (`n` is typically used for the Hilbert space dimension, and this does not clearly explain that there are arbitrarily many jump operators)

gautierronan: Ok, I didn't really like the previous notation, but I agree this one is not good. I'm reverting for lack of something better.

dynamiqs/mesolve/mesolve.py
pierreguilmin: Right sorry

dynamiqs/plots/utils.py
```diff
@@ -90,7 +90,7 @@ def gridplot(
     Warning:
         Documentation redaction in progress.
 
-    Notes:
+    Notes-: Shortcut to Matplotlib `plt.subplots()`
```
pierreguilmin, line 93: I would expand this.

pierreguilmin (changes requested):
Beautiful thanks a lot for all the fixes! Minor neat. Also should we stick with `Note` or `Notes`? We still use both.

gautierronan:
> Also should we stick with Note or Notes? We still use both.

So not sure on this one. I prefer "Note" to "Notes" on the documentation when there's no title (it's similar to "Warning") because there's usually a single comment. However, in docstrings, other sections are usually plural ("Args", "Returns", etc.). 

I can either unify everything to be "Note" instead of "Notes", or write "Notes: Note" in docstrings, which sounds kinda dumb haha.

pierreguilmin:
> > Also should we stick with Note or Notes? We still use both.
> 
> So not sure on this one. I prefer "Note" to "Notes" on the documentation when there's no title (it's similar to "Warning") because there's usually a single comment. However, in docstrings, other sections are usually plural ("Args", "Returns", etc.).
> 
> I can either unify everything to be "Note" instead of "Notes", or write "Notes: Note" in docstrings, which sounds kinda dumb haha.

Let's put "Note" everywhere!

pierreguilmin (approved):
Beautiful 🎨✨

PR #567, opened 2024-05-05
Introduce QArrays

abocquet:
Creates the basics for the `QArray` class. Is is not related to anything in the main codebase yet. The PR has a big volume but is 50% docs. If it is too annoying I can cut.
This PR only introduces shortcuts to existing dynamiqs helper functions and also the '&' operator for convience. 

For next PRs:
- [ ]  compatibilty with `TimeArray`
- [ ] compatibility in se/mesolve
- [ ] make functions (`destroy`, ...) return `QArray`s

Relates to https://linear.app/dynamiqs/issue/DYN-210/qarray-object

PR #577, opened 2024-05-16
Base structure for SparseQArray class

bsaoptima:
Adds the base structure for the `SparseQArray` class with simple arithmetic methods: `_add__`, `__sub__` and `__mul__`.

In following PRs:

- [x] https://github.com/dynamiqs/dynamiqs/pull/581
- [x] https://github.com/dynamiqs/dynamiqs/pull/583
- [x] https://github.com/dynamiqs/dynamiqs/pull/585
- [x] https://github.com/dynamiqs/dynamiqs/pull/590
- [x] https://github.com/dynamiqs/dynamiqs/pull/591

Related to DYN-48


abocquet:
Good job 👍 maybe we can wait for https://github.com/dynamiqs/dynamiqs/pull/567 to be merged before

dynamiqs/sparse.py
```diff
@@ -0,0 +1,124 @@
...
+import jax.numpy as jnp
+from jaxtyping import Array, ArrayLike, ScalarLike
+
+from .qarray import QArray
+
+
+class SparseQArray(QArray):
+    diags: Array
+    offsets: tuple[int, ...] = eqx.field(static=True)
+    dims: tuple[int, ...] = eqx.field(static=True)
```
abocquet, line 13: You probably don't want these fields to be static, it'll force jax to recompile for each array you pass

dynamiqs/sparse.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __add__(
+        self, other: ScalarLike | ArrayLike | SparseQArray
+    ) -> Array | SparseQArray:
+        if isinstance(other, ScalarLike):
+            if other == 0:
+                return self
+            return self.to_dense() + other
+        elif isinstance(other, ArrayLike):
```
abocquet, line 22: It feels like you could just do `diags + other` here

dynamiqs/sparse.py
```diff
@@ -0,0 +1,124 @@
...
+    def __radd__(self, other: Array) -> Array:
+        return self + other
+
+    def __sub__(
+        self, other: ScalarLike | ArrayLike | SparseQArray
+    ) -> Array | SparseQArray:
+        if isinstance(other, ScalarLike):
+            if other == 0:
+                return self
+            return self.to_dense() - other
```
abocquet, line 52: Same here

dynamiqs/sparse.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __rmul__(self, other: ArrayLike) -> Array:
+        return self * other
+
+
+def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
+    if dims1 != dims2:
+        raise ValueError(
+            f'QArrays have incompatible dimensions. Got {dims1} and {dims2}.'
+        )
```
abocquet, line 124:
```suggestion
def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
    if dims1 != dims2:
        raise ValueError(
            f'QArrays have incompatible dimensions. Expected `(..., n, k)` and `(..., k, m)`. Got {dims1} and {dims2}.'
        )
```

abocquet (changes requested):

dynamiqs/sparse.py
gautierronan: These are not `dims` of a matrix product, but the `dims` for the underlying tensor product. So they should be equal, and not `(..., n, k)` and `(..., k, m)`.

But we can add "Expected `dims1 == dims2`, but got {dims1} and {dims2}."

dynamiqs/sparse.py
gautierronan: This is not the same operation. You would be adding a constant value only to the non-zero diagonals, which is not what you expect from the standard `__add__` on an array. Calling `x+2` adds a `2` to all matrix elements, so we return a dense matrix of a sparse one.

dynamiqs/sparse.py
```diff
@@ -73,28 +69,29 @@ def _sub_dia(self, other: SparseQArray) -> SparseQArray:
...
+    def __rsub__(self, other: Array) -> Array:
+        return -self + other
 
     def __mul__(self, other: Array | SparseQArray) -> Array | SparseQArray:
         if isinstance(other, ScalarLike):
             if other == 0:
-                empty_diags = jnp.empty(0, self.shape[-1])
-                return SparseQArray(empty_diags, ())
-            return SparseQArray(other * self.diags, self.offsets)
+                diags = jnp.empty(0, self.shape[-1])
```
bsaoptima, line 80: @gautierronan we do not have a `shape` property in this PR, I think `self.shape[-1]` should instead be `self.dims[0]`.

gautierronan: Yes, I guess this will be fixed once we have implemented the `.shape` method. Let's just wait for this I think.

dynamiqs/sparse.py
abocquet: Oh right I see. Then we want to set a warning because this does 2 things we probably don't want to
- convert a sparse a dense matrix
- add value to all elements (A + x) instead of (A + x. 1)

dynamiqs/sparse.py
abocquet: Yes let's talk about it

pierreguilmin (changes requested):
Nice work! First batch of comments.

dynamiqs/sparse.py
gautierronan: Probably we'll start by keeping `offsets` static, and think about making them dynamic later on. It's not so clear yet whether this is doable with `jax.dynamic_slice` or not (because the function still requires a static size as input).

dynamiqs/sparse.py
```diff
@@ -0,0 +1,121 @@
...
+import equinox as eqx
+import jax.numpy as jnp
+from jaxtyping import Array, ArrayLike, Scalar, ScalarLike
+
+from .qarray import QArray
+
+
+class SparseQArray(QArray):
+    diags: Array
+    offsets: tuple[int, ...] = eqx.field(static=True)
```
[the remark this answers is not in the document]
bsaoptima: It would be better, but since some methods currently require offsets to be static and Python sets are not hashable objects, I will have to keep it as a tuple for the time being.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,121 @@
...
+
+import equinox as eqx
+import jax.numpy as jnp
+from jaxtyping import Array, ArrayLike, Scalar, ScalarLike
+
+from .qarray import QArray
+
+
+class SparseQArray(QArray):
+    diags: Array
```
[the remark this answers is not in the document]
bsaoptima: I will have a go at it, I don't think it would be that harmful, we already iterate through the diagonals and slice them so your suggestion makes sense.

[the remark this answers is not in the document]
pierreguilmin: Then we could also store `diags` and `offsets` as a dict `{offset: diag}`.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,124 @@
...
+            else:
+                out_offsets_diags[other_offset] = -other_diag
+
+        out_offsets = tuple(sorted(out_offsets_diags.keys()))
+        out_diags = jnp.array([out_offsets_diags[offset] for offset in out_offsets])
+
+        return SparseQArray(out_diags, out_offsets, self.dims)
+
+    def __rsub__(self, other: Array) -> Array:
+        return -self + other
```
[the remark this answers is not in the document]
pierreguilmin: You need to implement `__neg__`.

dynamiqs/sparse.py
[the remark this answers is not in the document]
bsaoptima: I agree for the comment and yes the offsets need to be static as there is conditional logic based on their value involved. I am trying to remove this condition and if successful in a timely manner, we can do another PR.

dynamiqs/sparse.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __mul__(self, other: Array | SparseQArray) -> Array | SparseQArray:
+        if isinstance(other, (complex, Scalar)):
+            diags, offsets = other * self.diags, self.offsets
+            return SparseQArray(diags, offsets, self.dims)
+        elif isinstance(other, Array):
+            return self._mul_dense(other)
+        elif isinstance(other, SparseQArray):
+            _check_compatible_dims(self.dims, other.dims)
+            return self._mul_sparse(other)
```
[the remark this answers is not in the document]
bsaoptima: Not sure what you mean by adding a warning here?

dynamiqs/sparse.py
```diff
@@ -0,0 +1,121 @@
...
+            else:
+                out_offsets_diags[other_offset] = other_diag
+
+        out_offsets = tuple(sorted(out_offsets_diags.keys()))
+        out_diags = jnp.stack([out_offsets_diags[offset] for offset in out_offsets])
+
+        return SparseQArray(out_diags, out_offsets, self.dims)
+
+    def __radd__(self, other: Array) -> Array:
+        return self + other
```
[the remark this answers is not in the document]
bsaoptima: I propose we keep it for the time being as some tests and benchmarks rely on the existence of this dunder. When the base class is done we can remove it.

dynamiqs/sparse.py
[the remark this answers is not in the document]
bsaoptima: I agree in principle, once we merge all PRs, I will make the change, so I don't miss anything.

pierreguilmin (changes requested):
I think we have to wait for https://github.com/dynamiqs/dynamiqs/pull/567 to be finished before moving further with this PR.

> > What about the less explicit namings v for diags and k for offsets? It's much easier to read the numerical operations after IMO, bc way less verbose.

> I agree in principle, once we merge all PRs, I will make the change, so I don't miss anything.

I'd prefer doing it from the start, it will make the reviews easier. Did you ask @gautierronan if he agrees with this specific change? We can adjust the other PRs one by one after as I review them, no need to propagate the change now.

pierreguilmin:
So should we make `diags` a `list[Array]` instead of `Array`? If not we should add a comment along the line of
```
diagonals are padded with 0 and gathered in a singl array
```

pierreguilmin:
Actually I think we should maybe use a `frozendict` to have static dictionnary {k: v} with JAX arrays values. And if not for some reason a `frozenset` for keys.

dynamiqs/sparse_qarray.py
```diff
@@ -0,0 +1,158 @@
...
+
+        out_offsets = tuple(sorted(out_offsets_diags.keys()))
+        out_diags = jnp.array([out_offsets_diags[offset] for offset in out_offsets])
+
+        return SparseQArray(out_diags, out_offsets, self.dims)
+
+    def __rsub__(self, other: Array) -> Array:
+        return -self + other
+
+    def __mul__(self, other: Array | SparseQArray) -> Array | SparseQArray:
```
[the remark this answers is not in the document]
gautierronan: This is not a matmul implementation, it's a dot product between two matrices. Why would you want to remove this operation from the API?

PR #578, opened 2024-05-16
Update the batching tutorial with the new API

gautierronan:
CI is not passing for two reasons:
 - I am using the `dq.coherent(N, alphas)` API which is yet to be implemented (will do asap),
 - I am using a `%timeit` which doctest doesn't like: any idea how to fix this?

Otherwise, comments are welcome on the tutorial.

Closes DYN-206.


pierreguilmin:
Beautiful start @gautierronan, thanks a lot. Here is a first iteration on top, with some reformulations / additions and trying to use tabs where relevant. Tell me if you like it, and I'm happy to discuss/revert what you don't like.

Note that:
- I tried removing the use of `jnp.array` and `jnp.stack`, they are not needed.
- I removed the plural for batched arguments, I think it's better practice to use singular everywhere.
- For now I removed the `jnp.vectorize` example. We need to put it back of course, but the proposed example was not relevant, because the function was *already* working for batched/unbatched arguments without having to use `jnp.vectorize`.

gautierronan:
New iteration of the tutorial. I mainly tried to reduce the number of admonitions in the second section, and made several rephrasings all around. I also removed the empty `jnp.vectorize` admonition, now tracked in https://linear.app/dynamiqs/issue/DYN-215/add-jnpvectorize-example-in-batching-simulations-tutorial.

I'm still not 100% satisfied with this tutorial, but I think it's good enough to merge.

pierreguilmin:
Great 🤩

pierreguilmin (approved):

docs/documentation/basics/batching-simulations.md
```diff
@@ -1,86 +1,255 @@
...
+    !!! Example "Example: Cartesian batching with `dq.mesolve`"
+        - `H` has shape _(2, 3, n, n)_,
+        - `jump_ops = [L0, L1]` has shape _[(4, 5, n, n), (6, n, n)]_,
+        - `rho0` has shape _(7, n, n)_,
+
+        then `result.states` has shape _(2, 3, 4, 5, 6, 7, ntsave, n, n)_.
+
+### Flat batching
+
+The simulation runs for each set of Hamiltonians, jump operators and initial states using broadcasting. This mode can be activated by setting `cartesian_batching=False` in [`dq.Options`][dynamiqs.Options]. In particular for [`dq.mesolve()`][dynamiqs.mesolve], each jump operator can be batched independently from the others.
```
pierreguilmin, line 80: The "In particular ..." is not specific to flat batching, we should move it elsewhere @gautierronan. I merge bc I don't want to block longer, this is already so much better than the previous version.

PR #581, opened 2024-05-17
Add `to_dense` and `to_sparse` conversion functions for SparseQArray

bsaoptima:
Adds two utility functions for the SparseQArray class: `to_dense` and `to_sparse`.

- `to_dense` is both a public function and class method that allows users to turn a SparseQArray object in a classical dense format.
- `to_sparse` is a public function that turns a DenseQArray object in a SparseQArray object. In a typical user workflow, they would first leverage this function to use the SparseQArray format.

Related to DYN-48


dynamiqs/sparse.py
```diff
@@ -122,3 +127,68 @@ def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
...
+    presents sparse features in the SparseDIA custom format.
+
+    Args:
+        other: NxN matrix to turn from dense to SparseDIA format.
+
+    Returns:
+        SparseDIA object which has 2 main attributes:
+            object.diags: Array where each row is a diagonal.
+            object.offsets: tuple of integers that represents the
+                            respective offsets of the diagonals
```
gautierronan, line 163: No need to specify the attributes here. Eventually we may redirect towards the documentation that describes a `SparseQArray`.

dynamiqs/sparse.py
```diff
@@ -122,3 +127,68 @@ def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
...
+    """
+    diagonals = []
+    offsets = []
+
+    N = other.shape[0]
+    offset_range = 2 * N - 1
+    offset_center = offset_range // 2
+
+    if not isinstance(other, Array):
+        other = jnp.asarray(other.array)
```
gautierronan, line 173:
```suggestion
    other = jnp.asarray(other.array)
```
(should be at the beginning of the function too).

dynamiqs/sparse.py
```diff
@@ -122,3 +127,68 @@ def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
...
+            object.diags: Array where each row is a diagonal.
+            object.offsets: tuple of integers that represents the
+                            respective offsets of the diagonals
+    """
+    diagonals = []
+    offsets = []
+
+    N = other.shape[0]
+    offset_range = 2 * N - 1
+    offset_center = offset_range // 2
```
gautierronan, line 170: Isn't this always `N` or `N-1` ?

dynamiqs/sparse.py
```diff
@@ -122,3 +127,68 @@ def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
...
+            padded_diagonal = jnp.pad(
+                diagonal, padding, mode='constant', constant_values=0
+            )
+            diagonals.append(padded_diagonal)
+            offsets.append(offset)
+
+    diagonals = jnp.array(diagonals)
+    offsets = tuple(offsets)
+
+    return SparseQArray(diagonals, offsets, (N, N))
```
gautierronan, line 194: `dims` is not necesseraly `(N, N)`. Here it should be inferred from the input `DenseQArray`, or given as an optional argument to the function if the user wants to specify it. Otherwise, we can indeed set it to `(N, N)`.

gautierronan (changes requested):
Great work! Here's a set of comments, mainly regarding the docstrings and function signatures. Otherwise it looks good.

pierreguilmin (approved):

PR #583, opened 2024-05-17
Add test functions for `__add__`, `__sub__`, `__mul__`

bsaoptima:
This PR is to add the test functions for the base methods we currently have implemented for `__add__`,  `__sub__` and `__mul__` in the SparseQArray class.

There was a small debate regarding if the base is `bagrici/utility-sparse` or `bagrici/base-sparse`, I think the former is the best choice since the tests depend on the `to_sparse` and `to_dense` utility functions. That way we can merge and then merge again on `bagrici/base-sparse`.


tests/test_sparse.py
```diff
@@ -0,0 +1,53 @@
...
+import jax.numpy as jnp
+import pytest
+
+import dynamiqs as dq
+
+
+class TestSparseDIA:
+    @pytest.fixture(autouse=True)
+    def _setup(self):
+        N = 4
```
gautierronan, line 12: I would be tempted to set `N` to slightly larger than `4` to make sure we are not lucky. Something like 6 or 10 should be plenty.

tests/test_sparse.py
```diff
@@ -0,0 +1,53 @@
...
+        out_dia_dia = (self.sparseA - self.sparseB).to_dense()
+        out_dia_dense = self.sparseA - self.matrixB
+        out_dense_dia = self.matrixA - self.sparseB
+        out_dense_dense = self.matrixA - self.matrixB
+
+        assert jnp.allclose(out_dense_dense, out_dia_dia, rtol=rtol, atol=atol)
+        assert jnp.allclose(out_dense_dense, out_dia_dense, rtol=rtol, atol=atol)
+        assert jnp.allclose(out_dense_dense, out_dense_dia, rtol=rtol, atol=atol)
+
+    def test_transform(self, rtol=1e-05, atol=1e-08):
```
gautierronan, line 38:
```suggestion
    def test_convert(self, rtol=1e-05, atol=1e-08):
```
(this test function should appear either after `test_mul` or before `test_add`)

gautierronan (changes requested):
A few small comments, mainly regarding the test matrix definitions. Otherwise it LGTM.

tests/test_sparse.py
```diff
@@ -9,11 +9,21 @@
...
     def _setup(self):
-        N = 4
-        self.matrixA = dq.destroy(N)
-        self.matrixB = dq.number(N)
-        self.sparseA = dq.to_sparse(self.matrixA)
-        self.sparseB = dq.to_sparse(self.matrixB)
+        N = 10
+        num_diags = 3
+        diags = jnp.arange(num_diags * N).reshape(num_diags, N)
+        offsets = tuple(range(-(num_diags // 2), (num_diags // 2) + 1))
```
gautierronan, line 15: As we discussed, I think there's a mismatch between `num_diags` and the size of `offsets`. Once this is fixed, it's good to go for me.

pierreguilmin (approved):

PR #585, opened 2024-05-17
Add basic properties & methods to `SparseQArray`

bsaoptima:
Adding the basic properties to `SparseQArray`:

- `shape`
- `ndim`
- `mT`


gautierronan:
From the RFC, I think this PR should add:
 - `dtype`
 - `shape`
 - `ndim`
 - `conj`
 - `dag`
 - `norm`
 - `unit`
 - `reshape`
 - `broadcast_to`
 - `ptrace` (this one is not easy, either we do it in a separate PR, or we convert to dense and call `dq.ptrace` in the appropriate way)

dynamiqs/sparse.py
```diff
@@ -1,17 +1,65 @@
...
     offsets: tuple[int, ...] = eqx.field(static=True)
     dims: tuple[int, ...] = eqx.field(static=True)
 
+    @property
+    def dtype(self) -> jnp.dtype:
+        return self.diags.dtype
+
+    @property
+    def shape(self) -> tuple[int, ...]:
+        return self.diags.shape
```
gautierronan, line 22: The shape of the matrix is not the shape of `self.diags`, it's (N, N).

dynamiqs/sparse.py
```diff
@@ -1,17 +1,65 @@
...
+
+    @property
+    def ndim(self) -> int:
+        return len(self.dims)
+
+    @property
+    @pack_dims
+    def mT(self) -> SparseQArray:
+        N = self.dims[0]
```
gautierronan, line 31: Should be `shape` here, not `dims`.

gautierronan (changes requested):
Two minor comments, otherwise it's good.

pierreguilmin (approved):

PR #586, opened 2024-05-18
Simplify operators documentation

pierreguilmin:


dynamiqs/utils/operators.py
```diff
@@ -34,24 +34,21 @@
 def eye(*dims: int) -> Array:
     r"""Returns the identity operator.
 
-    If only a single dimension is provided, `eye` returns the identity operator
-    of corresponding dimension. If instead multiples dimensions are provided, `eye`
-    returns the identity operator of the composite Hilbert space given by the product
-    of all dimensions.
```
gautierronan, line 40: I quite enjoyed these comments. This is kinda new behavior compared to QuTiP, so I wouldn't expect users to understand that just from the examples below. IMO we should keep it (same in other functions), unless you're really keen in removing it.

gautierronan (changes requested):
See comment below. I find the other changes are great though!

dynamiqs/utils/operators.py
pierreguilmin: Yes I agree, let me propose a mathematically precise formulation.

pierreguilmin:
I think there's no possible ambiguity with this explicit formulation. It also introduces a bit better the notations used in the examples after. Tell me what you think @gautierronan 😉 
Screenshot 2024-05-26 at 12 13 22

pierreguilmin:
I've slipped in a quick fix to use the same notations in `states.py`. 😇

gautierronan (approved):
Looks great!

PR #588, opened 2024-05-21
Fix progress bar

pierreguilmin:
Fixes on top of #428 and #587.

Left to do:
- [ ] Find a good progress bar style.
- [ ] Fix documentation output in places where we use `dq.sesolve/dq.mesolve` to include progress bar.
- [ ] Document progress bars class in public API.
- [ ] Choose either the term "progress bar" or "progress meter" for doc and API and stick to it.
- [ ] Add progress bar to non-diffrax solver > https://linear.app/dynamiqs/issue/DYN-212/add-progress-bar-to-non-diffrax-solver.

PR #589, opened 2024-05-21
New progress bar format

pierreguilmin:
On top of #588.

Proposed format:
```
|█████▉    |  59.1% ◆ total 00:02 ◆ remaining 00:01
```
```
|██████████| 100.0% ◆ total 00:04 ◆ remaining 00:00
```

This is a bit of a hot-fix, @abocquet can you make a PR on top so that `TqdmProgressMeter` has a `bar_format` attribute that can be set by the user at `__init__()`, and then dynamically creates the suited `_DiffraxTqdmProgressMeter` class upon call to `to_diffrax()`? (or just inherit directly from `dx.TqdmProgressMeter`) Example usage:
```python
custom_progress_bar = dq.progress_meter.TqdmProgressMeter(bar_format='{percentage:5.1f}% ({elapsed}/{remaining})')
dq.sesolve(..., options=dq.Options(progress_bar=custom_progress_bar))
```
Ideally `dx.TqdmProgressMeter` should be fully configurable (it should pass all tqdm related settings to the tqdm instance upon creation).

### DIFF
--- a/dynamiqs/progress_meter.py
+++ b/dynamiqs/progress_meter.py
@@ -32,7 +32,7 @@ class _DiffraxTqdmProgressMeter(dx.TqdmProgressMeter):
     @staticmethod
     def _init_bar() -> tqdm.tqdm:
         bar_format = (
-            '|{bar}| {percentage:4.1f}% ◆ total {elapsed} ◆ remaining {remaining}'
+            '|{bar}| {percentage:5.1f}% ◆ total {elapsed} ◆ remaining {remaining}'
         )
         return tqdm.tqdm(total=100, unit='%', bar_format=bar_format)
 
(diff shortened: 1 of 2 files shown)


gautierronan (approved):
I think the format is missing some indication of speed (useful for very short simulation to know how they scale). Also not a huge fan of the diamonds, but it's rather esthetic.

I can propose an alternative later if you want.

pierreguilmin:
I've added the two following tasks to https://linear.app/dynamiqs/issue/DYN-213/improve-progress-meter-support
- [ ] Find a progress bar format we're happy with (e.g. currently for fast solve the time is 00:00 😢).
- [ ] Support user-defined tqdm progress bar format.

PR #591, opened 2024-05-23
Integrating SparseQArray in the dynamiqs workflow.

bsaoptima:
In this PR, we are integrating the SparseQArray format to be compatible with `sesolve` & `mesolve` for all types of supported Hamiltonians. We will also implement the test functions to verify the accuracy of our results compared to the classical Dense format.


dynamiqs/time_array.py
```diff
@@ -30,6 +31,9 @@ def constant(array: ArrayLike) -> ConstantTimeArray:
         _(time-array object of shape (..., n, n) when called)_ Callable object
             returning $O_0$ for any time $t$.
     """
+    if isinstance(array, SparseQArray):
+        return ConstantTimeArray(array)
+
```
gautierronan, line 36: Here (and below), we should still keep the call to `check_shape` even if the input is a `SparseQArray`. So only the `jnp.asarray` should be discarded here.

gautierronan (changes requested):
That's great! Only a minor comment below, but otherwise it's good with me

PR #594, opened 2024-05-30
Small doc improvements (annotations, implicite time in ODEs/SDEs)

pierreguilmin:
A bunch of minor improvements to the doc.

Before | after
Screenshot 2024-05-31 at 00 09 05

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,6 +19,7 @@ theme:
     repo: fontawesome/brands/github
     admonition:
       example: material/human-male-board
+      annotation: material/arrow-right-circle
   logo: media/logo.png
   favicon: media/favicon.svg
   palette:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,7 +19,7 @@ theme:
     repo: fontawesome/brands/github
     admonition:
       example: material/human-male-board
-      annotation: material/arrow-right-circle
+    annotation: material/plus-circle
   logo: media/logo.png
   favicon: media/favicon.svg
   palette:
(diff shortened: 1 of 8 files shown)


mkdocs.yml
```diff
@@ -19,6 +19,7 @@ theme:
     repo: fontawesome/brands/github
     admonition:
       example: material/human-male-board
+      annotation: material/arrow-right-circle
```
gautierronan, line 22: This doesn't seem to work, right? It's not an arrow which is used but a plus.

gautierronan (dismissed):
I'm not really a fan of these annotations TBH. I tried to reduce their sizes in 0c9a7c35ebfb54d7587cef4b297c0af898f6f246 but I still don't find this 100% satisfactory.

I might be in favor of going back to footnotes, which makes it super clear that there's something additional to be read (here it's not super clear that the button is clickable).

I still approve the PR because it's not that important, and the other changes are good. Feel free to merge like that, or to revert back to footnotes if you agree with me.

mkdocs.yml
pierreguilmin: Right sorry, I just fixed it. See the doc here > https://squidfunk.github.io/mkdocs-material/reference/annotations/.

gautierronan (approved):

pierreguilmin:
> I might be in favor of going back to footnotes, which makes it super clear that there's something additional to be read (here it's not super clear that the button is clickable).

Footnotes have two issues:
- they don't work in the API
- they jump the page completely to the end, it's really annoying when reading

I would say it's quite natural to be "what's that" and click. I think the current is much too small though, see the mkdocs-material doc for example > https://squidfunk.github.io/mkdocs-material/reference/annotations/.

pierreguilmin:
I'll merge this and we can fix the style in another PR. Maybe we should use another icon? Like:
Screenshot 2024-06-06 at 22 02 13

PR #595, opened 2024-05-30
Add implicit solvers to dynamiqs

gautierronan:
Waiting on https://github.com/patrick-kidger/diffrax/issues/428 before marking the PR as ready.

Even on a very simple stiff problem (4-legged dissipative cat), the implicit solver is getting stuck. We may have to play with the root finding to improve this?
```python
import dynamiqs as dq
import jax.numpy as jnp

# units
kHz, MHz = 2 * jnp.pi * 1e-3, 2 * jnp.pi
ns, us = 1e-3, 1e0

# parameters
N = 16
alpha = jnp.sqrt(4.0)
k4 = 100.0 * kHz
T = 2 * us

# operators
a = dq.destroy(N)
i = dq.eye(N)

# Hamiltonian and jump operators
H = dq.zero(N)
jump_ops = [jnp.sqrt(k4) * (dq.powm(a, 4) - alpha**4 * i)]

# initial state
psi0 = dq.coherent(N, 0.0)

# time
tsave = jnp.linspace(0.0, T, 101)

# run simulation
solver = dq.solver.Kvaerno5(atol=1e-4, rtol=1e-4)
result = dq.mesolve(H, jump_ops, psi0, tsave, solver=solver)
print(result)
```

Closes DYN-159.

### DIFF
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -35,6 +35,8 @@ The **dynamiqs** Python API features two main types of functions: solvers of dif
         - Tsit5
         - Dopri5
         - Dopri8
+        - Kvaerno3
+        - Kvaerno5
         - Euler
         - Rouchon1
         - Rouchon2
--- a/dynamiqs/core/diffrax_solver.py
+++ b/dynamiqs/core/diffrax_solver.py
@@ -143,3 +143,11 @@ class Dopri8Solver(AdaptiveSolver):
 
 class Tsit5Solver(AdaptiveSolver):
     diffrax_solver = dx.Tsit5()
+
+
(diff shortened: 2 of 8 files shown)


gautierronan:
In the end, the solvers work as intended but my example was limited by single-precision. I added a warning in the solver page, and with https://github.com/dynamiqs/dynamiqs/pull/596, it should be enough.

I'm not adding tests for the implicit solvers since they're already tested by diffrax, but can do so if needed.

pierreguilmin (approved):
I have a few edits for the doc, I'll make a PR on top 😉 Thanks a lot for the investigation, that's a great addition!!

pierreguilmin (approved):

PR #596, opened 2024-05-30
First draft for single-precision in sharp bits

gautierronan:
A quick first draft. I'm not sure whether there are other things to mention in the bullet point list? Also, we can complete this as we go.

Putting this first in the sharp bits because I think it's quite a big one. But we can also move it after the operations.

### DIFF
--- a/docs/documentation/getting_started/sharp-bits.md
+++ b/docs/documentation/getting_started/sharp-bits.md
@@ -12,6 +12,22 @@ import dynamiqs as dq
 
 The syntax in dynamiqs is similar to [QuTiP](http://qutip.org/), a popular Python library for quantum simulation. However, there are some important differences that you should be aware of.
 
+### Floating-point precision
+
+In dynamiqs, all arrays are represented with **single-precision** floating-point numbers (complex64 or float32) by default, whereas QuTiP or NumPy arrays are represented with double-precision floating-point numbers (complex128 or float64). This choice is made for **performance**, as most problems do not require the higher precision provided by double-precision floating-point numbers. However, if needed, it is possible to switch back to double precision to accommodate more demanding computational tasks, using:
+```python
+dq.set_precision('double') # 'simple' by default
+```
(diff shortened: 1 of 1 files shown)


docs/documentation/getting_started/sharp-bits.md
```diff
@@ -12,6 +12,22 @@ import dynamiqs as dq
...
+### Floating-point precision
+
+In dynamiqs, all arrays are represented with **single-precision** floating-point numbers (complex64 or float32) by default, whereas QuTiP or NumPy arrays are represented with double-precision floating-point numbers (complex128 or float64). This choice is made for **performance**, as most problems do not require the higher precision provided by double-precision floating-point numbers. However, if needed, it is possible to switch back to double precision to accommodate more demanding computational tasks, using:
+```python
+dq.set_precision('double') # 'simple' by default
+```
+
+With single-precision, there are certain limitations to be aware of:
+
+ - **Large numbers**: Single-precision floating-point numbers have a smaller range than double-precision floating-point numbers. This means that you may encounter overflow errors when working with very large numbers. Thus, it is advised to **set the scale of your problems such that the numbers involved are close to unity**.
```
pierreguilmin, line 24: > you may encounter overflow errors when working with very large numbers

This basically never happens in our simulations, the real issue is what we discussed on Slack (arithmetic precision depends on magnitude).

gautierronan: New tentative version in [d89adc2](https://github.com/dynamiqs/dynamiqs/pull/596/commits/d89adc2798835fa5a0fbac0cda9e1fbe765f39fe). Since the problem is not so clear in my mind, I went with a broad phrasing.

pierreguilmin:
I've made a few minor reformulations.

gautierronan:
Good to merge on my end @pierreguilmin

pierreguilmin (approved):

PR #597, opened 2024-05-31
Add `Rouchon1` solver

pierreguilmin:
Related to DYN-165.

Turned out to be pretty straightforward.

A few things left to do in subsequent PRs, I voluntarly kept this one as small as possible:
- [ ] `Id` is rebuilt at every time step, fix it (or maybe no need bc of magic caching?)
- [ ] check if magic caching works here too, or if `M0` and `Mks` are recomputed every time for constant Hamiltonians
- [ ] improve solver documentation (explain the update rule, link relevant papers)
- [ ] add normalization
- [ ] add `Rouchon2`

To define higher-order versions we just need to implement `kraus_map(t0, t1, y0) -> y1`. 😉

### DIFF
--- a/dynamiqs/mesolve/merouchon.py
+++ b/dynamiqs/mesolve/merouchon.py
@@ -1,0 +1,74 @@
+# ruff: noqa: ANN001, ANN201, ARG002
+# we mostly ignore type hinting in this file for readability purposes
+
+from __future__ import annotations
+
+import diffrax as dx
+import jax.numpy as jnp
+from diffrax._custom_types import RealScalarLike, Y
+from diffrax._local_interpolation import LocalLinearInterpolation
+from jax import Array
+
+from ..core.abstract_solver import MESolver
+from ..core.diffrax_solver import FixedSolver
+from ..utils.utils.general import dag
+
+
+class AbstractRouchonTerm(dx.AbstractTerm):
+    # this class bypasses the typical Diffrax term implementation, as Rouchon schemes
+    # don't match the vf/contr/prod structure
+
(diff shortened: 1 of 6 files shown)


pierreguilmin:
cc @lasellem

dynamiqs/mesolve/merouchon.py
```diff
@@ -0,0 +1,80 @@
...
+
+class RouchonSolver(dx.AbstractSolver):
+    term_structure = AbstractRouchonTerm
+    interpolation_cls = LocalLinearInterpolation
+
+    def init(self, terms, t0, t1, y0, args):
+        pass
+
+    def order(self, terms):
+        return 1
```
gautierronan, line 42: I'm guessing this should be moved to subclasses (e.g. MERouchon1) since there also exist high-order solvers.

dynamiqs/mesolve/merouchon.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def step(self, terms, t0, t1, y0, args, solver_state, made_jump):
+        del solver_state, made_jump
```
gautierronan, line 45:
```suggestion
        del solver_state, made_jump, args
```
(unsure about this one, but it's not used either)

gautierronan (changes requested):
Two minor comments, otherwise it looks great! Thanks :)

dynamiqs/mesolve/merouchon.py
pierreguilmin: We could do this but it's useless as far as I understand (args is an empty tuple, we won't gain anything by deleting it).

pierreguilmin:
I've moved the remaining things to do to DYN-165.

PR #600, opened 2024-06-07
Specify time-arrays discontinuities in diffrax solvers

pierreguilmin:
Fixes DYN-218.

Thanks to @xiasitan for spotting the bug. 😉

## MWE
```python
import dynamiqs as dq
import jax
import jax.numpy as jnp

# PWCTimeArray
times = [0, 1, 1.1, 2]
values = [0, 10, 0]
H = dq.pwc(times, values, -dq.sigmax())
result = dq.sesolve(H, dq.ground(), [0, 2])
print(result)
print(dq.expect(dq.sigmaz(), result.states[-1]).real)

# ModulatedTimeArray
f = lambda t: jax.lax.cond(
    jnp.logical_and(1 <= t, t < 1.1), lambda _: 10, lambda _: 0, None
)
H = dq.modulated(f, -dq.sigmax(), discontinuity_ts=[1.0, 1.1])
result = dq.sesolve(H, dq.ground(), [0, 2])
print(result)
print(dq.expect(dq.sigmaz(), result.states[-1]).real)
```

Before this PR (wrong result):
```
# 5 steps
# -1.0
# 5 steps
# -1.0
```

After this PR (correct result):
```
# 15 steps
# 0.4161431
# 15 steps
# 0.4161431
```

## Old comment
Note that this PR breaks the JIT. The reason is that `jnp.unique` is data-dependent, see doc here > https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.unique.html. We have three solutions:
1. Either `jump_ts` in Diffrax does not need an array with unique values, that would be ideal.
2. Otherwise we coud specify statically the `size` argument of `jnp.unique`, which is fine if the `H`, `jump_ops` are defined outside JIT, we can just feed `len(times)`.
3. Otherwise we could specify a default `size` to `jnp.unique` (specifically made for this purpose), but then we need to pick a `fill_value`, and we need to check what fits Diffrax.

Left to do to fully close the issue after we've fixed the JIT:
- [x] set `discontinuities_ts` to private or document in `TimeArray` API
- [x] pull `discontinuities_ts` up to the `modulated`/`timecallable` API
- [x] document this in the time-dependent tutorial

Note that for `modulated` _we could also_ automatically scan the function to detect discontinuities, but that's a bit niche, I think a first solution where we delegate to the user is good enough for now.

### DIFF
--- a/dynamiqs/time_array.py
+++ b/dynamiqs/time_array.py
@@ -122,6 +122,7 @@ def modulated(
     # discontinuity_ts
     if discontinuity_ts is not None:
         discontinuity_ts = jnp.asarray(discontinuity_ts)
+        discontinuity_ts = jnp.sort(discontinuity_ts)
 
     # make f a valid PyTree that is vmap-compatible
     f = BatchedCallable(f)
@@ -163,6 +164,7 @@ def timecallable(
     # discontinuity_ts
     if discontinuity_ts is not None:
         discontinuity_ts = jnp.asarray(discontinuity_ts)
+        discontinuity_ts = jnp.sort(discontinuity_ts)
 
     # make f a valid PyTree that is vmap-compatible
     f = BatchedCallable(f)
@@ -236,6 +238,7 @@ class TimeArray(eqx.Module):
     @property
     @abstractmethod
     def discontinuity_ts(self) -> Array | None:
+        # must be sorted, not necessarily unique values
         pass
 
     @abstractmethod
--- a/dynamiqs/time_array.py
+++ b/dynamiqs/time_array.py
@@ -84,7 +84,12 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
     return PWCTimeArray(times, values, array)
 
 
-def modulated(f: callable[[float], Array], array: ArrayLike) -> ModulatedTimeArray:
+def modulated(
+    f: callable[[float], Array],
+    array: ArrayLike,
+    *,
+    discontinuity_ts: ArrayLike | None = None,
+) -> ModulatedTimeArray:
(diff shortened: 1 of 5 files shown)


pierreguilmin:
I'm also thinking, for the propagator solver to work with PWC pulse ([DYN-47](https://linear.app/dynamiqs/issue/DYN-47/use-pwc-hamiltonian-format-in-propagator-solvers)), we'll probably need this uniquely sorted time-array anyway.

pierreguilmin:
Turns out the easy solution works, thanks @gautierronan for looking into this!
> 1. Either jump_ts in Diffrax does not need an array with unique values, that would be ideal.

dynamiqs/time_array.py
```diff
@@ -167,6 +186,9 @@ class TimeArray(eqx.Module):
         mT _(TimeArray)_: Returns the time-array transposed over its last two
             dimensions.
         ndim _(int)_: Number of dimensions.
+        discontinuity_ts _(Array | None)_: Times at which there is a discontinuous jump
+            in the time-array values (the array is always sorted, but does not
+            necessarily contain unique values).
```
gautierronan, line 191: Related to the main comment. I don't think the array is necesseraly sorted at this point.

gautierronan (changes requested):
Just a single comment: I think we need to sort the `discontinuity_ts` array somewhere inside the `time_array.py` file. Right now, I believe we can pass an unsorted array and it will never be sorted if `sesolve` is used. For `mesolve`, it goes through `_concatenate_sort` so its fine.

The rest looks good to me!

pierreguilmin:
> Just a single comment: I think we need to sort the discontinuity_ts array somewhere inside the time_array.py file.

Good catch!!

PR #602, opened 2024-06-09
Add support for scalar in ModulatedTimeArray

gautierronan:
Went for the easy solution with a simple `f = lambda t: jnp.asarray(f(t))`. Did not check, but I'm pretty sure this should be transparent with the JIT.

Closes DYN-220.


pierreguilmin (approved):

pierreguilmin (dismissed):
Great thanks!

pierreguilmin (approved):

PR #603, opened 2024-06-09
Add examples to timearray creation utilities

gautierronan:
A few simple examples in the API, complementary to the corresponding tutorial.

### DIFF
--- a/dynamiqs/time_array.py
+++ b/dynamiqs/time_array.py
@@ -75,15 +75,22 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
             returning $O(t)$ for any time $t$.
 
     Examples:
-        >>> times = jnp.linspace(0.0, 1.0, 11)
-        >>> values = jnp.linspace(1.0, 10.0, 10)
-        >>> H = dq.pwc(times, values, dq.sigmaz())
+        >>> times = [0.0, 1.0, 2.0]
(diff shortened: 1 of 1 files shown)


dynamiqs/time_array.py
```diff
@@ -138,6 +168,16 @@ def timecallable(f: callable[[float], Array]) -> CallableTimeArray:
...
+
+    Examples:
+        >>> f = lambda t: jnp.cos(jnp.pi * t) * dq.sigmaz()
+        >>> H = dq.timecallable(f)
+        >>> H(0.0)
+        Array([[ 1.+0.j,  0.+0.j],
+               [ 0.+0.j, -1.+0.j]], dtype=complex64)
+        >>> H(1.0)
+        Array([[-1.+0.j, -0.+0.j],
+               [-0.+0.j,  1.-0.j]], dtype=complex64)
```
pierreguilmin, line 180: We should **never** give an example for timecallable that is actually doable with `dq.modulated`, this is misleading on what `timecallable` should be used for. We can for example us the same example as the one in time-dependent operators.

pierreguilmin (changes requested):
Great! Wdyt about exactly copy-pasting the examples from the time-dep operators tutorial? (including for PWC showing the result at all times, I think it really clarifies what's happening with the bound specified by `times`) I think we should avoid varying documentation serving the same purpose, to reduce later maintenance work.

pierreguilmin (approved):

PR #616, opened 2024-06-21
Merge and clean up SparseQArray with the rest of QArray

gautierronan:
I suggest reading commit by commit: it's a bunch of minor stuff meant to uniformize the implementation with the rest of QArray.

### DIFF
--- a/dynamiqs/time_array.py
+++ b/dynamiqs/time_array.py
@@ -78,6 +78,7 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
         )
 
     # array
+    array = jnp.asarray(array, dtype=cdtype())
     check_shape(array, 'array', '(n, n)')
 
     return PWCTimeArray(times, values, array)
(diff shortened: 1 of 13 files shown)


pierreguilmin (commented):
Apart from the discussion on wheter we should revert 192a341 (which would modify this PR), happy to merge.

gautierronan:
Added a few more commits, all rather small. Ideally we have tests passing for `tests/qarrays/test_sparse_qarray.py` which is still not the case. 

> Apart from the discussion on wheter we should revert [192a341](https://github.com/dynamiqs/dynamiqs/commit/192a34147a23b1580cc6b366f3a0d9ff451dc1a5) (which would modify this PR), happy to merge.

We can revert in a separate PR, but we should discuss the sparse API before that IMO.

dynamiqs/time_array.py
```diff
@@ -78,7 +78,6 @@ def pwc(times: ArrayLike, values: ArrayLike, array: ArrayLike) -> PWCTimeArray:
         )
 
     # array
-    array = jnp.asarray(array, dtype=cdtype())
```
pierreguilmin, line 81: We need this back I think.

dynamiqs/time_array.py
```diff
@@ -109,7 +108,6 @@ def modulated(f: callable[[float, ...], Array], array: ArrayLike) -> ModulatedTi
...
```
pierreguilmin, line 112: Also here.

pierreguilmin (changes requested):

dynamiqs/time_array.py
gautierronan: Good catch

PR #622, opened 2024-06-25
Rename `SparseQArray` to `SparseDIAQArray`

pierreguilmin:


abocquet (changes requested):
We are not planning to implement another format for now so there is no need to specify this I think

gautierronan:
> We are not planning to implement another format for now so there is no need to specify this I think

Yes, but if we do implement another sparse format, it's easier to make this change now than to have to do it later on (might break the API).

dynamiqs/qarrays/sparse_dia_qarray.py
```diff
@@ -298,16 +298,16 @@ def _construct_diags(offsets: tuple[int, ...], other: ArrayLike) -> Array:
     return diags
 
 
-def to_sparse(x: DenseQArray | Array) -> SparseQArray:
-    r"""Returns the input matrix in the `SparseQArray` format.
+def to_sparse_dia(x: DenseQArray | Array) -> SparseDIAQArray:
```
gautierronan, line 301: I am conflicted on this API. WDYT about `def to_sparse(..., format=dq.SparseQArray) -> QArray` or something like that?

pierreguilmin: Meh, why don't you like this version? The other sparse formats may need extra constructor arguments that we are not yet aware of.

PR #625, opened 2024-06-27
Integrate `QArray` in creation utils

pierreguilmin:
On top of #624.

I ported
- [x] `dynamiqs/dark/dark.py`
- [x] `dynamiqs/utils/states.py`
- [x] `dynamiqs/utils/operators.py`
- [x] `dynamiqs/utils/optimal_control.py`

Remaining things from these files (task added to https://linear.app/dynamiqs/issue/DYN-210/qarray-object):
- [ ] `fock()` is bugged for batched number due to the call to `vectorize`

PR #626, opened 2024-06-27
Fix docstrings utils with new `QArray` string `__repr__`

pierreguilmin:
On top of #625. Some doctests are still breaking, basically anytime we pass a list of `QArray`s to `dq.asqarray`. Also fixed `to_qutip` to catch the `QArray` dims.


gautierronan (approved):
LGTM

PR #627, opened 2024-07-03
Filter FutureWarning: unhashable type

gautierronan:
Temporary fix for the warnings coming from diffrax + jax 0.4.30.
```
~/miniconda3/lib/python3.11/site-packages/jax/_src/core.py:678: FutureWarning: unhashable type: <class 'jax._src.interpreters.partial_eval.DynamicJaxprTracer'>. Attempting to hash a tracer will lead to an error in a future JAX release.
  warnings.warn(
```

See also https://github.com/patrick-kidger/diffrax/issues/445.

Closes DYN-223.


pierreguilmin:
Great, thanks for investigating 🙏

PR #630, opened 2024-07-17
Update mesolve vector_field for QArrays

gautierronan:
On top of https://github.com/dynamiqs/dynamiqs/pull/620.

This PR makes the following simple `mesolve` example run without errors:
```python
import dynamiqs as dq
import jax.numpy as jnp

H = dq.number(8)
jump_ops = [dq.destroy(8)]
tlist = jnp.linspace(0.0, 1.0, 10)
psi0 = dq.basis(8, 0)
result = dq.mesolve(H, jump_ops, psi0, tlist)
```

In practice, this only requires modifying the `vector_field` function to avoid the call to `jnp.stack`. It now uses a simple `for` loop. Below are several benchmarks to show that this doesn't impact performance.

<details>
<summary>Benchmark sums of matmuls</summary>

```python
import jax.numpy as jnp
import jax
import timeit

# benchmark sum of matmuls
@jax.jit
def f(A, B, C):
    tmp = jnp.zeros_like(A)
    for x in [A, B, C]:
        tmp += x @ x
    return tmp

@jax.jit
def g(A, B, C):
    x = jnp.stack([A, B, C])
    return jnp.sum(x @ x, axis=0)

@jax.jit
def h(A, B, C):
    x = [A, B, C]
    return sum([x_ @ x_ for x_ in x])

N = 1000
A = jnp.ones((N, N))
B = jnp.ones((N, N))
C = jnp.ones((N, N))

print(jnp.allclose(f(A, B, C), g(A, B, C)))
print(jnp.allclose(f(A, B, C), h(A, B, C)))

%timeit f(A, B, C).block_until_ready() # 46.6 ms ± 2.28 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit g(A, B, C).block_until_ready() # 55 ms ± 8.29 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit h(A, B, C).block_until_ready() # 51.2 ms ± 7.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

</details>

<details>
<summary>Benchmark vector_field variants</summary>

```python
import jax.numpy as jnp
import jax
import timeit

# benchmark vector_field
@jax.jit
def vf1(H, Ls, y):
    tmp = -1j * H @ y
    for L in Ls:
        L_dag = L.conj().mT
        L_y = L @ y
        tmp += 0.5 * (L_y @ L_dag - L_dag @ L_y)
    return tmp + tmp.conj().mT

[...]

PR #631, opened 2024-07-17
Convert TimeArray to QArrays

gautierronan:
On top of https://github.com/dynamiqs/dynamiqs/pull/630.

Conversion of the `time_array.py` file to use QArrays. Nothing super surprising: the main difficulty is that we need to use `isqarraylike` instead of `isinstance(x, QArrayLike)` because we specify list content in the typing.

The following now runs without erroring:
```python
import dynamiqs as dq
import jax.numpy as jnp

H = dq.number(8)
H += dq.modulated(jnp.sin, dq.destroy(8) + dq.create(8))
jump_ops = [dq.destroy(8)]
tlist = jnp.linspace(0.0, 1.0, 10)
psi0 = dq.basis(8, 0)
result = dq.mesolve(H, jump_ops, psi0, tlist)
```


dynamiqs/time_array.py
```diff
@@ -362,11 +364,11 @@ def broadcast_to(self, *new_shape: int) -> TimeArray:
     def conj(self) -> TimeArray:
         return PWCTimeArray(self.times, self.values.conj(), self.array.conj())
 
-    def __call__(self, t: ScalarLike) -> Array:
-        def _zero(_: float) -> Array:
+    def __call__(self, t: ScalarLike) -> QArray:
+        def _zero(_: float) -> QArray:
```
pierreguilmin, line 368: This method should return a `Qarray`.

dynamiqs/time_array.py
```diff
@@ -529,40 +531,40 @@ def broadcast_to(self, *new_shape: int) -> TimeArray:
     def conj(self) -> TimeArray:
         return SummedTimeArray([tarray.conj() for tarray in self.timearrays])
 
-    def __call__(self, t: ScalarLike) -> Array:
+    def __call__(self, t: ScalarLike) -> QArray:
         return jax.tree_util.tree_reduce(
-            jnp.add, [tarray(t) for tarray in self.timearrays]
+            lambda x, y: x + y, [tarray(t) for tarray in self.timearrays]
         )
```
pierreguilmin, line 537: Does this return a `QArray`?

dynamiqs/qarrays/dense_qarray.py
```diff
@@ -116,11 +114,11 @@ def __repr__(self) -> str:
     def __mul__(self, y: QArrayLike) -> QArray:
         super().__mul__(y)
 
-        if isinstance(y, get_args(ScalarLike)):
+        if isinstance(y, ScalarLike):
```
pierreguilmin, line 117: Why removing `get_args`?
```python
>>> from jaxtyping import ScalarLike
>>> isinstance(1, ScalarLike)
TypeError: Subscripted generics cannot be used with class and instance checks
```

dynamiqs/qarrays/qarray.py
```diff
@@ -327,7 +328,7 @@ def __neg__(self) -> QArray:
     @abstractmethod
     def __mul__(self, y: QArrayLike) -> QArray:
         """Element-wise multiplication with a scalar or an array."""
-        if not isinstance(y, get_args(ScalarLike)):
+        if not isinstance(y, ScalarLike) and not isinstance(y, DynamicJaxprTracer):
```
pierreguilmin, line 331: Could you explain this?

pierreguilmin (dismissed):
Minor changes. Could you also review all the arithmetic methods by keeping in mind that `QArrayLike` can now be e.g. a list of list? (so for example `-y` doesn't work).

pierreguilmin:
Also we should rename this to `TimeQArray`, in another PR to ease the review.

pierreguilmin:
> Also we should rename this to `TimeQArray`, in another PR to ease the review.

And also rename the attributes like `array` to `qarray`.

abocquet (approved):

PR #636, opened 2024-07-23
Codebase restructuration

gautierronan:
This PR restructures the codebase following issues with the `sepropagator` PR (see https://github.com/dynamiqs/dynamiqs/pull/629#discussion_r1686441820).

With this new structure, we have:
- renamed internal solvers to "integrators"
- used verbose names for all integrators, to avoid name collisions
- fixed imports to allow importing private functions inside `sesolve.py`, `mesolve.py`, etc.

Also, all unique combinations of a Solver (`Euler`, `Propagator`, `Dopri5`, ...) and a Core function (`sesolve`, `mesolve`, ...) will yield a unique Integrator, and reciprocally.

You can review commit by commit. There are no logic changes, only restructuration and import modifications.


dynamiqs/integrators/mesolve/rouchon_integrator.py
```diff
@@ -31,7 +31,7 @@ def prod(self, vf, control):
         pass
 
 
-class RouchonSolver(dx.AbstractSolver):
+class RouchonDXSolver(dx.AbstractSolver):
```
gautierronan, line 34: Note: this inherits from `dx.AbstractSolver` so I classified this as a `DXSolver` (another object separate from solvers or from integrators)

pierreguilmin (commented):
Looks amazing 👏 We just need to fix the doc rendering. Also can we match what we do for the rest of the code for the `__init__`, where each module is responsible for exposing what needs to be? So:
- in `dynamiqs/__init__.py` do a `from .integrators import *`
- in `dynamiqs/integrators/__init__.py` do a `from .apis import *`
- in `dynamiqs/integrators/api/__init__.py` do
```python
from .sesolve import sesolve
from .mesolve import mesolve
from .smesolve import smesolve

__all__ = ['sesolve', 'mesolve', 'smesolve']
```

gautierronan:
> Looks amazing 👏 We just need to fix the doc rendering. Also can we match what we do for the rest of the code for the `__init__`, where each module is responsible for exposing what needs to be? So:

I intentionally made it differently in `dynamiqs/integrators/__init__.py` to allow for importing `_mesolve` or `_vectorized_mesolve` with a `from dq.integrators.apis.mesolve import _mesolve`. If we keep the same imports as usual, this is impossible because the file name `mesolve.py` has the same name as the function `mesolve()`.

I'll try to keep something a bit closer to the usual in the following commit.

gautierronan:
Ok, so https://github.com/dynamiqs/dynamiqs/pull/636/commits/84a134e888e0bd60889ff622089728bd79143d91 fixes the documentation, and also raises the `__all__ = ['mesolve', 'sesolve', 'smesolve']` to `dynamiqs/integrators/__init__.py` to allow private function importing.

pierreguilmin:
Beautiful! The only renaming thing is to reorganise the tests (file hierarchy and naming).

pierreguilmin:
But we can do it in another PR, let's merge this one 🎉

pierreguilmin (approved):

PR #637, opened 2024-07-24
Add `sepropagator` solver 

dkweiss31:
follow up to https://github.com/dynamiqs/dynamiqs/pull/629 after the api restructuring

Note that I slightly deviated from the suggested implementation. I did not feel it was necessary to implement the classes `SEPropagatorEulerIntegrator, SEPropagatorDopri5Integrator,...` because it is `sesolve` that is actually doing the integration. From a class inheritance perspective, there is no reason why `SEPropagatorDiffraxIntegrator` (which maybe we do want to rename to `SEPropagatorDynamiqsIntegrator`?) should inherit from `DiffraxIntegrator`, since `SEPropagatorDiffraxIntegrator` will not be implementing a `terms` method. Very happy to reconsider this decision if people feel strongly :).

PR #643, opened 2024-07-28
Improve `sepropagator`

pierreguilmin:
Improvements on top of #637. I advise reading commits by commits, and especially last commit introduces a bunch of modifications.

Here's what remains to be done performance wise:
- [[DYN-228](https://linear.app/dynamiqs/issue/DYN-228)] Except if I'm missing something, propagators for time difference of 0 are as costly to compute as the others, we should explicitely skip these times.
    ```python
    import jax
    import dynamiqs as dq
    
    @jax.jit
    def expm(t, x):
        return dq.expm(t * x)
    
    key = jax.random.PRNGKey(42)
    x = dq.rand_dm(key, (256, 256))
    
    %timeit -r1 -n1 -q expm(0, x)
    %timeit expm(1, x)  # 60.2 ms ± 17.4 ms
    %timeit expm(0, x)  # 70.3 ms ± 38.4 ms
    ```
- [[DYN-229](https://linear.app/dynamiqs/issue/DYN-229)] We could reduce propagators more efficiently than sequentially from left to right, e.g. with a divide-and-conquer strategy (cost would go from $\mathcal{O}(n_\text{steps} n^3) \to \mathcal{O}(\log(n_\text{steps}) n^3)$ ).
- [[DYN-201](https://linear.app/dynamiqs/issue/DYN-201)] Caching to avoid computing the propagators for the same $\Delta t$. Note that this is not as relevant as before because this step is vectorized anyway, but it would be quite relevant for the ME.

@dkweiss31 I think the best course of action is to open a PR grouping all sesolve/mesolve/sepropagator `Expm` solvers (with support for PWC Hamiltonian [DYN-47](https://linear.app/dynamiqs/issue/DYN-47) and batch-compute of the exponential [DYN-202](https://linear.app/dynamiqs/issue/DYN-202)) based on this one. Then the first speedup is probably quite easy to implement, and we can keep the next two for later times.


dynamiqs/integrators/sepropagator/diffrax_integrator.py
```diff
@@ -0,0 +1,46 @@
...
+    Kvaerno3Integrator,
+    Kvaerno5Integrator,
+    Tsit5Integrator,
+)
+
+
+class SEPropagatorDiffraxIntegrator(DiffraxIntegrator, SEPropagatorIntegrator):
+    @property
+    def terms(self) -> dx.AbstractTerm:
+        # define Schrödinger term d|psi>/dt = - i H |psi>
```
dkweiss31, line 20: I would change to 
```python
# define Schrödinger term dU/dt = - i H U
``` 
in this context

dynamiqs/integrators/sepropagator/expm_integrator.py
```diff
@@ -26,46 +25,33 @@ def __init__(self, *args):
...
-        times = _concatenate_sort(t0, self.H.discontinuity_ts, self.ts)
-        inbounds = (times[:-1] >= t0) & (times[1:] <= self.ts[-1])
-        t_diffs = jnp.where(inbounds, jnp.diff(times), 0.0)
+        # === find all times at which to stop in [t0, t1]
+        # find all times where the solution should be saved (self.ts) or at which the
+        # Hamiltonian changes (self.H.discontinuity_ts)
+        disc_ts = self.H.discontinuity_ts
+        if disc_ts is not None:
+            disc_ts = disc_ts.clip(self.t0, self.t1)
+        times = _concatenate_sort(jnp.asarray([self.t0]), self.ts, disc_ts)  # (ntimes,)
```
dkweiss31, line 34: What is the reason for `jnp.asarray([self.t0])` as opposed to `jnp.asarray(self.t0).reshape(-1)`? I used the latter because of the comment here https://github.com/dynamiqs/dynamiqs/blob/main/dynamiqs/integrators/core/propagator_integrator.py#L50 .

dynamiqs/integrators/sepropagator/expm_integrator.py
```diff
@@ -26,46 +25,33 @@ def __init__(self, *args):
...
 
-        # === combine propagators to get the propagator at each requested time
-        def _reduce(prev_prop: Array, next_prop: Array) -> Array:
-            # notice the ordering of prev_prop and next_prop, want
-            # next_prop to be to the left of prev_prop
-            total_prop = next_prop @ prev_prop
-            return total_prop, total_prop
+        # === batch-compute the propagators on each time interval
+        Hs = jax.vmap(self.H)(times[:-1])  # (ntimes-1, n, n)
+        step_propagators = expm(-1j * delta_ts[:, None, None] * Hs)  # (ntimes-1, n, n)
```
dkweiss31, line 41: I think this is not quite right and will fail when Hs is batched. I confirmed this by changing the batch dimensions for H in the tests from (3,) and (4,) to (4,) and (4, 5) (unfortunate collision between batch dimension and tsave dimension in the tests) I would suggest instead
```suggestion
        dims = tuple(range(-Hs.ndim + 1, 0))
        delta_ts = jnp.expand_dims(delta_ts, dims)
        step_propagators = expm(-1j * delta_ts * Hs)  # (ntimes-1, nH, n, n)
```

dynamiqs/integrators/sepropagator/expm_integrator.py
```diff
@@ -26,46 +25,33 @@ def __init__(self, *args):
...
-        t_idxs = jnp.argmin(jnp.abs(times - self.ts[:, None]), axis=1)
-        t_idxs = jnp.where(t_idxs > 0, t_idxs - 1, t_idxs)
-        propagators = propagators_for_times[t_idxs]
-        # note that we can't take the output of scan as final_prop, because
-        # it could correspond to a time window of H.times. However
-        # the final element of propagators will correspond to the propagator
-        # at the final time
-        final_prop = propagators[-1]
-        propagators = jnp.moveaxis(propagators, 0, -3)
+        _, Us = jax.lax.scan(step, self.y0, step_propagators)  # (ntimes-1, n, n)
```
dkweiss31, line 49:
```suggestion
        _, Us = jax.lax.scan(step, self.y0, step_propagators)  # (ntimes-1, nH, n, n)
```

dynamiqs/integrators/sepropagator/expm_integrator.py
```diff
@@ -26,46 +25,33 @@ def __init__(self, *args):
...
+
+        # === extract the propagators at the save times ts
+        Us = Us[jnp.searchsorted(times[1:], self.ts)]  # (nts, n, n)
```
dkweiss31, line 52:
```suggestion
        Us = Us[jnp.searchsorted(times[1:], self.ts)]  # (nts, nH, n, n)
```

dynamiqs/integrators/sepropagator/diffrax_integrator.py
dkweiss31, line 20: I would change to 
```suggestion
        # define Schrödinger term dU/dt = - i H U
```
in this context

dkweiss31:
These look great @pierreguilmin thank you! I've left some comments. Regarding the first issue of propagators for delta_t=0.0, I can reproduce your results. The below first attempt to conditionally compute the exponential doesn't seem to offer speedup...
```python
import jax
import jax.numpy as jnp
from dynamiqs.utils.operators import eye
from dynamiqs.utils.utils.general import expm
from dynamiqs import rand_herm


@jax.jit
def conditionally_apply_expm(delta_t, H_for_delta_t):
    return jax.lax.cond(
        delta_t != 0.0,
        lambda t, H: expm(-1j * t * H),
        lambda t, H: eye(n),
        delta_t,
        H_for_delta_t
    )

@jax.jit
def unconditionally_apply_expm(delta_t, H_for_delta_t):
    return expm(-1j * delta_t * H_for_delta_t)

n = 256
key = jax.random.PRNGKey(42)
H = rand_herm(key, shape=(n, n))
tsave = jnp.linspace(0.0, 10.0, 11)
tsave = jnp.sort(jnp.concatenate((tsave, tsave)))
delta_ts = jnp.diff(tsave)

%timeit -r1 -n1 jax.vmap(conditionally_apply_expm, in_axes=(0, None))(delta_ts, H)
%timeit -r1 -n1 jax.vmap(unconditionally_apply_expm, in_axes=(0, None))(delta_ts, H)

%timeit jax.vmap(conditionally_apply_expm, in_axes=(0, None))(delta_ts, H)  # 787 ms ± 72.3 ms
%timeit jax.vmap(unconditionally_apply_expm, in_axes=(0, None))(delta_ts, H)  # 723 ms ± 13.4
``` 

> @dkweiss31 I think the best course of action is to open a PR grouping all sesolve/mesolve/sepropagator `Expm` solvers (with support for PWC Hamiltonian [DYN-47](https://linear.app/dynamiqs/issue/DYN-47) and batch-compute of the exponential [DYN-202](https://linear.app/dynamiqs/issue/DYN-202)) based on this one. Then the first speedup is probably quite easy to implement, and we can keep the next two for later times.

Sounds great, do you prefer that next PR to be done concurrent with this one or I can wait until this is merged?

dynamiqs/integrators/sepropagator/expm_integrator.py
dkweiss31: It goes without saying we should change the batch dims in the tests!

dynamiqs/integrators/sepropagator/expm_integrator.py
pierreguilmin: Both options work in this context , I chose `jnp.asarray([self.t0])` so that it fits on a single line 😇 Ideally we should convert `t0` upfront in `dq.Options` so we don't have to worry about this downstream.

dynamiqs/integrators/sepropagator/expm_integrator.py
pierreguilmin: Because the code is vectorized in `_vectorized_sepropagator` we don't need to care about batch dimensions for `H`, so we can write the code for a single `H` of shape `(n, n)` and the magic of vmap will do the rest. It's also the reason why I don't indicate the batching in the shape comments.

I edited the test with your suggestion to convince you. 😈

pierreguilmin:
> The below first attempt to conditionally compute the exponential doesn't seem to offer speedup...

Mmm, that's odd.

> Sounds great, do you prefer that next PR to be done concurrent with this one or I can wait until this is merged?

We can wait for this one to be merged. 👍

pierreguilmin:
Now that I think about it, we could vectorized the computation in the ODE case as well by splitting the time into chunks. 🥲 This is even advantageous for `sesolve` with time-dependent Hamiltonian if the number of time steps is greater than `n`, because you go from $\mathcal{O}(n_\text{steps}n^2)\to \mathcal{O}(\log(n_\text{steps})n^3)$. But it's a mess to choose the chunks size, and not easy for the adaptive stepping, so I'd suggest we ignore this comment. 😇 cc @gautierronan

pierreguilmin:
@dkweiss31 I can't wrap my head around why what I wrote
```python
Us = Us[jnp.searchsorted(times[1:], self.ts)]  # (nts, n, n)
```
is correct, can you confirm there's no mistake here? 😅

dynamiqs/integrators/sepropagator/expm_integrator.py
dkweiss31: Ahha! You are right!

dkweiss31:
> @dkweiss31 I can't wrap my head around why what I wrote
> 
> ```python
> Us = Us[jnp.searchsorted(times[1:], self.ts)]  # (nts, n, n)
> ```
> 
> is correct, can you confirm there's no mistake here? 😅

I think this looks right to me and is just simpler and more elegant than what I had before. Taking the slice `times[1:]` takes care of the "-1 issue" whereby indices in `times` and indices in `Us` are offset by 1. Moreover we are guaranteed to always have at least two entries of `t0` in times (and even if some values of `discontinuity_times` are earlier than `t0`, they get clipped to `t0`), so `times[1]==t0`. For this entry we then ask for `Us[0]`, which is the identity matrix 👍 (because there are at least two entries of `t0` in times!). So, it looks good to me 😄

pierreguilmin:
> we are guaranteed to always have at least two entries of t0 in times

I think that's not true if a user specifies `t0` in `dq.Options` which is not included in `ts`.

pierreguilmin:
Great, so if @gautierronan could quickly review I think we're ready to merge. 😊

gautierronan:
> Great, so if @gautierronan could quickly review I think we're ready to merge. 😊

Ok, will review before the end of the week!

dynamiqs/integrators/sepropagator/expm_integrator.py
```diff
@@ -26,46 +25,33 @@ def __init__(self, *args):
...
+        delta_ts = jnp.diff(times)  # (ntimes-1,)
 
-        # === combine propagators to get the propagator at each requested time
-        def _reduce(prev_prop: Array, next_prop: Array) -> Array:
-            # notice the ordering of prev_prop and next_prop, want
-            # next_prop to be to the left of prev_prop
-            total_prop = next_prop @ prev_prop
-            return total_prop, total_prop
+        # === batch-compute the propagators on each time interval
+        Hs = jax.vmap(self.H)(times[:-1])  # (ntimes-1, n, n)
```
gautierronan, line 40: That's a cool one. Never thought about doing this!

gautierronan (approved):
LGTM. Very nice PR!

PR #664, opened 2024-08-02
Move rand_* to dq.rand.*

abocquet:
Closes https://linear.app/dynamiqs/issue/DYN-232/move-dqrand-to-dqrand


pierreguilmin (changes requested):
Can we reverse the naming `dq.rand.imag` to `dq.rand.complex`? I read `imag` as imaginary part at first, similar to `x.real`, `x.imag`, or *imaginary number*.

abocquet:
Unfortunately `complex` is a reserved keyword of python, so I had to come up with something else. I get your point though, and any better name is welcome

pierreguilmin (approved):

pierreguilmin:
Great 😁

PR #665, opened 2024-08-06
Get pwc or modulated time array prefactor value

dkweiss31:
It is useful in many contexts to have the ability to natively ask about the value of the function, not just the overall Hamiltonian. Certainly people will have opinions on how best to implement this, this PR is mostly to get the discussion rolling

### DIFF
--- a/dynamiqs/time_array.py
+++ b/dynamiqs/time_array.py
@@ -413,9 +413,6 @@ class ConstantTimeArray(TimeArray):
     def __call__(self, t: ScalarLike) -> Array:  # noqa: ARG002
         return self.array
 
-    def pulse_value_at_time(self, t: ScalarLike) -> Array:
-        return self.__call__(t)
-
     def __mul__(self, y: ArrayLike) -> TimeArray:
         return ConstantTimeArray(self.array * y)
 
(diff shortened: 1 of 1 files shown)


pierreguilmin:
Tell me if you agree first:
1. It should only be implemented for `PWCTimeArray` and `ModulatedTimeArray`, it does not make much sense for constant and callable time arrays IMO.
2. I would rename the function `factor(t)` (the notion of "pulse" is a bit oriented towards control, you can also have e.g. drifting Hamiltonians with no "pulse").

dkweiss31:
> Tell me if you agree first:
> 
> 1. It should only be implemented for `PWCTimeArray` and `ModulatedTimeArray`, it does not make much sense for constant and callable time arrays IMO.
> 2. I would rename the function `factor(t)` (the notion of "pulse" is a bit biased towards control, you can also have e.g. drifting Hamiltonians with no "pulse").

Totally agreed on point 1. I mostly added them to callable and constant to be consistent, but I agree that they only really make sense for modulated and pwc. 

Re point 2, agreed that "pulse" is not good. Though "factor" might be a bit vague IMO, maybe "prefactor" or "function_value"?

pierreguilmin (approved):
@dkweiss31 you can merge if you agree with the proposed change.

PR #673, opened 2024-08-09
Nits fight back

pierreguilmin:
To be continued...

I advise reading commit by commit.

### DIFF
--- a/dynamiqs/integrators/core/diffrax_integrator.py
+++ b/dynamiqs/integrators/core/diffrax_integrator.py
@@ -202,6 +202,8 @@ class MEDiffraxIntegrator(DiffraxIntegrator):
         def vector_field(t, y, _):  # noqa: ANN001, ANN202
             tmp = -1j * self.H(t) @ y
             Ls = [L(t) for L in self.Ls]
+            # note: benchmarks show that using stacking or a for loop yields equivalent
+            # execution speeds for this code segment
             for L in Ls:
                 Ldag = L.dag()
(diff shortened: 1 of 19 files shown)


dynamiqs/integrators/core/diffrax_integrator.py
```diff
@@ -201,11 +201,11 @@ def terms(self) -> dx.AbstractTerm:
...
-            for L in self.Ls:
-                Lt = L(t)
-                Ltdag = Lt.dag()
-                Lt_y = Lt @ y
-                tmp += 0.5 * (Lt_y @ Ltdag - Ltdag @ Lt_y)
+            Ls = [L(t) for L in self.Ls]
+            for L in Ls:
+                Ldag = L.dag()
+                L_y = L @ y
+                tmp += 0.5 * (L_y @ Ldag - Ldag @ L_y)
```
gautierronan: I don't think this commit is useful. You're storing all L(t) values instead of storing them one by one. The code is more elegant, but I don't think it's faster and it's less memory-efficient.

But in any case, I'll try to rework this implementation to find something faster.

gautierronan (approved):
LGTM. Small comment below but I'm still merging because I'll rework this at some point.

dynamiqs/integrators/core/diffrax_integrator.py
pierreguilmin: Yes, it was only for style 😇

PR #675, opened 2024-08-09
Fix some of the doctests

pierreguilmin:
On top of #673.

Fix 27/68 failing, we now have 278 passing 41 failed, mostly due to `SparseDIAQArray` print format.

PR #676, opened 2024-08-10
Sparse dia `__repr__` and fix associated doctest

pierreguilmin:
On top of #675.

Very satisfying 😋
```python
>>> import dynamiqs as dq
>>> import jax.numpy as jnp
>>> jnp.set_printoptions(precision=3)
>>> a = dq.destroy(6)
>>> a.dag() @ a + a @ a @ a + a.dag()
SparseDIAQArray: shape=(6, 6), dims=(6,), dtype=complex64, ndiags=3
[[    ⋅         ⋅         ⋅     2.449+0.j     ⋅         ⋅    ]
 [1.   +0.j 1.   +0.j     ⋅         ⋅     4.899+0.j     ⋅    ]
 [    ⋅     1.414+0.j 2.   +0.j     ⋅         ⋅     7.746+0.j]
 [    ⋅         ⋅     1.732+0.j 3.   +0.j     ⋅         ⋅    ]
 [    ⋅         ⋅         ⋅     2.   +0.j 4.   +0.j     ⋅    ]
 [    ⋅         ⋅         ⋅         ⋅     2.236+0.j 5.   +0.j]]
```

The way `__repr__` is written is quite robust to any chosen `jnp.set_printoptions` parameter, e.g. `precision`, but it introduces a subtle bug which we have to fix. The symbol `∙` is supposed to mean "not stored", not "0". With the proposed `__repr__` implementation, when a stored diagonal contains a `0` it is replaced by mistake with a `⋅`, e.g.
```python
>>> dq.number(4)
SparseDIAQArray: shape=(4, 4), dims=(4,), dtype=complex64, ndiags=1
[[  ⋅      ⋅      ⋅      ⋅   ]
 [  ⋅    1.+0.j   ⋅      ⋅   ]
 [  ⋅      ⋅    2.+0.j   ⋅   ]
 [  ⋅      ⋅      ⋅    3.+0.j]]
```
(the first item should be 0.+0.j)

Happy for any better `__repr__` implementation suggestions!

### DIFF
--- a/dynamiqs/qarrays/sparse_dia_qarray.py
+++ b/dynamiqs/qarrays/sparse_dia_qarray.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+import re
 import warnings
 from collections import defaultdict
 from typing import get_args
@@ -78,6 +79,10 @@ class SparseDIAQArray(QArray):
             )
 
         return SparseDIAQArray(self.dims, out_offsets, out_diags)
+
+    @property
+    def ndiags(self) -> int:
+        return len(self.offsets)
 
     def conj(self) -> QArray:
         return SparseDIAQArray(self.dims, self.offsets, self.diags.conj())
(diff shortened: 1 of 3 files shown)


gautierronan (approved):
Looks good. I don't see any other solution to the 0 value not printed other than re-implementing our own `__repr__` from start to end with a for loop over diagonals. But this is somewhat annoying to do while staying compatible with `np.set_printoptions`. We would probably need to use `np.array2string` appropriately for each diagonal, and then re-represent them over each line. I'm fine with keeping your solution for now.

See the implementation of jax: 
https://github.com/google/jax/blob/7f680aaab8beab48f21056a172d1572a7067a63b/jax/_src/array.py#L375-L395
and the implementation of numpy:
https://github.com/numpy/numpy/blob/1d49c7f7ff527c696fc26ab2278ad51632a66660/numpy/_core/arrayprint.py#L540-L565

pierreguilmin:
Yes, let me open a ticket.

PR #680, opened 2024-08-15
Add sparse DIA constructor

pierreguilmin:
Constructor for sparse dia that takes care of the diagonals padding.

E.g. for `dq.create`:
```diff
- diag = jnp.sqrt(jnp.arange(1, dim + 1, dtype=cdtype())).at[-1].set(0.0)
- SparseDIAQArray(diags=diag[None, :], offsets=(-1,), dims=(dim,))
+ diag = jnp.sqrt(jnp.arange(1, dim, dtype=cdtype()))
+ sparsedia({-1: diag})
```
or `dq.sigmax`:
```diff
- diags = jnp.array([[1.0, 0.0], [0.0, 1.0]], dtype=cdtype())
- return SparseDIAQArray(diags=diags, offsets=(-1, 1), dims=(2,))
+ sparsedia({-1: [1.0], 1: [1.0]}, dtype=cdtype())
```

Apart from simplifying the code, it also decouples the internals of `SparseDIAQArray` from the specification API (which never changes, whatever the underlying implementation).

### DIFF
--- a/dynamiqs/qarrays/types.py
+++ b/dynamiqs/qarrays/types.py
@@ -4,12 +4,19 @@ from typing import Any, Union, get_args
 
 import jax.numpy as jnp
 from jax import Array
-from jaxtyping import ArrayLike
+from jaxtyping import ArrayLike, DTypeLike
 from qutip import Qobj
 
 from .qarray import QArray
 
-__all__ = ['QArray', 'QArrayLike', 'asqarray', 'asjaxarray', 'isqarraylike']
+__all__ = [
+    'QArray',
+    'QArrayLike',
+    'asqarray',
+    'asjaxarray',
+    'isqarraylike',
+    'sparsedia',
+]
 
 # In this file we define an extended array type named `QArrayLike`. Most
 # functions in the library take a `QArrayLike` as argument and return a `QArray`.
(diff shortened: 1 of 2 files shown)


gautierronan (changes requested):
The implementation looks good. I'm simply bothered by the API naming for these constructors. We now have:
- `asqarray` which constructs a `DenseQArray` from the input of an array
- `sparsedia` which constructs a `SparseDIAQArray` from the input of the dictionnary of diagonals

This is a very symmetric situation which is not reflected by the function naming.

The solution I would propose is to have the following functions:
- `asqarray(x, dims, layout)` which accepts an additional `layout` argument as input with the same behavior as utility functions,
- `dense_qarray(x, dims, dtype)` which is a constructor for the dense layout, used internally instead of the current `asqarray`,
- `sparsedia_qarray(offsets_diags, dims, dtype)` as introduced by this PR.

Note that `dense_qarray` and `sparsedia_qarray` should then probably live inside their respective class files, which will eliminate the circular import issue. This is also much more symmetric with our `TimeArray` implementation. WDYT?

pierreguilmin:
Yes, agreed 💯 Should we merge this one and do the renamings in another PR?

pierreguilmin:
I don't think it will solve the circular import unfortunately, but if you find a nice way to get rid of it I'm all ears!

gautierronan:
> Yes, agreed 💯 Should we merge this one and do the renamings in another PR?

Good with me!

> I don't think it will solve the circular import unfortunately, but if you find a nice way to get rid of it I'm all ears!

Not sure yet, but wanted to look at this at some point because we have quite a few circular import fixes in the code now.

gautierronan (approved):

PR #681, opened 2024-08-27
Add `floquet`

dkweiss31:
Closes DYN-231

Add Floquet solvers and associated tests. The API as it stands now supplies two functions: `floquet` and `floquet_t` for obtaining the floquet modes (and associated quasienergies) at t=0, and obtaining the floquet modes at other times t, respectively. I'm not thrilled about the name `floquet_t`, nor the name of e.g. the respective integrator `FloquetIntegrator_t`, so any suggestions for better names are welcome.


gautierronan:
Quick comment raised by @BenjaminDAnjou: `jnp.linalg.eig` is only supported on the CPU backend for now (see https://github.com/google/jax/issues/1259). This jax issue is live since 2019, so I don't think we can expect progress on this front.

Is there any other way to compute the Floquet basis @dkweiss31? Otherwise, maybe we could perform the workaround suggested in the linked issue, that is callback to the CPU implementation of `jnp.linalg.eig` using [external callbacks](https://jax.readthedocs.io/en/latest/notebooks/external_callbacks.html). Note that this is also what PyTorch does for `eig` on GPU.

dkweiss31:
Thanks for this, I hadn't appreciated that `eig` is only supported on CPUs. I'm not aware of another robust numerical technique for computing the floquet modes. Happy to try out the callbacks workaround

BenjaminDAnjou:
For information, the people in my group who know more than me about these implementations seem to think that implementing the callback workaround in the most sensible thing to do.

dynamiqs/utils/quantum_utils/general.py
```diff
@@ -975,6 +975,25 @@ def entropy_vn(x: ArrayLike) -> Array:
...
+        # Force this computation to be performed on the cpu by jit-ing and
+        # explicitly specifying the device.
+        with jax.default_device(jax.devices('cpu')[0]):
+            return jax.jit(jnp.linalg.eig)(_x)
+
+    return jax.pure_callback(
+        _eig_cpu,
+        (jnp.ones(x.shape[:-1], dtype=cdtype()), jnp.ones(x.shape, dtype=cdtype())),
+        x.astype(cdtype()),
+        vectorized=True,
```
gautierronan, line 993: This seems deprecated according to documentation of `jax.pure_callback`.

dynamiqs/utils/quantum_utils/general.py
```diff
@@ -975,6 +975,25 @@ def entropy_vn(x: ArrayLike) -> Array:
...
+    def _eig_cpu(_x: Array) -> tuple[Array, Array]:
+        # Force this computation to be performed on the cpu by jit-ing and
+        # explicitly specifying the device.
+        with jax.default_device(jax.devices('cpu')[0]):
+            return jax.jit(jnp.linalg.eig)(_x)
+
+    return jax.pure_callback(
+        _eig_cpu,
+        (jnp.ones(x.shape[:-1], dtype=cdtype()), jnp.ones(x.shape, dtype=cdtype())),
+        x.astype(cdtype()),
```
gautierronan, line 992: Why doesn't the following work?
```suggestion
        x,
```

dynamiqs/result.py
```diff
@@ -122,6 +132,42 @@ def _str_parts(self) -> dict[str, str | None]:
         return d | {'Propagators': array_str(self.propagators)}
 
 
+class FloquetResult(Result):
+    """Result of the Floquet integration.
+
+    Attributes:
+        modes _(array of shape (..., ntsave, n, n, 1))_: Saved Floquet modes
+        quasienergies _(array of shape (..., n))_: Saved quasienergies
+        T _(array of shape (...))_: Drive period
```
gautierronan, line 141: Missed this one.
```suggestion
        T _(scalar)_: Drive period
```

dynamiqs/result.py
```diff
@@ -122,6 +132,42 @@ def _str_parts(self) -> dict[str, str | None]:
...
+        solver _(Solver)_: Solver used.
+        gradient _(Gradient)_: Gradient used.
+        options _(Options)_: Options used.
+
+    Note-: Result of running multiple simulations concurrently
+        The resulting Floquet modes and quasienergies are batched according to the
+        leading dimensions of the Hamiltonian `H`. The shape of the drive period T as
+        well as the leading dimensions of `tsave` supplied to
+        [`dq.floquet`][dynamiqs.floquet] must be broadcastable to the shape of `H`. The
+        `cartesian_batching` flag is not relevant and does not affect the results.
```
gautierronan, line 153: Also forgot to update this since `T` and `tsave` are not batched anymore.

gautierronan (changes requested):
Thanks a lot @dkweiss31 for the PR, and sorry again for the late review.

In the most recent commits, I propose several changes to your PR:

- Unified the implementation of `floquet_t0` and `floquet_t` in a single integrator. This is mainly such that we don't need to recompute the propagator multiple times. Here, it is computed only once both for the t=t0 modes and for the later propagation of the t=t modes.
- Removed batching over `T`. This is kinda annoying, but related to not having support for batching over `tsave`. I think it really simplifies the implementation. Again, I'm quite down to rethink this in a later PR. In the meantime, I would propose to simply provide an example in the main docstring to show how to batch over multiple drive periods using an external `jax.vmap`. This is rather straightforward to achieve actually.
- Enforced that `tsave[0]` and `tsave[-1]` live within one period of each other. The other solution would indeed be to use a `jnp.mod(tsave, T)` to simplify the life of users, but for now, it simplifies our life haha.
- Proposed to rename `quasiens` to `quasienergies` and `floquet_modes` to `modes` in an effort to have short but meaningful names. Happy to reconsider this one if you think it was better previously.
- Removed the `safe` argument in favor of equinox runtime error checking, which is quite neat.

Let me know if all these changes work with you!

To finish the PR, I think there's mainly two things remaining:

- [ ] Adding the docstring example for batching over `T`/`tsave` (drive period)
- [x] Fixing tests with this new version

dkweiss31:
Hey @gautierronan , thank you very much for this review and for the significant simplifications you introduced. I am very happy with all of your proposed changes and will write up an example for T batching shortly.

I had been working with jax 0.4.29, and upgraded based on your comment that the `vectorized` option in `pure_callback` is deprecated. On updating how we call `pure_callback` I ran into issues with my tests hanging, which is apparently a known issue for jax > 0.4.31 https://github.com/jax-ml/jax/issues/24255. It seems however that there is an implementation of eig for GPU in the works https://github.com/jax-ml/jax/pull/24663 that should hopefully be merged soon. I propose then to just use `jax.numpy.linalg.eig` for now, and then make updates once the new GPU implementation is available.

gautierronan (approved):
Thanks a lot @dkweiss31! All good on my end! Let's wait for the review of @pierreguilmin and aim for merging next week :)

dkweiss31:
Sounds great, thanks very much @gautierronan !

pierreguilmin (dismissed):
Beautiful work, congratulations @dkweiss31, that's an amazing addition to the library! 😍 And @gautierronan I loved how your thorough review helped it converge into this clean and simple API. I'm adding a few nits on top. Happy to merge whenever.

pierreguilmin (approved):

PR #682, opened 2024-08-30
Refactor how `final_state` and `final_propagator` are saved

dkweiss31:
I've experienced issues trying to use functions like `jax.lax.cond` where in the functions passed to `cond` I reference a `result.final_state`, which goes through a conditional check on `options.save_states`. It is nicer to save the `final_state` directly after the integration and not have the ugly conditional logic that we have now.

### DIFF
--- a/dynamiqs/integrators/apis/mepropagator.py
+++ b/dynamiqs/integrators/apis/mepropagator.py
@@ -111,7 +111,7 @@ def _vectorized_mepropagator(
     # `n_batch` is a pytree. Each leaf of this pytree gives the number of times
     # this leaf should be vmapped on.
 
-    # the result is vectorized over `_saved` `final_state` and `infos`
+    # the result is vectorized over `_saved` `infos` and `final_state`
     out_axes = MEPropagatorResult(False, False, False, False, 0, 0, 0)
 
     if not options.cartesian_batching:
--- a/dynamiqs/integrators/apis/mepropagator.py
+++ b/dynamiqs/integrators/apis/mepropagator.py
@@ -111,8 +111,8 @@ def _vectorized_mepropagator(
     # `n_batch` is a pytree. Each leaf of this pytree gives the number of times
     # this leaf should be vmapped on.
 
-    # the result is vectorized over `_saved` and `infos`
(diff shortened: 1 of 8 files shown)


pierreguilmin:
I'm not very happy with this solution because we duplicate the information. 😕

dkweiss31:
Right, that is not ideal. We could have it so that `result.states` is not `None` only if `save_states==True`, but the final state is always saved as indicated here. This would also avoid the somewhat hacky `collect_saved` logic that we have now:
```python
def collect_saved(self, saved: Saved, ylast: Array) -> Saved:
    # if save_states is False save only last state
    if not self.options.save_states:
        saved = eqx.tree_at(
            lambda x: x.ysave, saved, ylast, is_leaf=lambda x: x is None
        )
    return saved
``` 
An annoying problem I continuously run up against is that the dimensions of `result.states` depends on `options.save_states`. This would help address that problem.

pierreguilmin:
What about the following plan:

1. always save `states` with shape:
   - `(..., ntsave, n, 1)` if `save_states=True` 
   - `(..., 1, n, 1)` if `save_states=False`
2. changing `final_states` to:
   ```python
   @property
   def final_state(self) -> Array:
       return self.states[..., -1, :, :]
   ```
3. Fix the doc to smthg along:
   ```diff
   - states (array of shape (..., ntsave, n, 1)) – Saved states.
   + states (array of shape (..., nsave, n, 1)) – Saved states, with `nsave = len(tsave)` if `options.save_states` is set to `True` and `ntsave = 1` otherwise.
   ```


It doesn't duplicate the information and fixes all following issues you mentionned:
- issue 1
    > I've experienced issues trying to use functions like `jax.lax`.cond where in the functions passed to cond I reference a `result.final_state`, which goes through a conditional check on `options.save_states`.
- issue 2
   > This would also avoid the somewhat hacky `collect_saved` logic that we have now
- issue 3
   > Moreover, this would additionally address the issue that the docstring of the various `Result` subclasses are not correct when `save_states=False`

pierreguilmin:
cc @gautierronan

dkweiss31:
@pierreguilmin im happy with that solution! Will implement

PR #683, opened 2024-08-31
Batch laguerre series computation in wigner

gautierronan:
Not actually faster for now. Not sure why? If you have an idea to speed it up, I can try, otherwise we can close the PR.

On main: 30.9 ms ± 517 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
On this branch: 86.6 ms ± 25.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Benchmark code:
```python
import dynamiqs as dq
import timeit

psi = dq.unit(dq.coherent(30, 2.0) + dq.coherent(30, -2.0))
dq.plot.wigner(psi)

%timeit dq.wigner(psi)[0].block_until_ready()
```

### DIFF
--- a/dynamiqs/utils/quantum_utils/wigner.py
+++ b/dynamiqs/utils/quantum_utils/wigner.py
@@ -6,6 +6,7 @@ from functools import partial
 import jax
 import jax.numpy as jnp
 from jax import Array, lax
+from jax.scipy.special import factorial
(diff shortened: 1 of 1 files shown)


gautierronan:
Closing because it doesn't work. Tracked in DYN-243.

PR #632, opened 2024-07-18
Add  `__pow__` and `powm()` features

bsaoptima:
This PR adds the `__pow__` magic method and the ̀powm()` method with associated tests. We still need to figure out the circular import problems to make everything work but the code runs smoothly locally.

PR #684, opened 2024-09-01
Add pow, powm and tests

gautierronan:
Replaces https://github.com/dynamiqs/dynamiqs/pull/632.

For `powm`, we cannot use a `jax.lax.scan` since the input and output pytree structures of the scan function change from one power to the next (i.e. different offsets).


pierreguilmin (approved):
The JIT might scale badly for large `n`s.

PR #687, opened 2024-09-02
Introduce global settings file

gautierronan:
This PR makes a few different changes to the `jax_utils.py` file:

1. Renames it into `global_settings.py`
2. Moves `set_layout` into this file instead of `qarray/utils.py`
3. Changes the API of `set_layout` from `set_layout(layout: Layout)` to `set_layout(layout: Literal['dense', 'dia'])` to mimic other functions in the file
4. Moves the `to_qutip` utility to a method of the QArray class only, similar to the `to_jax` method (mostly because it was a standalone function for now).

Can review commit by commit.


pierreguilmin:
Great, here are a few minor commits on top.

The only thing I disagree with is
> Moves the `to_qutip` utility to a method of the QArray class only, similar to the to_jax method (mostly because it was a standalone function for now).

This utility is very useful for people working with simple jax Arrays too, I think we should add it back.

gautierronan:
> This utility is very useful for people working with simple jax Arrays too, I think we should add it back.

No problem adding it back, but it should not be within the "Global Settings" section (and I quite like having this). We could put it within "QArray utilities" together with `dq.stack` ?

gautierronan:
Should be good @pierreguilmin

pierreguilmin (approved):
Molto bene

PR #690, opened 2024-09-02
Fix ruff

gautierronan:


pierreguilmin:
I had two other fixes from ruff.

pierreguilmin (approved):

PR #692, opened 2024-09-04
Fix sum and squeeze to return `Array` or `QArray` depending on axis

pierreguilmin:


dynamiqs/qarrays/qarray.py
```diff
@@ -432,3 +432,9 @@ def _check_compatible_dims(dims1: tuple[int, ...], dims2: tuple[int, ...]):
         raise ValueError(
             f'QArrays have incompatible dimensions. Got {dims1} and {dims2}.'
         )
+
+
+def _in_last_two_dims(axis: int | tuple[int, ...] | None, ndim: int) -> bool:
+    axis = (axis,) if isinstance(axis, int) else axis
+    last_two_dims = (-1, -2, ndim - 1, ndim - 2)
+    return axis is None or any(a in last_two_dims for a in axis)
```
gautierronan, line 440:
```suggestion
    return axis is None or any(a % ndim in [ndim - 1, ndim - 2] for a in axis)
```

gautierronan (approved):
A small suggestion below for readability, otherwise I'm good to merge

PR #694, opened 2024-09-05
Fix CI bugs related to recent releases of mkdocstrings-python

gautierronan:
Left to do:

- [ ]  Fix table of contents for `/python_api/index.md`
- [ ]  Reduce `/docstrings/*.jinja` files to minimal blocks


gautierronan:
@pierreguilmin Good to review on my end. To review, you can look commit by commit, or all commits but the last (which is just formatting all jinja files to a 2-space tab with a Jinja formatter).

pierreguilmin (approved):
It's a bit hard to review haha, but the doc looks good and we can always fix things later. 😊 Thanks a lot for the plumbering work, it's not the most fun but it's super valuable in the long term 💪

PR #704, opened 2024-09-11
Add benchmarks

gautierronan:
This PR adds scripts that can be used to benchmark a selected set of problems using `dynamiqs`, `QuTiP`, `QuTiP-JAX` or `QuantumOptics.jl`.

There's a lot of copy-pasted code, but it's actually quite difficult to find a unique structure that fits all libraries.

I'm not sure this is the final version of the public benchmarks we want to have, but I wanted to have this public for conferences where I show benchmarks so anyone can eventually check the validity of these benchmarks for themselves.


gautierronan:
Closing for now, as the library evolved quite a lot since these benchmarks. We may reopen at some point.

PR #705, opened 2024-09-13
Fix Python API table by adding systematic members

gautierronan:
This is a bit cumbersome, but I think this is the most reliable solution.

In short, the last documentation PR added the ability to look for sub-files to search for members. However, this also adds imported functions to the table, so we filter them using the `members` option.


pierreguilmin (approved):

pierreguilmin:
Issue arised because of #694.

pierreguilmin (approved):

pierreguilmin:
Many thanks to @HaroldErbin for bringing this to our attention. 🙌

PR #707, opened 2024-09-14
Setup doc versionning

pierreguilmin:
This PRs introduces a doc version, like e.g. NumPy does > https://numpy.org/doc/stable/index.html.

I think we can stay with minor versions granurality (`0.2`, `0.3`) to avoid having 1000 entries. And it would be nice to always have the `main` branch under the `dev` tag.

Screenshot 2024-09-14 at 11 22 54
Screenshot 2024-09-14 at 11 23 57
Screenshot 2024-09-14 at 11 24 17

@gautierronan can you:
- [x] help me fix the homepage that gets broken for some reason
- [x] fix the version selector design (if you want)
- [x] fix the banner design when looking at the doc for a version which is not the latest

The documentation on mike is a bit scarce:
- https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/
- https://github.com/jimporter/mike

To actually create new versions locally and test the code, you need to fetch the `gh-pages` branch and:
```shell
mike delete --all
mike deploy --update-aliases 0.2 stable
mike deploy dev
mike set-default stable
mike serve
```

⚠️ Please never use the `--push` option, or manually push the `gh-pages`, I don't think I've added protection rules on this branch so you would edit the actual website.

So the idea is that
- if we want to update the doc for a past release, we do smthg like
  ```shell
  mike deploy 0.2
  ```
- if we want to make a new doc release, we do smthg like
  ```shell
  mike deploy --update-aliases 0.3 stable
  ```
- to have the `main` branch always up to date on the `dev` tag, we'll add to the CI smthg like
  ```shell
  mike deploy dev
  ```

Remaining todo:
- [x] fix Python API page broken links (@gautierronan any idea?)
- [x] fix the banner appearing on stable version (but not on 0.2, although they are supposed to be the same)
- [x] test the documentation
- [x] add to the CI
- [x] manually push `0.2` as the latest version


pierreguilmin:
Ha, also note that it edits all documentation urls as such `http://127.0.0.1:8000/v0.1/python_api/index.html`, and we can use `http://127.0.0.1:8000/latest/documentation/index.html` for the latest version.

pierreguilmin:
Thanks a lot for the help @gautierronan ! 🙏

Nit: the version selector sometimes encroaches on the search bar. Maybe we should set a fixed length for the selector box, rather than relative to the string inside?
Screenshot 2024-09-26 at 15 35 43

Apart from that, the only remaining thing I'd need help with is:
> - [ ]  fix the banner design when looking at the doc for a version which is not the latest

docs/overrides/main.html
```diff
@@ -26,3 +26,10 @@
 <link rel="icon" class="js-site-favicon" type="image/svg+xml" href="{{ config.theme.favicon | url}}">
 <meta name="generator" content="mkdocs-{{ mkdocs_version }}, mkdocs-material-9.1.21">
 {% endblock %}
+
+{% block outdated %}
+  You're not viewing the latest version.
+  <a href="{{ '../' ~ base_url }}">
+    Click here to go to latest.
```
pierreguilmin, line 33: Nit: @gautierronan do you prefer "latest" or "stable"? I have a slight preference for "latest" as I find it more descriptive (it's what JAX is using), but for example NumPy and PyTorch use "stable". 🤷‍♂️

gautierronan:
> Apart from that, the only remaining thing I'd need help with is:

Yes I had not finished working on that, but didn't get more time yet.

> Nit: the version selector sometimes encroaches on the search bar. Maybe we should set a fixed length for the selector box, rather than relative to the string inside?

Good catch, will fix.

docs/overrides/main.html
gautierronan: Yes I changed to "stable" because that was used by Numpy and Pytorch, and I also prefer it a bit on my end. I prefer "latest" a bit less, but wouldn't mind it either.

gautierronan: Also, while changing to "stable", I found that the banner now appears when using the url `.../stable/...` but not when using `.../0.2/...` so not sure why. That was not the case before (it only appeared on `.../dev/...`, as expected).

pierreguilmin: Ok, let's go for stable :) We can maybe edit this banner text then.

> Also, while changing to "stable", I found that the banner now appears when using the url .../stable/... but not when using .../0.2/... so not sure why. That was not the case before (it only appeared on .../dev/..., as expected).

Weird, did you do the `mike set-default stable`?

gautierronan: > Weird, did you do the mike set-default stable?
Yes I did. Are you seeing the same problem locally as well?

gautierronan:
All good on my end. I leave the rest to you @pierreguilmin!

PR #713, opened 2024-09-18
Fix QuTiP dependencies on dev-qarray

gautierronan:
Suggest reviewing commit by commit. For now, no conversion from qutip DIA to dynamiqs DIA, but it's rather easy to do if we check `isinstance(some_qobj._data, qutip.core.data.Dia)` and then run `some_qobj._data.to_scipy()` and extract the diagonals and offsets from this. I was simply not sure whether we wanted to do that or not.

Related to DYN-191


dynamiqs/integrators/_utils.py
```diff
@@ -21,7 +21,7 @@
 from .core.abstract_integrator import AbstractIntegrator
 
 
-def _astimearray(x: QArrayLike | TimeArray) -> TimeArray:
+def astimearray(x: QArrayLike | TimeArray) -> TimeArray:
```
pierreguilmin, line 24: We should really do these kind of renamings in the `main` branch, we're going to live with both `dev-qarray` and `main` for at least a month, so we want them to stay as close as can be. 😊

dynamiqs/qarrays/qarray.py
```diff
@@ -45,7 +45,7 @@ class QArray(eqx.Module):
     def __check_init__(self):
         # === ensure dims is a tuple of ints
         if not isinstance(self.dims, tuple) or not all(
-            isinstance(d, int) for d in self.dims
+            isinstance(d, int) or np.issubdtype(d.dtype, int) for d in self.dims
```
pierreguilmin, line 48: We should only support `tuple[int, ...]` (otherwise we have to change the typing everywhere). IMO the calling method should convert `np.int` object to proper `int`.

pierreguilmin (changes requested):

gautierronan:
Replaced by https://github.com/dynamiqs/dynamiqs/pull/733.

PR #714, opened 2024-09-18
Support double precision wigner computation on GPU

### LINKED ISSUE #712: Wigner computation on GPU for double precision
I am simulating some dynamics in double precision, and it is useful to make movies of the Wigner function over time. This requires many Wigner function computations, which currently happen on the CPU ("Wigner computation is not yet supported on GPU for double precision (float64). The `state` array will be copied to the CPU to compute the Wigner distribution. Performance penalty is expected.") 

This is not a dealbreaker, but it would be excellent if Wigner computations could also be run on the GPU.

pierreguilmin:
Closes DYN-172.

Closes #712.


pierreguilmin:
￼"@abocquet I can't reproduce the bug we encountered, it seems to be working fine now. 🤷‍♂️

abocquet (approved):
🤷‍♂️

gautierronan (approved):

PR #715, opened 2024-09-18
Update qutip dependency to support `>= 5.0`

pierreguilmin:
This is mostly to fix `uv` install. Dynamiqs is not yet fully compatible with the newest QuTiP version.


pierreguilmin:
See work in progress in https://linear.app/dynamiqs/issue/DYN-191/fix-compatibility-with-qutip-50.

pierreguilmin:
Arf but obv tests are failing, let me think.

PR #716, opened 2024-09-19
Partial support for QuTiP >= 5.0

pierreguilmin:
Before this PR, we enforced the dependency on QuTiP < 5.0 because `qt.QObj` in versions > 5.0 does not have a native conversion to NumPy or JAX arrays, unlike in versions < 5.0. This limitation prevents `qt.QObj` objects from being used directly in all Dynamiqs functions. This PR lifts the dependency constraint because it breaks the installation with `uv`, although the original problem remains unresolved.

For native conversion support in QuTiP > 5.0, see https://github.com/qutip/qutip/pull/2533.

Related to https://linear.app/dynamiqs/issue/DYN-191/fix-compatibility-with-qutip-50.

### DIFF
--- a/dynamiqs/utils/jax_utils.py
+++ b/dynamiqs/utils/jax_utils.py
@@ -35,7 +35,7 @@ def to_qutip(x: ArrayLike, dims: tuple[int, ...] | None = None) -> Qobj | list[Q
                [1.+0.j],
                [0.+0.j]], dtype=complex64)
         >>> dq.to_qutip(psi)
-        Quantum object: dims = [[3], [1]], shape = (3, 1), type = ket
+        Quantum object: dims=[[3], [1]], shape=(3, 1), type='ket', dtype=Dense
         Qobj data =
         [[0.]
          [1.]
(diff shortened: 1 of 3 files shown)


pierreguilmin:
Thanks to @gaspardbb for noticing this.

gautierronan (changes requested):
You also need to call this `qobj_to_array` function inside `mesolve` and `sesolve`, when calling `jnp.asarray`, otherwise it won't work I think.

pierreguilmin:
> You also need to call this `qobj_to_array` function inside mesolve and sesolve, when calling jnp.asarray, otherwise it won't work I think.

Yes we'd need to call it everywhere we use `jnp.asarray` (thus most functions of the library, not just mesolve and sesolve). I'm proposing to wait to see if https://github.com/qutip/qutip/pull/2533 gets merged, otherwise we can write our own fix for the whole library. In the meantime, this PR fixes the installation with `uv`.

gautierronan (approved):
Ok let's keep it like that for now

PR #722, opened 2024-09-21
Improve tutorials reading flow

gautierronan:
Several small improvements to tutorials, in an attempt to improve the reading workflow and have them less irregular ("saccadé"). For instance, moved some notes to the end of the section, removed line jumps, and rephrased a few things. This PR also removes the `.dag()` sharp bit which will no longer be the case on `dev-qarray` (hence why I'm basing on this branch and not on `main`). 

For the time-dependent operators tutorial, the sentence constructions are a bit less systematic, but I just think it reads better this way (easier to follow).


docs/documentation/basics/batching-simulations.md
```diff
@@ -114,6 +111,8 @@ The simulation runs for each set of Hamiltonians, jump operators and initial sta
 
     For example for `dq.sesolve()` with `H` of shape _(2, 3, n, n)_, `psi0` can be of shape: _(n, 1)_, _(3, n, 1)_, _(2, 1, n, 1)_, _(2, 3, n, 1)_, _(..., 2, 3, n, 1)_, etc. By playing with the arguments shape, you have complete freedom over the simulation you want to run.
 
+??? Note "What is broadcasting?"
+    JAX and NumPy broadcasting semantics are very powerful and allow you to write concise and efficient code. For more information, see the [NumPy documentation on broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html).
```
pierreguilmin, line 115: I think it was better before. For people who don't know about broadcasting, they could check what it means. Now, they might read to the end of the section without understanding it.

docs/documentation/basics/time-dependent-operators.md
```diff
@@ -101,12 +90,7 @@ In Dynamiqs, PWC operators are defined by three array-like objects:
 - `values`: the constant values $(c_0, \ldots, c_{N-1})$ for each time interval, of shape _(..., N)_,
 - `array`: the array defining the constant operator $O_0$, of shape _(n, n)_.
 
-To construct a PWC operator, pass these three arguments to the [`dq.pwc()`][dynamiqs.pwc] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. This object then returns an array with shape _(..., n, n)_ when called at any time $t$.
-
-!!! Note
-    The argument `times` must be sorted in ascending order, but does not need to be evenly spaced. When calling the resulting time-array object at time $t$, the returned array is the operator $c_k\ O_0$ corresponding to the interval $[t_k, t_{k+1}[$ in which the time $t$ falls. If $t$ does not belong to any time intervals, the returned array is null.
-
-Let's define a PWC operator:
+To construct a PWC operator, these three arguments must be passed to the [`dq.pwc()`][dynamiqs.pwc] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. When called at some time $t$, this object then returns an array with shape _(..., n, n)_. For example, let us define a PWC operator with two evenly-spaced pieces:
```
pierreguilmin, line 93: Suggestion:
```suggestion
To construct a PWC operator, these three arguments must be passed to the [`dq.pwc()`][dynamiqs.pwc] function, which returns a [`TimeArray`][dynamiqs.TimeArray] object. When called at some time $t$, this object then returns an array with shape _(..., n, n)_. For example, let us define a PWC operator $H(t)$ with constant value $H(t)=3\sigma_z$ for $t\in[0, 1[$ and $H(t)=-2\sigma_z$ for $t\in[1, 2[$ :
```

pierreguilmin (changes requested):
Great! Minor suggestions.

pierreguilmin:
Also, could you base these changes on main and create another PR targeting `dev-qarray` for the sharp bit removal? 🙇‍♂️ The reason is that if we want or need to change these documentation files on `main` before we have merged `dev-qarray`, it's going to create an inextricable conflict mess.

PR #723, opened 2024-09-21
Improve GIFs creation performance and fix the color quantization bug

pierreguilmin:
This PR revamps the way we create GIFs. The main idea is to write frames on an in-memory `BytesIO()` object (a file-like interface for working with bytes data in memory), instead of on disk.

<p align="center">
    
</p>

Main features:
- ~3x faster (e.g. cat inflation or Kerr oscillator example: 23s → 8s)
- fix GIF colors quantization
- avoid writing temporary directory and files when creating GIFs
- simplify the API of `dq.plot.gifit` and `dq.plot.wigner_gif`

Small things:
- replace the `imageio` dependency with the smaller `pillow`
- use `dq.plot.gifit` in the implementation of `dq.plot.wigner_gif`, instead of duplicating code
- `dq.plot.gifit` old keywords arguments `gif_duration` and `fps` should now be passed to the returned function
- `dpi` is now controlled as for all other figures using default Matplotlib parameters
- group doctest fixtures related to Matplotlib
- change default dpi for all documentation plots to `150`
- add a `rendergif` doctest fixture (similar to `renderfig`)
- add a `dpi` keyword argument to `dq.plot.mplstyle()`

Closes DYN-173. Related to DYN-230.

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -1,4 +1,5 @@
 from collections.abc import Sequence
+from pathlib import Path
 
 import jax.numpy as jnp
 import matplotlib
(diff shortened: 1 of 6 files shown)


gautierronan (approved):
This is amazing, such a great addition! 🚀

PR #724, opened 2024-09-22
Refactor integrators

pierreguilmin:
This PR tries to clean the mess of integrators, by
- adding docstrings to every solver
- splitting objects between abstract classes, interfaces and mixins
- removing a bug introduced by messing with Python MRO in `DiffraxIntegrator`

It doesn't change the logic anywhere. Now almost all classes only inherit from a single abstract class (+ some interface and/or mixins).

Essentially, any class attributes can now be defined using the typical `eqx.Module` attribute syntax, and we can trust Python MRO to distribute the attributes to the correct class when using multiple inheritance. The `__init__` should be called with `kwargs` for clarity and to avoid having to guess the correct arguments order.

Minor things:
- use `__check_init__` and `__post_init__` from equinox

Now writing your custom solver is a bit like doing your groceries: pick the base classes you need, replace the ones you want, and everything should work fine.

### DIFF
--- a/dynamiqs/integrators/apis/mepropagator.py
+++ b/dynamiqs/integrators/apis/mepropagator.py
@@ -21,6 +21,7 @@ from .._utils import (
     catch_xla_runtime_error,
     get_integrator_class,
 )
+from ..core.abstract_integrator import MEPropagatorIntegrator
 from ..mepropagator.expm_integrator import MEPropagatorExpmIntegrator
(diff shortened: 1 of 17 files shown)


gautierronan (approved):
It was a bit tedious to review so I might have missed minor things, but overall this looks great. This should indeed be much simpler to define new solvers now!

PR #725, opened 2024-09-23
Fix delayed imports by refactoring qarrays/ folder

gautierronan:
This PR fixes almost all delayed imports throughout the code by refactoring the structure of the whole qarrays/ folder.

The proposed structure is now (by order of enclosed imports, i.e. each file can import any of the files above itself):

- /layout.py defines the Layout enumerator (unchanged)
- /qarray.py defines `QArray`, `QArrayLike` and` isqarraylike`
- /dense_qarray.py defines `DenseQArray`, and all the methods directly associated to this class, including some conversion functions to/from `DenseQArray`
- /sparsedia_qarray.py similarly
- /type_conversion.py defines public functions to convert to/from `QArrays`
- /utils.py defines utility functions on `QArrays` (only `stack` for now)

With this structure, no file inside qarrays/ should ever need to import a file from the utils/ folder, thus avoiding circular imports. Two exceptions remain inside /dense_qarray.py, which are `ptrace` and `norm` for which the logic is inside /utils but is also used for `DenseQArray`. The solution would be to define `_ptrace` and `_norm` inside /dense_qarray.py and to import these private functions from utils/quantum_utils/general.py. Do we want to do that @pierreguilmin?

Finally, the various type conversion functions now follow the same naming conventions:

- `asdense`/`asjaxarray`/`asqobj`/... are public API methods to convert any `QArrayLike` to a given type. 
- `_dense_to_qobj`/`_sparsedia_to_dense`/... are private methods to convert a specific type to another specific type.

I am very open to renaming these if you have better suggestions. In particular, we have `asqobj(...)` and `x.to_qutip()` which are both user-facing and achieve the same thing with a different name. Similarly, we have `asjaxarray(...)` and `x.to_jax()`.

Left to do (probably in separate PRs for the documentation):

- [ ] Write documentation for new public type conversion methods
- [ ] Fix existing documentations (e.g. `to_qutip`)
- [x] (if relevant) Renaming of the various new methods
- [ ] (if approved) Introduction of `_ptrace` and `_norm` inside /dense_qarray.py

For review, I suggest going commit by commit. The only annoying commit to review will be the first one, which mixes existing code moved around with new code. Related to DYN-210.


dynamiqs/qarrays/dense_qarray.py
```diff
@@ -195,20 +202,25 @@ def __rmatmul__(self, y: QArrayLike) -> QArray:
...
             dims = self.dims
-            data = asjaxarray(y) @ self.data
+            data = _getjaxarray(y) @ self.data
         else:
             return NotImplemented
 
         return DenseQArray(dims, data)
 
-    def __and__(self, y: QArray) -> QArray:
+    def __and__(self, y: QArrayLike) -> QArray:
```
pierreguilmin, line 211: This wasn't planned, I don't think we should support `&` with other types than `QArray`. Can we revert? (or do it in another PR, then we also need to fix the base class type)

pierreguilmin (changes requested):
Beautiful 👏

> Two exceptions remain inside /dense_qarray.py, which are ptrace and norm for which the logic is inside /utils but is also used for DenseQArray. The solution would be to define _ptrace and _norm inside /dense_qarray.py and to import these private functions from utils/quantum_utils/general.py. Do we want to do that @pierreguilmin?

As a temporary solution yes, the proper way to do all this anyway is https://linear.app/dynamiqs/issue/DYN-224/conform-to-array-api, which we're slowly moving away for now. 🥲 But that's fine! Can we do it in another PR?

> asqobj(...) and x.to_qutip() which are both user-facing and achieve the same thing with a different name. Similarly, we have asjaxarray(...) and x.to_jax().

Ha yes, let's rename everything so they match!

Random suggestions on top of the comments. Note that I'm happy to merge this as is and we can implement the fixes in other PRs, if that's more convenient.
- Regarding 509498a, if you search for `"qarrays import"` you'll find files where the import has not been fixed.
- We should have a user facing `_sparsedia_constructor` in `type_conversion`.
- A few implementations (e.g. `dq.asqobj`, `_getjaxarray`, `asdense`, ...) are breaking the list support such as `[[1, 2, 3]]`, everytime we do smthg like:
  ```
    elif isinstance(x, list):
        return [... for sub_x in x]
  ```
  It's not trivial to fix though: we should only do the `[... for sub_x in x]` logic for `isinstance(x, sequence)` (because we also want to support tuples of qarrays), and if the list is more than depth 2. I suggest reverting to what we had (i.e. we don't support sequences of qarray yet, and treat lists as smthg that should be converted to a jax array).
- Nit: we sometimes accept `dims: int | tuple[int, ...] | None = None` and sometimes `dims: tuple[int, ...] | None = None`, let's only support the second one.
- Nit: we sometimes do `... for _x in x` and sometimes `... for sub_x in x`, let's only do the second one.
- Nit: the logic in `_dense_to_qobj` is repeated with `asqobj`, it's a bit unfortunate.
- Nit: could we move `_getjaxarray` to `qarray.py` and find a better name?
- Nit: the shape check for `asqobj` is gone.

dynamiqs/qarrays/dense_qarray.py
gautierronan: That was a weird one cause we support QArrayLike already in SparseDIAQArray. Should I revert both?

pierreguilmin: Ha I didn't noticed, I propose to revert `SparseDIAQArray` as well. I think we should not overload our newly introduced operators on existing array-like objects.

gautierronan:
@pierreguilmin Addressed your comments. You can review the new commits one by one.

The documentation will be in a separate PR. Also, I won't do the move of `_ptrace` and `_norm` here because it could be a bit tedious, and might require some broader thinking with the Array API conforming.

pierreguilmin (approved):
ezgif-2-2f9270b436

PR #726, opened 2024-09-23
Move ptrace tuple args to *args

abocquet:
Probably one of the most popular use for `ptrace` is to trace a single mode out, to plot it for instance.
This PR makes the API more straightforward by turning `state.ptrace((0,))` into `state.ptrace(0)`

### DIFF
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -54,7 +54,7 @@ class DenseQArray(QArray):
         data = jnp.broadcast_to(self.data, shape)
         return DenseQArray(self.dims, data)
 
-    def ptrace(self, *keep: tuple[int, ...]) -> QArray:
+    def ptrace(self, *keep: int) -> QArray:
         from ..utils.quantum_utils.general import ptrace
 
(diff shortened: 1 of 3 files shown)


dynamiqs/qarrays/dense_qarray.py
```diff
@@ -54,7 +54,7 @@ def broadcast_to(self, *shape: int) -> QArray:
         data = jnp.broadcast_to(self.data, shape)
         return DenseQArray(self.dims, data)
 
-    def ptrace(self, keep: tuple[int, ...]) -> QArray:
+    def ptrace(self, *keep: tuple[int, ...]) -> QArray:
```
pierreguilmin, line 57:
```suggestion
    def ptrace(self, *keep: int) -> QArray:
```

pierreguilmin (changes requested):
Yes 👍 Minor type fix.

dynamiqs/qarrays/dense_qarray.py
abocquet: Oh right sorry

pierreguilmin (approved):

PR #733, opened 2024-09-24
Fix qutip dep

gautierronan:
Replaces https://github.com/dynamiqs/dynamiqs/pull/713. On top of https://github.com/dynamiqs/dynamiqs/pull/725.

### DIFF
--- a/dynamiqs/qarrays/qarray.py
+++ b/dynamiqs/qarrays/qarray.py
@@ -36,6 +36,22 @@ def _asjaxarray(x: QArrayLike) -> Array:
         return jnp.asarray(x)
 
 
+def _dims_to_qutip(dims: tuple[int, ...], shape: tuple[int, ...]) -> list:
+    dims = list(dims)
+    if shape[-1] == 1:  # [[3], [1]] or [[3, 4], [1, 1]]
+        dims = [dims, [1] * len(dims)]
+    elif shape[-2] == 1:  # [[1], [3]] or [[1, 1], [3, 4]]
+        dims = [[1] * len(dims), dims]
+    elif shape[-1] == shape[-2]:  # [[3], [3]] or [[3, 4], [3, 4]]
+        dims = [dims, dims]
+    return dims
+
+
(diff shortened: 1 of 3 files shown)


pierreguilmin:
@gautierronan can you rebase this?

dynamiqs/qarrays/qarray.py
```diff
@@ -35,6 +35,21 @@ def _asjaxarray(x: QArrayLike) -> Array:
...
+        dims = [dims, [1] * len(dims)]
+    elif shape[-2] == 1:  # [[1], [3]] or [[1, 1], [3, 4]]
+        dims = [[1] * len(dims), dims]
+    elif shape[-1] == shape[-2]:  # [[3], [3]] or [[3, 4], [3, 4]]
+        dims = [dims, dims]
+    return dims
+
+
+def _dims_from_qutip(dims: list) -> tuple[int, ...]:
+    dims = np.max(dims, axis=0)
```
pierreguilmin, line 50: Cute. 🤗

pierreguilmin (changes requested):
I think we also need to fix `_asjaxarray`.

gautierronan:
> I think we also need to fix _asjaxarray.

Not sure what you mean by this, but if it's not explicitely related to the Qobj support, then I will address it in a subsequent PR. There are quite a few things that are problematic right now with these conversion functions.

pierreguilmin:
> > I think we also need to fix _asjaxarray.
> 
> Not sure what you mean by this, but if it's not explicitely related to the Qobj support, then I will address it in a subsequent PR. There are quite a few things that are problematic right now with these conversion functions.

Ha it's just that `dq.asjaxarray()` should support `Qobj`. 😉

pierreguilmin (dismissed):

gautierronan:
Fixed the support for Qobj in `dq.asjaxarray`. I will make a subsequent PR to improve all these conversion functions, which is a bit clunky right now.

pierreguilmin (approved):

PR #736, opened 2024-09-26
Minor nits on top of qarray types refactoring

pierreguilmin:
Follow-up to #725.

@gautierronan we also need to fix (or add todos):
-  [ ] `_dims_to_qutip` should support a `dims=None` (potentially called in `asqobj`)
- [ ] `_asjaxarray` is exactly the same as `asjaxarray`, should we juste rename the first?
- [ ] Methods accepting `QArrayLike` don't actually support `QArrayLike` (`asdense`, `assparsedia`, `asqobj`). Basically anywhere we take in a `QArrayLike`, we should write the method to support, for example:
   ```python
    x = (np.array([[[1, 2, 3]], [4, 5, 6]]]), dq.fock(3, (0, 1)))
    ```
- [ ] `_warn_qarray_dims` doesn't work on `qobj`, doesn't work on list of qarray.

### DIFF
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -21,6 +21,7 @@ _bkron = jnp.vectorize(jnp.kron, signature='(a,b),(c,d)->(ac,bd)')
 
 def _dense_to_qobj(x: DenseQArray) -> Qobj | list[Qobj]:
     if x.ndim > 2:
+        # TODO: generalize to any nested sequence with the appropriate shape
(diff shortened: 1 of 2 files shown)


gautierronan (approved):
Let's fix your to-do list in an other PR. PR https://github.com/dynamiqs/dynamiqs/pull/725 is indeed incomplete and requires some more work.

I'll wait for this PR and for https://github.com/dynamiqs/dynamiqs/pull/733 to be merged, and will propose a new PR to finish up this conversion function refactoring.

PR #738, opened 2024-09-26
Add Rx, Ry, Rz, T and Toffoli gates

vaicecaze:
Implemented Rx, Ry, Rz, T and Toffoli gates


pierreguilmin (commented):
Great! I've made several minor edits on top, mostly cosmetics. I recommend reviewing each new commit individually so you know what to address in future pull requests. Don't hesitate to ask if you have any questions about the proposed changes.

Also, you can see that the CI (continuous integration) failed on this PR on your latest commit. We run each example of the documentation to verify that the output is correct. You can run this specific part of the CI locally with `task doctest`, or for a single file with, for example, `pytest dynamiqs/utils/operators.py`. Note that the output may be different from what you see in a console/notebook, because we use the setting `jnp.set_printoptions(precision=3, suppress=True)`.

In another PR, we should vectorize `rx`, `ry` and `rz`! See e.g. what we do for the `displace` operator. So the new API should be `def rx(theta: ArrayLike) -> Array`, and for example
```python
>>> dq.rx([jnp.pi/3, jnp.pi/2, jnp.pi]).shape
(3, 2, 2)
```
Note that it should work for *any* input shape, e.g. if `theta` has shape `(3, 4, 5)`, then `rx(theta)` should have shape `(3, 4, 5, 2, 2)`. Check out `jnp.vectorize` to do this properly. 😉

Let's wait for @gautierronan's review before merging.

gautierronan (approved):
LGTM! At some point we might want to think about the naming of these functions (+`dq.hadamard`) as it feels a bit inconsistent. Maybe we could add a namespace `dq.gate.*`, or a section in the documentation? Just random ideas, for now it's good with me.

PR #742, opened 2024-10-03
Vectorize Rx, Ry and Rz gates

vaicecaze:
Closes DYN-251.


pierreguilmin (approved):
Thanks @vaicecaze! I think the implementation was a bit off, I fixed it in last commit.

pierreguilmin (approved):

PR #743, opened 2024-10-03
Adapt task doctest to different OS

vaicecaze:


gautierronan:
Hi @vaicecaze, thanks for fixing this. Instead of adding a file, I would suggest the following fix that only uses shell script:
```shell
cmd = '''
  echo "\n>>> pytest dynamiqs" &&
  case "$OSTYPE" in
    linux*|darwin*|solaris*) rm -f docs/figs_code/*.{png,gif} ;;
    msys*|cygwin*|win32*)    del /Q docs\\figs_code\\*.png docs\\figs_code\\*.gif ;;
    *)                       echo "Unknown OS: $OSTYPE" && exit 1 ;;
  esac &&
  pytest dynamiqs
'''
```
It seems to work on macOS, I'll let you check if that also works under Windows.

vaicecaze:
Hi @gautierronan,
Thanks for your reply!
'case' is not recognized on Windows so I adapted again the code to the OS. I removed end-of-line '\n' characters as they appear as invalid. Hope it also works on macOS.

pierreguilmin:
> It doesn't seem to be working on linux unfortunately. I'm not sure how to fix this in a clean way TBH.

I think the original solution was ok (with the Python script). We should strive to support all development platforms.

> Also, @pierreguilmin, why do we need this rm call?

Otherwise if the doctest fails the doc figures are not updated and come from the latest run of the doc (e.g. btw two branches), which is quite confusing.

gautierronan:
> I think the original solution was ok (with the Python script). We should strive to support all development platforms.

Alright, I'm fine either way, let's go back to the original solution then. Because the python script is to remove documentation files, maybe we can put this `remove_file.py` somewhere inside `docs/` to not clutter the main directory.

vaicecaze:
Hi @gautierronan!
I just reverted to commit 902d900a7ff7baf28beb6526059c10ada5a80507 with the remove_files.py method and moved it to docs/ folder. Tell me if it's okay for you.

pyproject.toml
```diff
@@ -135,11 +135,11 @@ cmd = 'echo "\n>>> pytest -n=auto tests" && pytest -n=auto tests'
 help = "run the unit tests suite (pytest)"
 
 [tool.taskipy.tasks.doctest-code]
-cmd = 'echo "\n>>> pytest dynamiqs" && rm -f docs/figs_code/*.{png,gif} && pytest dynamiqs'
+cmd = 'echo "\n>>> pytest dynamiqs" && python docs/remove_files.py && pytest dynamiqs'
 help = "check code docstrings examples (doctest)"
 
 [tool.taskipy.tasks.doctest-docs]
-cmd = 'echo "\n>>> pytest docs" && rm -f docs/figs_docs/*.{png,gif} && pytest docs'
+cmd = 'echo "\n>>> pytest docs" && rm python docs/remove_files.py && pytest docs'
```
gautierronan, line 142:
```suggestion
cmd = 'echo "\n>>> pytest docs" && python docs/remove_files.py && pytest docs'
```
(typo)

gautierronan (changes requested):
Apart from the small typo below, if the CI passes, then it's good for me!

docs/remove_files.py
```diff
@@ -0,0 +1,19 @@
+import sys
+from pathlib import Path
+
+from jax.typing import ArrayLike
+
+
+def remove_files(pattern: ArrayLike):
```
gautierronan, line 7:
```suggestion
def remove_files(pattern: str):
```
And this should be a string I guess?

gautierronan (dismissed):
Thanks @vaicecaze!

gautierronan (approved):

PR #744, opened 2024-10-04
Add thermal_dm() state method

vaicecaze:
Removed remove_files.py file in this PR because of compatibility issues with tests

Closes DYN-249.


pierreguilmin:
I think this stills needs a bit more work, am I right? If you need anything from our side @vaicecaze don't hesitate to ask. Otherwise there's no rush, it's just to ensure you're not stuck somewhere. 😊

pierreguilmin (changes requested):
I've added a commit with a few fixes on top. @vaicecaze:
- the return should be complex
- `dq.thermal_dm(4, [0.1, 0.2, 0.3])` raises an exception, it should return an array of shape `(3, 4, 4)`

gautierronan (dismissed):

PR #746, opened 2024-10-04
Add an example tutorial on the Kerr oscillator

gautierronan:
First iteration for an "advanced" example. I think this should be quite useful for people who want to get started in Dynamiqs but have already used QuTiP or similar.


pierreguilmin (changes requested):
High-level comments to start:
- Why studying a driven oscillator if we remove it later? I suggest removing the drive from the beginning.
- The GIF is going way too fast.
- Can we choose to plot a physical observable rather than $\braket{a}$ ?

On the nits part, I propose a few conventions across the documentation (following the basic examples):
- When we define parameters we use the order: Hilbert space size, system parameters, save time stuff.
- Renamings (`N` -> `n`, `sim_time` -> `tend`, `alpha` -> `alpha0`).
- Always use an odd number of time steps for the saving times (`100` -> `101`).
- Reorder code to define operators then initial state, then save time (mesolve arguments ordering).

gautierronan:
> Why studying a driven oscillator if we remove it later? I suggest removing the drive from the beginning.

In the end I added a rather big section that tackles the transmon regime. This way, we fully use this time-dependent drive, and provide several more applications of dynamiqs.

> Can we choose to plot a physical observable rather than ⟨ a ⟩ ?

We could do it with computing `a+adag` and `1j*(a - adag)` separately and recombining them, but I don't really see the point. It's called "expectation operators" and not "observable operators". They don't need to be hermitian for us to compute their traces.

pierreguilmin (approved):
Wow 🤯 I can't do a proper review, happy to merge! The only issue might be the runtime of this example? We can fix this in another PR to avoid slowing down the CI too much (either by reducing `n` if possible, but most probably by just excluding it from the CI and hard saving the figures).

gautierronan:
> The only issue might be the runtime of this example? We can fix this in another PR to avoid slowing down the CI too much (either by reducing `n` if possible, but most probably by just excluding the file from the CI and hard saving the figures).

Indeed the examples are somewhat slow, although I already tried to keep that to a minimum. I don't think reducing `n` will help since it's already very small. I could tweak parameters to reduce the number of epochs in the last example typically.

pierreguilmin:
> I could tweak parameters to reduce the number of epochs in the last example typically.

I don't think we should overly constrain the tutorials due to runtime. Of course, they should be kept to a maximum of a few minutes. But even then, if we have several tutorials that take around 1 minute to run each, it will quickly bloat the CI.

We can merge though, it will be better to take care of this change in another PR. 😊

PR #748, opened 2024-10-08
Add an `Event` solver for `dq.jssesolve()`

dkweiss31:
**EDIT by @pierreguilmin: we're splitting this work into multiple PRs to adapt all the original work by @dkweiss31 to the library. Subsequent PRs that use parts of this one should reference it in their description so we can maintain a clear record.**

Closes DYN-124

I only advise reading commit by commit if you'd like a good laugh at the various false starts. The key breakthrough in the implementation was in understanding how to use buffers in the context of while loops in order to update arrays we want to save during each loop iteration (see e.g. [here](https://github.com/patrick-kidger/diffrax/blob/main/diffrax/_integrate.py#L422-L462)).  There are still a few outstanding TODOs/issues:

1. Should we save stats for the jump trajectories (since they each are composed of multiple calls to diffeqsolve)
2. The `_run()` function in `mcsolve/diffrax_integrator.py` is quite similar to `run()` in `core/diffrax_integrator.py`. Should they be combined?
3. I occasionally see nans pop up when I ask for many (>100) trajectories. They occur in the intermediate states/expectation values and are thus probably due to the interpolation we do in the interior of the while loop body, though I'm not certain of this. 

I figure though that the implementation is close enough to final to open the PR.


gautierronan:
Hi @dkweiss31, it's awesome to see this now live as a PR! 

As a warning, like the `floquet` solvers, we might only properly review and merge this PR after the release of QArrays, which is our main focus for now (expected release early November). I hope this is ok with you. In case you would need it sooner than that, we can definitely speed up the review though!

dkweiss31:
Understood! Best of luck with the QArray release :)

basnijholt:
Really looking forward to this! Please let me know whenever this is in any usable state and I will try running it in my workflows 😄

gautierronan:
@dkweiss31 We've just merged `dev-qarray` into `main`, so I'm quite down to review your PR soon (within next two weeks). Could you update the PR and add support for QArrays? Hopefully that shouldn't be too difficult (let me know if you need help on that). Thanks!

dkweiss31:
@gautierronan definitely! Congrats on getting Qarray merged 🥳

dynamiqs/integrators/apis/mcsolve.py
```diff
@@ -0,0 +1,272 @@
...
+    tsave: ArrayLike,
+    *,
+    keys: ArrayLike = jax.random.split(jax.random.key(31), num=10),  # noqa: B008
+    exp_ops: list[QArrayLike] | None = None,
+    solver: Solver = Tsit5(),  # noqa: B008
+    root_finder: AbstractRootFinder | None = optx.Newton(1e-5, 1e-5, optx.rms_norm),  # noqa: B008
+    gradient: Gradient | None = None,
+    options: Options = Options(),  # noqa: B008
+) -> MCSolveResult:
+    r"""Perform Monte-Carlo evolution, unraveling the master equation.
```
gautierronan, line 52:
```suggestion
    r"""Solve the Lindblad master equation through Monte-Carlo sampling of its jump unraveling.
```
(maybe the formulation is not 100% exact and could be improved, the comment is mostly to use the "Solve ..." similar to other functions)

dynamiqs/integrators/apis/mcsolve.py
```diff
@@ -0,0 +1,272 @@
...
+
+    We follow the algorithm outlined in Abdelhafez et al. to efficiently perform
```
gautierronan, line 54: Can you add a link to the paper, and maybe the year of publication?

dynamiqs/integrators/apis/mcsolve.py
```diff
@@ -0,0 +1,272 @@
...
+    *,
+    keys: ArrayLike = jax.random.split(jax.random.key(31), num=10),  # noqa: B008
+    exp_ops: list[QArrayLike] | None = None,
+    solver: Solver = Tsit5(),  # noqa: B008
+    root_finder: AbstractRootFinder | None = optx.Newton(1e-5, 1e-5, optx.rms_norm),  # noqa: B008
+    gradient: Gradient | None = None,
+    options: Options = Options(),  # noqa: B008
+) -> MCSolveResult:
+    r"""Perform Monte-Carlo evolution, unraveling the master equation.
+
```
gautierronan, line 53: We should add a generic comment about the Lindblad master equation (similarly to `mesolve`) before detailing how the monte carlo solver actually works.

dynamiqs/integrators/apis/mcsolve.py
```diff
@@ -0,0 +1,272 @@
...
+        H _(qarray-like or time-array of shape (bH?, n, n))_: Hamiltonian.
+        Ls _(list of qarray-like or time-array, of shape (nL, n, n))_: List of
+            jump operators.
+        psi0 _(qarray-like of shape (bpsi?, n, 1))_: Initial state.
+        tsave _(array-like of shape (nt,))_: Times at which the states and expectation
+            values are saved. The equation is solved from `tsave[0]` to `tsave[-1]`, or
+            from `t0` to `tsave[-1]` if `t0` is specified in `options`.
+        keys _(Array of shape (ntraj,))_: Total number of jump trajectories to
+            simulate, not including the no-jump trajectory. Defaults to a list of keys
+            of length 10.
```
gautierronan, line 89: Should state explicitely that the `keys` length corresponds to the number of trajectories sampled, similarly to https://github.com/dynamiqs/dynamiqs/blob/ef8cc5fe3e4dba6987313a621e9559b268c988d0/dynamiqs/integrators/apis/dsmesolve.py#L134-L136.

dynamiqs/integrators/apis/mcsolve.py
```diff
@@ -0,0 +1,272 @@
...
+            operators for which the expectation value is computed.
+        solver: Solver for the integration. Defaults to
+            [`dq.solver.Tsit5()`](/python_api/solver/Tsit5.html).
+        root_finder: Root finder passed to dx.diffeqsolve() to find the exact time an
+            event occurs. Can be `None`, in which case the root finding functionality
+            is not utilized. It is recommended to pass a root finder (such as the
+            default Newton root finder) so that event times are not determined by the
+            integration step sizes in diffeqsolve. However there are cases where the
+            root finding can fail, causing the whole simulation to fail. Passing `None`
+            for `root_finder` can alleviate the issue in these cases.
```
gautierronan, line 100: We should also refer to the `diffrax` documentation with a link somewhere inside this description.

dynamiqs/integrators/apis/mcsolve.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+        shapes = [x.shape[:-2] for x in [H, *Ls, psi0]] + [keys.shape[:-1]]
+        bshape = jnp.broadcast_shapes(*shapes)
+        nvmap = len(bshape)
+        # broadcast all vectorized input to same shape
+        n = H.shape[-1]
+        H = H.broadcast_to(*bshape, n, n)
+        Ls = [L.broadcast_to(*bshape, n, n) for L in Ls]
+        psi0 = psi0.broadcast_to(*bshape, n, 1)
+        keys = jnp.broadcast_to(keys, (*bshape, *keys.shape[-1:]))
```
gautierronan, line 197:
```suggestion
        keys = jnp.broadcast_to(keys, (*bshape, keys.shape[-1]))
```
nit

dynamiqs/integrators/core/diffrax_integrator.py
```diff
@@ -225,3 +226,18 @@ def vector_field(t, y, _):  # noqa: ANN001, ANN202
...
+class MCDiffraxIntegrator(DiffraxIntegrator, MCInterface):
+    """Integrator solving the Schrödinger equation with Diffrax."""
+
+    # subclasses should implement: diffrax_solver, discontinuity_ts
+
+    @property
+    def terms(self) -> dx.AbstractTerm:
+        def vector_field(t, y, _):  # noqa: ANN001, ANN202
+            L, H = self.L(t), self.H(t)
+            Hnh = sum_qarrays(-1j * H, *[-0.5 * _L.dag() @ _L for _L in L])
```
gautierronan, line 240: Important comment: we should never call `_L.dag() @ _L` here, otherwise the complexity of the solver will shoot up to $O(n^3)$ in the dense case instead of the expected $O(n^2)$ for the Monte Carlo solver. This would really defeat its purpose.

dynamiqs/integrators/core/interfaces.py
```diff
@@ -16,14 +20,25 @@ class SEInterface(eqx.Module):
     H: TimeArray
 
 
-class MEInterface(eqx.Module):
+class _MInterface(eqx.Module):
+    def L(self, t: Scalar) -> list[QArray]:
+        return [_L(t) for _L in self.Ls]  # (nLs, n, n)
```
gautierronan, line 25: This calls `self.Ls` which is not defined in this context (only defined in child classes).

I think I would keep `MEInterface` as before, and inherit `MCInterface` from it. The MC solver is still solving the ME, so it's not unexpected IMO.

dynamiqs/qarrays/sparsedia_qarray.py
```diff
@@ -345,6 +345,8 @@ def _check_key_in_batch_dims(key: int | slice | tuple, ndim: int):
     valid_key = False
     if isinstance(key, (int, slice)):
         valid_key = ndim > 2
+    elif isinstance(key, Array):
+        valid_key = len(key.shape) == 0 and ndim > 2
```
gautierronan, line 349: Fixed separately in https://github.com/dynamiqs/dynamiqs/pull/864 since it's unrelated to `mcsolve`.

gautierronan (changes requested):
Great PR @dkweiss31! Thanks again for this very valuable contribution.

Here's a first batch of review comments. I still have to review:

- [x] the core logic inside `integrators/mcsolve/diffrax_integrator.py`
- [x] the results class inside 'result.py`
- [ ] tests

dkweiss31:
Thanks for this first review @gautierronan !

Confused why ci is failing? Something about TimeArray?

dynamiqs/integrators/core/diffrax_integrator.py
gautierronan: You don't actually need to compute the non-hermitian Hamiltonian, you only need to compute its action on a ket. So one way to go about it is to directly compute `sum_qarray(-1j * H @ y, *[-0.5 * _L.dag() @ (_L @ y) for _L in L])`. Whether that's the optimal way or writing it, I'm not sure, but it retains the $O(n^2)$ complexity.

gautierronan: If possible, it would be great to benchmark this actually. I'm guessing which version has the shortest runtime will depend mostly on two factors:
 - Hilbert space size (yours should win in small sizes)
 - Sparsity (yours should win if all jump operators are sparse)

If this is true, we should probably add an option to choose between both versions (but we can start with just one for now to not overload the PR).

gautierronan: Here's for instance something I wrote recently to benchmark different `mesolve` vector fields with and without vectorization of the Liouvillian:

```python
import jax
import jax.numpy as jnp
import dynamiqs as dq

# === Prepare objects ===
# parameters
n = 500
delta = 0.1
epsilon = 2
gamma = 1
nth = 0.8

# operators
a = dq.destroy(n)
H = delta * dq.dag(a) @ a + epsilon * (a + dq.dag(a))
L = [jnp.sqrt(gamma * (1 + nth)) * a, jnp.sqrt(gamma * nth) * dq.dag(a)]
y = dq.todm(dq.fock(n, 0))
liouv = dq.slindbladian(H, L)
yvec = dq.operator_to_vector(y)

# === Functions to benchmark ===
@jax.jit
def vf_dynamiqs(H, L, y):
    """Dynamiqs vector field."""
    L = dq.stack(L)
    Ld = L.dag()
    LdL = (Ld @ L).sum(0)
    tmp = (-1j * H - 0.5 * LdL) @ y + 0.5 * (L @ y @ Ld).sum(0)
    return tmp + tmp.dag()

@jax.jit
def vf_liouvillian(liouv, yvec):
    """Liouvillian vector field."""
    return liouv @ yvec

@jax.jit
def vf_liouvillian_dynamic(H, L, yvec):
    """Liouvillian vector field with dynamical Liouvillian computation."""
    liouv = dq.slindbladian(H, L)
    return liouv @ yvec

# === Run benchmarks ===
vf_dynamiqs(H, L, y).block_until_ready()
%timeit vf_dynamiqs(H, L, y).block_until_ready()
# n=50  / 167 µs ± 9.51 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500 / 38.7 ms ± 10.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

vf_liouvillian(liouv, yvec).block_until_ready()
%timeit vf_liouvillian(liouv, yvec).block_until_ready()
# n=50  / 68.5 µs ± 4.41 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500 / 2.32 ms ± 261 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

vf_liouvillian_dynamic(H, L, yvec).block_until_ready()
%timeit vf_liouvillian_dynamic(H, L, yvec).block_until_ready()
# n=50  / 134 µs ± 3.62 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500 / 10.1 ms ± 808 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

gautierronan:
> Confused why ci is failing? Something about TimeArray?

Yes, we recently renamed `TimeArray` to `TimeQArray` to highlight the fact that it only works with `QArrays` now (unfortunately!). See https://github.com/dynamiqs/dynamiqs/pull/853. I guess you need to delete the `time_array.py` file, and maybe look for any instance of `TimeArray` in your branch.

dynamiqs/integrators/core/diffrax_integrator.py
dkweiss31: Ahh I see! This is a very good point. Which would you prefer we go with for now for this PR before revisiting? I'm happy with either solution.

gautierronan: From a quick benchmark, it seems my version is always faster actually. So I'd suggest going with this.

<details>
<summary>Here's the corresponding benchmark</summary>

```python
import jax
import jax.numpy as jnp
import dynamiqs as dq
from dynamiqs.qarrays.utils import sum_qarrays

dq.set_layout('dense')

# === Prepare objects ===
# parameters
n = 5000
delta = 0.1
epsilon = 2
gamma = 1
nth = 0.8

# operators
a = dq.destroy(n)
H = delta * dq.dag(a) @ a + epsilon * (a + dq.dag(a))
L = [jnp.sqrt(gamma * (1 + nth)) * a, jnp.sqrt(gamma * nth) * dq.dag(a)]
y = dq.fock(n, 0)

# === Functions to benchmark ===
@jax.jit
def vf1(H, L, y):
    Hnh = sum_qarrays(-1j * H, *[-0.5 * _L.dag() @ _L for _L in L])
    return Hnh @ y

@jax.jit
def vf2(H, L, y):
    return sum_qarrays(-1j * H @ y, *[-0.5 * _L.dag() @ (_L @ y) for _L in L])

# === Run benchmarks ===
vf1(H, L, y).block_until_ready()
%timeit vf1(H, L, y).block_until_ready()
# === sparse
# n=50   / 51.1 µs ± 5.74 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500  / 50.6 µs ± 1.17 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=5000 / 59.2 µs ± 616 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# === dense
# n=50   / 86.9 µs ± 307 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500  / 12.6 ms ± 640 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# n=5000 / 9.4 s ± 453 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

vf2(H, L, y).block_until_ready()
%timeit vf2(H, L, y).block_until_ready()
# === sparse
# n=50   / 49.1 µs ± 223 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500  / 50.2 µs ± 855 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=5000 / 53.4 µs ± 1.29 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# === dense
# n=50   / 50 µs ± 3.21 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# n=500  / 832 µs ± 62.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
# n=5000 / 208 ms ± 30.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
</details>

dynamiqs/result.py
```diff
@@ -287,6 +289,158 @@ class MESolveResult(SolveResult):
...
+        no_jump_states _(Array)_: Saved no-jump states.
+        final_no_jump_state _(Array)_: Saved final no-jump state.
+        jump_states _(Array)_: Saved states for jump trajectories.
+        final_jump_states _(Array)_: Saved final states for jump trajectories.
+        no_jump_prob _(Array)_: No jump probability.
+        jump_times _(Array)_: Times at which each trajectory experienced a jump. This
+            quantity has shape ..., options.max_jumps where the array is filled with
+            nans for the final options.max_jumps - num_jumps values.
+        num_jumps _(Array)_: Number of jumps each jump trajectory experienced. The times
+            at which each jump occurred is saved in jump_times.
```
gautierronan, line 325: This API doesn't quite make sense to me. `mcsolve` is a function to solve the Lindblad master equation, so the only thing it must return is a set of density matrices at each time specified in `tsave`, and the corresponding `expects` is they're requested (in a very similar manner to `mesolve`). 

Of course, because it is computed in a peculiar way, we can definitely also return information regarding the no-jump/jump trajectories, but it should be secondary and IMO wrapped in a practical manner. Maybe all these should be inside a `nojump_trajectories` and `jump_trajectories` data classes, somewhat like you're implementing in the backend already?

gautierronan (changes requested):
Here is a second batch of review comments, mostly about the integration method and the results API.

I also added two commits (https://github.com/dynamiqs/dynamiqs/pull/748/commits/875e9ce52e7ad6677e6b32cf8b382341cffa1d36 and https://github.com/dynamiqs/dynamiqs/pull/748/commits/ba55857a7254eed6729eaf224314aed87d77bc87) for a small refactoring to avoid duplicate code. It's breaking the CI, but only for a single `mcsolve` related test, so I'm guessing it's rather a statistical error.

My two main comments are regarding:
- the sampling of `new_times` with respect to the user-provided `tsave`. I think `new_times` should always be a sample of `tsave`, and not a completely new object (more details below).
- the results API, which should reflect the fact that we solve the ME, and not the jump SSE.

Also, I find the whole integration a bit complicated, but I could not find a way to really improve it. I think the algorithm of Abdelhafez et al. might be at fault, and that we could improve upon it with smarter trajectory samplings (but this is completely different beast to handle). For instance, I don't see a good reason why all the sampled jump trajectories have equal probabilities. Maybe we could find a smart way to sample the most probable jump trajectories instead?

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,279 @@
...
+                dt0=tsave[1] - tsave[0],
+                y0=y0,
+                saveat=saveat,
+                stepsize_controller=self.stepsize_controller,
+                adjoint=adjoint,
+                event=event,
+                max_steps=self.max_steps,
+                progress_meter=self.options.progress_meter.to_diffrax(),
+            )
+        return solution
```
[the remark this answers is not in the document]
gautierronan: (see latest two commits for a proposed fix)

dynamiqs/integrators/apis/mcsolve.py
dkweiss31: Right: in fact the behavior here is a bit nonstandard and depends on `options.cartesian_batching`: here is my proposed change. Let me know if you think this api should be changed
```suggestion
        keys _(Array of shape (...ntraj,))_: PRNG keys used to sample the jump trajectories.
            With options.cartesian_batching=True, this array must be 1D and the number of
            elements defines the number of sampled trajectories, not including the no-jump
            trajectory. With options.cartesian_batching=False, the leading dimensions 
            of keys must be broadcastable with those of H, Ls, psi0, with the final
            dimension determining the number of jump trajectories.  
```

dynamiqs/integrators/core/diffrax_integrator.py
dkweiss31: Got it, thanks for running the benchmark!

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # === solve differential equation with diffrax
+        return self.diffeqsolve(
+            t0=tsave[0], t1=tsave[-1], y0=y0, saveat=saveat, event=event
+        )
+
+    def run(self):
+        # === run no jump, extract no-jump probability
+        no_jump_solution = self._run(self.y0, self.ts, 0.0)
+        no_jump_saved = jax.vmap(self.save)(unit(no_jump_solution.ys[0]))
```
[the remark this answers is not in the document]
dkweiss31: > This line should be called later on, together with the save logic.

Why do you say that? I think this needs to be called first to extract the no-jump probability.



> Not sure, but the `jax.vmap` might be superfluous here.

I tried removing the vmap and tests failed, I guess the `self.save` doesn't natively support batching over times?

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+            no_jump_result, jump_result, no_jump_prob, jump_times, num_jumps
+        )
+
+    def _loop_over_jumps(self, key: Array, no_jump_prob: Array) -> State:
+        """loop over jumps until the simulation reaches the final time"""
+        rand_key, sample_key = jax.random.split(key)
+        rand = jax.random.uniform(rand_key, minval=no_jump_prob)
+        first_result = self._run(self.y0, self.ts, rand)
+        first_jump_time = first_result.ts[1][0]
+        time_diffs = self.ts - first_jump_time
```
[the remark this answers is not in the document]
dkweiss31: I have it this way because the loop is reserved for: applying a jump, and then solving until the next termination of `diffeqsolve` (whether that be because of an event or because `t1` is reached). I could alternatively have the loop be: solve until termination, then apply a jump. In this case we would still have a part of the evolution outside the loop, that is the final stretch from last jump to the final time. One needs to choose one or the other, and I chose the former.

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+class MCSolveDiffraxIntegrator(MCDiffraxIntegrator, MCSolveIntegrator, SolveSaveMixin):
+    """Integrator computing the time evolution of the Monte-Carlo unraveling of the
+    Lindblad master equation using the Diffrax library."""
+
+    def _run(self, y0, tsave, rand):
+        # === prepare saveat
+        subsaveat_a = dx.SubSaveAt(ts=tsave)  # save solution regularly
+        subsaveat_b = dx.SubSaveAt(t1=True)  # save last state
```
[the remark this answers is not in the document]
dkweiss31: Actually having both is quite useful because `subsaveat_b` contains the info about the jump time and the final state at the jump time

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+        jump_result = self.traj_result(jump_saved)
+
+        return self.result(
+            no_jump_result, jump_result, no_jump_prob, jump_times, num_jumps
+        )
+
+    def _loop_over_jumps(self, key: Array, no_jump_prob: Array) -> State:
+        """loop over jumps until the simulation reaches the final time"""
+        rand_key, sample_key = jax.random.split(key)
+        rand = jax.random.uniform(rand_key, minval=no_jump_prob)
```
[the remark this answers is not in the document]
dkweiss31: Hmm, I like the idea of a more explicit name but maybe we still want `rand` in it to signal the randomness aspect? what about `rand_jump_norm`?

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+            # extract saved y values to interpolate, replace infs with nans
+            psis_after_jump = result_after_jump.ys[0]
+            psi_inf_to_nan = jtu.tree_map(
+                lambda _y: jnp.where(jnp.isinf(_y), jnp.nan, _y), psis_after_jump
+            )
+            # setting fill_forward_nans_at_end prevents nan states from getting saved
+            psi_coeffs = dx.backward_hermite_coefficients(
+                new_times, psi_inf_to_nan, fill_forward_nans_at_end=True
+            )
+            psi_interpolator = dx.CubicInterpolation(new_times, psi_coeffs)
```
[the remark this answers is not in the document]
dkweiss31: This is a very tricky point and something that significantly complicates the implementation. The issue is mentioned in the first two lines of the extended comment, which is that `new_times` needs to have the same shape as `self.ts`. It is not enough for the shape of the output of `_outer_body_fun` to be static, but indeed all array sizes need to be determined at compile time for all code underneath a jit. Thus the `new_times` array must have the same shape as `self.ts`. 

What I don't think I've explored that this comment made me think of (maybe this is what you meant) is making `new_times` an array that has a static shape, filled only with times from the original `self.ts` and then padded with nans. Preliminary tests with `dx.diffeqsolve` indicate tha diffrax does not complain if some of the times in `SaveAt` are nans. This would avoid the need for interpolation, I think.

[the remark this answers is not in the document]
dkweiss31: See 2cf82b6d1725555e05f35577a0470ddc2d01fb91 below for a simplified implementation that does not require interpolation based on the above discussion. Thanks very much for making this point!

dynamiqs/integrators/mcsolve/diffrax_integrator.py
```diff
@@ -0,0 +1,258 @@
...
+        """given a state psi at time t that should experience a jump,
+        randomly sample one jump operator from among the provided jump_ops.
+        The probability that a certain jump operator is selected is weighted
+        by the probability that such a jump can occur. For instance for a qubit
+        experiencing amplitude damping, if it is in the ground state then
+        there is probability zero of experiencing an amplitude damping event.
+        """
+        Ls = stack([L(t) for L in self.Ls])
+        Lsd = dag(Ls)
+        probs = expect(Lsd @ Ls, psi)
```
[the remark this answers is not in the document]
dkweiss31: Re your comment in vector field, should we do soemthing like
```suggestion
        Ls = stack([L(t) for L in self.Ls])
        Ls_on_psi = [L @ psi for L in Ls]
        probs = jnp.asarray([_psi.dag() @ _psi for _psi in Ls_on_psi]).squeeze(-1)
```
so that we only have matrix-vector products instead of the current matrix-matrix product?

I think though you are pointing out something different, that is the use of stack: if I just use a listcomp, I get an error when I try to sample the indicated jump operator
```    
return Ls[sample_idx]
jax.errors.TracerIntegerConversionError: The __index__() method was called on traced array with shape int32[].
The error occurred while tracing the function _fn at /Users/danielweiss/PycharmProjects/equinox/equinox/_make_jaxpr.py:36 for jit. This concrete value was not available in Python because it depends on the values of the arguments _dynamic_flat[8] and _dynamic_flat[11].
```
So I'm not sure how to proceed without using `stack`

dynamiqs/result.py
dkweiss31: Thanks for this comment Ronan: I now have the states saved explicitly as density matrices as appropriate for a master-equation solver. I now also have dedicated dataclasses that save data associated with the jump and no-jump trajectories. I have made that data (states saved as kets, no-jump probability, jump times, number of jumps) accessible as part of the public api still, let me know your thoughts on that.

dkweiss31:
> Here is a second batch of review comments, mostly about the integration method and the results API.
> 
> I also added two commits ([875e9ce](https://github.com/dynamiqs/dynamiqs/commit/875e9ce52e7ad6677e6b32cf8b382341cffa1d36) and [ba55857](https://github.com/dynamiqs/dynamiqs/commit/ba55857a7254eed6729eaf224314aed87d77bc87)) for a small refactoring to avoid duplicate code. It's breaking the CI, but only for a single `mcsolve` related test, so I'm guessing it's rather a statistical error.
> 
> My two main comments are regarding:
> 
> * the sampling of `new_times` with respect to the user-provided `tsave`. I think `new_times` should always be a sample of `tsave`, and not a completely new object (more details below).
> * the results API, which should reflect the fact that we solve the ME, and not the jump SSE.
> 
> Also, I find the whole integration a bit complicated, but I could not find a way to really improve it. I think the algorithm of Abdelhafez et al. might be at fault, and that we could improve upon it with smarter trajectory samplings (but this is completely different beast to handle). For instance, I don't see a good reason why all the sampled jump trajectories have equal probabilities. Maybe we could find a smart way to sample the most probable jump trajectories instead?

Thanks for this careful review @gautierronan . I've tried to address all of your comments, see my responses to your review comments. In response to your main comments:

1. I have adjusted the sampling of `new_times` to be only times in the original `self.ts`. This eliminates entirely the need for interpolation: please see the new simplified implementation
2. I've adjusted the api to return density matrices, with the underlying jump and no-jump information saved in dedicated dataclasses. That information is still available as part of the public api, which we can discuss if you disagree with.

I'm interested in your smarter trajectory samplings idea: how would we identify the most probable jump trajectories?

basnijholt:
@dkweiss31, is this in a usable state ATM?

We would like to start using it ASAP 😄

gautierronan:
@basnijholt I believe it's usable, but not in its final form yet.

Our plan for January is mainly to push these PRs:
- https://github.com/dynamiqs/dynamiqs/pull/887 (merged today)
- https://github.com/dynamiqs/dynamiqs/pull/881 (probably next week)
- https://github.com/dynamiqs/dynamiqs/pull/877 
- https://github.com/dynamiqs/dynamiqs/pull/748

such that we may settle all the stochastic-related stuff. We're also planning to release v0.3.0 this week with QArray support, and so stochastic solvers will probably be for v0.3.1. Apologies for the long wait, we're trying to push this asap!

dkweiss31:
Hi @gautierronan , I'm wondering what the next steps are for this PR? Would be great to tie a bow on this long mcsolve journey!

gautierronan:
> Hi @gautierronan , I'm wondering what the next steps are for this PR? Would be great to tie a bow on this long mcsolve journey!

Apologies for the long wait. My agenda is finally freeing up.

With @pierreguilmin we discussed a bit the future of `mcsolve`, especially in regard to the recent stochastic solvers we pushed to main. We're thinking of doing the following:

1. Redirect your `mcsolve` implementation towards `jssesolve` (see the proposed API [here](https://www.dynamiqs.org/stable/python_api/integrators/jssesolve.html)). This is purely a naming change, since the Monte-Carlo solver for the master equation and the stochastic jump equation correspond to the same equation. 
2. Add a dedicated solver/method for `mesolve` that would internally use `jssesolve` to estimate a master equation integration, i.e. `dq.mesolve(..., solver=dq.solver.MonteCarlo(keys=keys))`. 

We believe this really answers the two very different needs for `mcsolve`: people who want to simulate stochastic trajectories (`jssesolve`) and people who want to simulate the master equation with an efficient numerical method (`dq.solver.MonteCarlo`). Our issue with the current implementation of `mcsolve` was that it didn't really correspond to a single well-defined ODE.

If you agree with this and with the API, I propose the following next steps:

- I can deal with merging your PR into main, and taking care of aligning with API with our proposed `jssesolve` one.
- I'd like to spend a bit of time trying to simplify the core implementation of this PR. I feel like there's many imbricated functions that we could simplify (not sure, but I'd like to give it a try).
- @pierreguilmin reviews
- In a separate PR, we can deal with point (2) above, which should be quite straightforward once (1) is done.

My timeline for the first two actions is EOW.

dkweiss31:
> 1. Redirect your `mcsolve` implementation towards `jssesolve` (see the proposed API [here](https://www.dynamiqs.org/stable/python_api/integrators/jssesolve.html)). This is purely a naming change, since the Monte-Carlo solver for the master equation and the stochastic jump equation correspond to the same equation.

This is non-obvious to me: it seems there are terms in the jump SSE that don't show up in the case of `mcsolve`? I'm willing to believe that the additional terms are the result of writing the equation that `mcsolve` attempts to solve as a true ODE, but I'd advocate for an explicit discussion in the `jssesolve` API of that "mapping."
> 2. Add a dedicated solver/method for `mesolve` that would internally use `jssesolve` to estimate a master equation integration, i.e. `dq.mesolve(..., solver=dq.solver.MonteCarlo(keys=keys))`.

This sounds great.
> We believe this really answers the two very different needs for `mcsolve`: people who want to simulate stochastic trajectories (`jssesolve`) and people who want to simulate the master equation with an efficient numerical method (`dq.solver.MonteCarlo`). Our issue with the current implementation of `mcsolve` was that it didn't really correspond to a single well-defined ODE.

See my questions above: also I'm wondering if it makes sense to have an alias of `mcsolve` that uses `jssesolve` under the hood? I worry that the average user is less familiar with the jump SSE than `mcsolve` (this is certainly true for me, but I understand it is perhaps not true for you good folks), and it would not be obvious to me if I saw only `jssesolve` that I'd be able to extract `mcsolve` functionality from it.
> * I can deal with merging your PR into main, and taking care of aligning with API with our proposed `jssesolve` one.
> * I'd like to spend a bit of time trying to simplify the core implementation of this PR. I feel like there's many imbricated functions that we could simplify (not sure, but I'd like to give it a try).

Both sound great, thanks!

gautierronan:
The fact that `mcsolve` and `jssesolve`are the same equation is indeed non-obvious, but they are! If you look at the JSSE equation, you indeed find the usual non-hermitian evolution in the first two terms ($-i H - 1/2 \sum_k L_k^\dagger L_k$), with an additional term to make sure the state is normalized at all times during this evolution ($-\sum_k \langle L_k^\dagger L_k \rangle$). Then, there is the jump term that projects onto a random jump operator with a given probability, just like in `mcsolve` ($\sum_k L_k / \sqrt{\langle L_k^\dagger L_k \rangle} - 1$).

In other words, if we ask "what is the ODE that `mcsolve` solves", I think there are only two valid answers: the JSSE or the ME. This is made very clear in the initial QuTiP paper from 2012, for which `mcsolve` already existed. At the time, `mcsolve` was just a way to solve the ME, just like they had an `odesolve` and a `essolve` that were both solvers for the ME!

Today in Dynamiqs, we made the clear choice to separate the ODE solvers (`sesolve`, `mesolve`, ...) from the methods with which they are solved (`Tsit5`, `Dopri5`, `Euler`, `Expm`, ...). If we were to introduce `mcsolve` as is, it would go against this choice, as we would have two different solvers for the ME (or two solvers for the JSSE). I know this is what QuTiP is doing today, but it was not their original choice, and I don't believe this is the right way to advance.

> I worry that the average user is less familiar with the jump SSE than mcsolve

That being said, I agree with you on this point. I think that giving good exposure to `dq.solver.MonteCarlo` should answer part of this concern. We can also make a clear note in the documentation about the link between `dq.solver.MonteCarlo` and `jssesolve` (inside each of these two docstrings at least).

dkweiss31:
Understood! Just two more comments from my end:

In the current `mcsolve` implementation, the state is only renormalized after a jump occurs. We then use the decay of the norm to help decide when to have jumps. Not sure if that changes anything about this mapping onto `jssesolve`.

I see also in the `jssesolve` API that the only solver (method?) supported is `EulerMaruyama`. For `mcsolve` as written, any of the solvers `Tsit5`, `Dopri8`, etc. supported as in e.g. `sesolve` or `mesolve`. Not sure how to resolve this inconsistency (if there indeed is one).

gautierronan:
> In the current mcsolve implementation, the state is only renormalized after a jump occurs. We then use the decay of the norm to help decide when to have jumps. Not sure if that changes anything about this mapping onto jssesolve.

That's equivalent. With the current implementation, we could also renormalize the state at all times and just keep track of the change in normalization for random number sampling.

> I see also in the jssesolve API that the only solver (method?) supported is EulerMaruyama. For mcsolve as written, any of the solvers Tsit5, Dopri8, etc. supported as in e.g. sesolve or mesolve. Not sure how to resolve this inconsistency (if there indeed is one).

Indeed, I have remove this default EulerMaruyama default method. It will probably be reimplemented in a later PR, but is separate from the Tsit5, Dopri8, etc methods.

pierreguilmin:
Great!

Quick remark:

> > I see also in the jssesolve API that the only solver (method?) supported is EulerMaruyama. For mcsolve as written, any of the solvers Tsit5, Dopri8, etc. supported as in e.g. sesolve or mesolve. Not sure how to resolve this inconsistency (if there indeed is one).

> Indeed, I have remove this default EulerMaruyama default method. It will probably be reimplemented in a later PR, but is separate from the Tsit5, Dopri8, etc methods.

`Tsit5`, `Dopri8` are not methods to solve SDEs. We could for example call the method you guys are implementing `Event`, where one argument of the method would be the ODE solver used for the no-jump evolution. Other methods would be `EulerMaruyama`, or `Milstein`, or `Rouchon` (SDE solvers). So to be clear:
```diff
- dq.jssesolve(..., method=dq.method.Tsit5())
+ dq.jssesolve(..., method=dq.method.Event(ode_solver=Tsit5()))
+ dq.jssesolve(..., method=dq.method.EulerMaruyama())
```
It's very important to make this distinction in my opinion.

pierreguilmin:
Random other remarks:
- `root_finder` is specific to this "Event" method (and should not be an option of the general solver).
- `max_jumps` is the same as `nmaxclick` (the distinction between clicks and jumps is that the detector can click with no jump for jsmesolve, and we match the names for `jssesolve`)
- Can we move the solver implementation to a different file? It's not a diffrax integrator _per se_, plus this file is becoming too large.

gautierronan:
@dkweiss31 Absolutely. To be very clear, I see the `smart_sampling=False` option implemented now only as an intermediate step, and understand very well the need for the `smart_sampling=True` option. I simply find important to implement the former first, such that the overall code structure is fixed. Hopefully, adding this option should not be too hard once we have converged on this version.

I didn't think yet in much details as to how to implement it, but I have some ideas. Mainly, one large difference with your previous implementation is that vectorization over `keys` is now handled inside `_vectorized_jssesolve` (like it is for other stochastic solvers). This should eventually help with advanced batching formats.

dynamiqs/integrators/core/save_mixin.py
```diff
@@ -80,3 +89,18 @@ def postprocess_saved(self, saved: Saved, ylast: PyTree) -> Saved:
...
+
+
+class JumpSolveSaveMixin(SolveSaveMixin):
+    """Mixin to assist jump SSE/SME integrators computing time evolution with data
+    saving.
+    """
+
+    def save(self, y: PyTree) -> Saved:
+        # force state normalization before saving
+        return super().save(unit(y))
```
gautierronan, line 101: Side note @pierreguilmin (unrelated to this PR). If we want to always return normalized states in other solvers, this is quite straightforward with this construction, and we could easily add a tracker for the accumulated norm error too. We might want to generalize this to other solvers.

pierreguilmin (approved):
Wow, that's a beautiful piece of code! Congratulations to @dkweiss31 for the heavy lifting, and to @gautierronan for adapting it to the library. I made a few small modifications. The two potentially controversial ones are:
- Renamed "no jump" to "no click", because for the jump SME click != jump.
- Removed the batching over `keys`, it would require proper documentation and support across all our stochastic solvers. We can add this later (although I'm not sure it's needed).

I didn't thoroughly review the logic in `event_integrator.py`, I have a few ideas for improvement but nothing big or urgent. Since it's well isolated in a single file, it will be easy to iterate on it.

What's next for this PR? I propose merging it into `main` after your reviews, and continuing the development, specifically:
- Reimplement @dkweiss31's `smart_sampling` option (and the associated `noclick_states`, `noclick_prob`).
- We need a way to specify the stepsize controller `dtmax` when `root_finder=None`, otherwise there's no way for the user to control the precision of the click times.
- Add a `MonteCarlo` method to `mesolve()`.

Fun fact: @dkweiss31 we first met almost exactly a year ago after our March Meeting talk with @gautierronan (on March 5th!). I still remember you coming to the mic to ask "Do you have a Monte-Carlo solver?". It's been a long journey, but we're finally seeing the end! 😁

dkweiss31:
> * Renamed "no jump" to "no click", because for the jump SME click != jump.

I'm happy with this change! We have gone all in on the interpretation in terms of stochastic trajectories where the jumps are signaled by a detection event, so this change is consistent.
> * Removed the batching over `keys`, it would require proper documentation and support across all our stochastic solvers. We can add this later (although I'm not sure it's needed).
> 

Agreed, happy to discuss as part of a future PR if we think this is necessary. 
> 
> Fun fact: @dkweiss31 we first met almost exactly a year ago after our March Meeting talk with @gautierronan (on March 5th!). I still remember you coming to the mic to ask "Do you have a Monte-Carlo solver?". It's been a long journey, but we're finally seeing the end! 😁

I remember it well! You said something to the effect of "no, but you should implement it!" Everyone laughed, but here we are. It obviously took longer than expected but its been an amazing ride and more to come!

dkweiss31:
I'd be very happy to give this a go! I'm hoping that what I had before can be ported over with minimal headache. I'll try and come up with a detailed brief of how I'd implement it and what the results should look like before delving in full steam.

PR #756, opened 2024-10-14
Add returns to plot functions

gautierronan:
Proposal to return plotted images/collections/lines in all our plot functions.

It's quite useful to be able to customize plots, and it essentially doesn't cost us anything. Plus it mimics the behavior of many `matplotlib` functions, e.g. `dq.plot.wigner` is essentially a glorified `plt.imshow` so it should return the same kind of object.

### DIFF
--- a/dynamiqs/plot/hinton.py
+++ b/dynamiqs/plot/hinton.py
@@ -121,6 +121,7 @@ def _plot_hinton(
 
     return squares
 
+
 @optional_ax
 def hinton(
     x: ArrayLike,
--- a/dynamiqs/plot/wigner.py
+++ b/dynamiqs/plot/wigner.py
@@ -235,6 +235,7 @@ def wigner_mosaic(
 
     return ims
 
+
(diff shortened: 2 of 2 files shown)


abocquet (approved):

pierreguilmin:
I'm torn on this idea, on one hand it's convenient to obtain the main plot object from the function for advanced plots, on the other hand:
- Plot methods will output something in every Python intepreter (see the failing CI > https://github.com/dynamiqs/dynamiqs/actions/runs/11327666084/job/31499221146?pr=756) or notebook, which is a bit invading, especially since the return won't be used most of the time.
- Every sub-object of the plot can already be accessed in one line with the native Matplotlib methods `ax.get_images()`/`ax.get_children()`/`ax.patches`.

gautierronan:
Following discussion with @pierreguilmin, closing for now because we're not sure the upsides outweigh the downsides. Happy to reopen this if we change our mind later on, depending on requests too.

PR #778, opened 2024-10-22
Simplify implementation of vectorization utilities

gautierronan:
Small simplifications on the vectorization code (remove the `False`/`None` logic, fix typings, use `jax.tree`).

Suggest reviewing commit by commit.

Also, I think it will bug if a jump operator is a `SummedTimeArray` which is batched over. We should probably play the same tricks as with the Hamiltonian, although this is quite niche.


pierreguilmin (approved):
Nice 🥳

PR #786, opened 2024-11-13
Add references btw lindbladian functions

gaspardbb:
Spent some time reimplementing `slindbladian` because I did not know it existed. 
Don't know if you have specific ideas on how to write those "see also", for instance in [jax.linalg.eig](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.eig.html). If that's the case lmk and I'll change it

### DIFF
--- a/dynamiqs/utils/vectorization.py
+++ b/dynamiqs/utils/vectorization.py
@@ -223,9 +223,9 @@ def slindbladian(H: ArrayLike, jump_ops: ArrayLike) -> Array:
     Note:
         This superoperator is also sometimes called *Liouvillian*.
 
-    Note:
-        See also [`dq.lindbladian`][dynamiqs.utils.quantum_utils.general.lindbladian] to
-        access the Lindbladian only through matrix-vector product.
+    See also:
+        - [`dq.lindbladian`][dynamiqs.utils.quantum_utils.general.lindbladian]:
+        apply the Lindbladian only through matrix-matrix products.
 
     Args:
         H _(array_like of shape (..., n, n))_: Hamiltonian.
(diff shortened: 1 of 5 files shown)


dynamiqs/utils/vectorization.py
```diff
@@ -223,6 +223,10 @@ def slindbladian(H: ArrayLike, jump_ops: ArrayLike) -> Array:
     Note:
         This superoperator is also sometimes called *Liouvillian*.
 
+    Note:
+        See also [`dq.lindbladian`][dynamiqs.utils.quantum_utils.general.lindbladian] to
+        access the Lindbladian only through matrix-vector product.
```
gautierronan, line 228: Maybe with the following formatting, which is indeed closer to JAX? We don't have a specific admonition for this, but maybe we could add one.
```suggestion
    Note: See also
        - [`dq.lindbladian`][dynamiqs.utils.quantum_utils.general.lindbladian]:
        access the Lindbladian only through matrix-vector product.
```

gautierronan, line 228: Also, if should say "matrix-matrix product" instead of "matrix-vector product" I think.

gautierronan (changes requested):
Lovely thank you! Could you also do `dq.dissipator` and `dq.sdissipator` in a similar fashion?

dynamiqs/utils/vectorization.py
gaspardbb: I think it would be useful to add one indeed. Those links are really useful for newcomers in those big libraries  dynamiqs belongs to. 
If you do that I'll wait for the formatting before changing my commit

gautierronan: Did that in https://github.com/dynamiqs/dynamiqs/pull/787 :)

gautierronan (approved):
LGTM! Thank you @gaspardbb

PR #787, opened 2024-11-13
Introduce a See Also admonition

gautierronan:
See https://github.com/dynamiqs/dynamiqs/pull/786. I kept the same color as the Examples admonition for now. I think the overall color palette of the website will be reworked at some point.


gaspardbb (approved):
Looks good to me!

PR #798, opened 2024-11-18
Document conversion functions

gautierronan:
- Add docstrings for all QArray conversion functions
- Merge `qarrays/utils.py` and `qarrays/type_conversion.py` into a single file (`qarrays/utils.py`) for single table in documentation API index
- Add missing `dq.to_numpy` function
- Minor fixes in these functions

Suggest reviewing commit by commit.


dynamiqs/time_array.py
```diff
@@ -206,13 +206,13 @@ def timecallable(
...
+        QArray: shape=(2, 2), dims=(2,), dtype=complex64, layout=dense
+        [[0.5+0.j 0. +0.j]
+         [0. +0.j 0.5+0.j]]
         >>> H(1.0)
-        QArray: shape=(2, 2), dims=(2,), dtype=float32, layout=dense
-        [[1. 0.]
-         [0. 0.]]
+        QArray: shape=(2, 2), dims=(2,), dtype=complex64, layout=dense
+        [[1.+0.j 0.+0.j]
+         [0.+0.j 0.+0.j]]
```
pierreguilmin, line 215: This is quite annoying. Is there a simple workaround? (we can merge the PR and fix it later)

pierreguilmin: Also in `dynamiqs/utils/quantum_utils/general.py`.

dynamiqs/qarrays/utils.py
```diff
@@ -80,3 +179,220 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+        QArray: shape=(3, 2, 2), dims=(2,), dtype=float32, layout=dia, ndiags=1
+        [[[1.+0.j   ⋅   ]
+          [  ⋅    1.+0.j]]
+        <BLANKLINE>
+         [[1.+0.j   ⋅   ]
+          [  ⋅    1.+0.j]]
+        <BLANKLINE>
+         [[1.+0.j   ⋅   ]
+          [  ⋅    1.+0.j]]]
+    """
```
pierreguilmin, line 356: There's something fishy with the indicated and actual types here.

pierreguilmin (changes requested):
Huge work 🤯 Minor comments + some commits (feel free to revert anything). I also added a doctest that fails, but I think it's a bug, it should pass and is a nice example for the documentation. 😇

dynamiqs/time_array.py
gautierronan: What do you find annoying? I think a QArray should always be complex valued, so IMO it's expected. We don't have to enforce this dtype (easy to change), but then the dtype can be `int32` or whatever else.

dynamiqs/qarrays/utils.py
gautierronan: It's related to the question above, on whether we want to enforce the dtype to be complex or not. For now, `sparsedia_from_dict` is NOT enforcing the dtype, but `to_jax()` (called by `SparseDIAQArray.__repr__`) is, hence the mismatch.

pierreguilmin:
I've added the dtype issue to the meta linear ticket on qarray, merging this as soon as the CI passes.

pierreguilmin (approved):

PR #802, opened 2024-11-21
Document qarray

pierreguilmin:
First version of the documentation.

### DIFF
--- a/dynamiqs/qarrays/qarray.py
+++ b/dynamiqs/qarrays/qarray.py
@@ -96,7 +96,7 @@ class QArray(eqx.Module):
     Note-: Shortcuts methods to use quantum utilities
         Many functions of the library can be called directly on a qarray rather than
         through the functional API. For example, you can use `x.dag()` instead of
-        `dq.isdag(x)`. Here is the complete list of these shortcuts:
+        `dq.dag(x)`. Here is the complete list of these shortcuts:
 
         | QArray method    | Corresponding function call                          |
         |------------------|------------------------------------------------------|
--- a/dynamiqs/qarrays/qarray.py
+++ b/dynamiqs/qarrays/qarray.py
@@ -93,35 +93,40 @@ class QArray(eqx.Module):
         addition/subtraction, element-wise multiplication and matrix multiplications
         with other qarray-like objects.
 
-    Note-: Shortcuts methods to use quantum utilities
+    Note: Shortcuts methods to use quantum utilities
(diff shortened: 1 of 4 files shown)


dynamiqs/qarrays/qarray.py
```diff
@@ -66,10 +66,61 @@ def _dims_from_qutip(dims: list[list[int]]) -> tuple[int, ...]:
...
+        layout _(Layout)_: Data layout, either `dq.dense` or `dq.dia`.
+        dims _(tuple of ints)_: Hilbert space dimension of each subsystem.
+        mT _(QArray)_: Returns the qarray transposed over its last two dimensions.
+
+    Note: Arithmetic operation support
+        Qarrays support elementary operations, such as element-wise
+        addition/subtraction, element-wise multiplication and matrix multiplications
+        with other qarray-like objects.
+
+    Note-: Shortcuts methods to use quantum utilities
```
gautierronan, line 96: I think this note and its table are quite important (equally to the rest of the docstring IMO). I would propose not having it as a note. This is what PyTorch does for instance, [here](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.mH).

Actually, I would even go further and provide a table for the _all_ methods of a `dq.QArray`, such that:

- for methods that are duplicates of quantum utilities (e.g. `.dag()`), just have a description saying "See `dq.this`" or "Alias of `dq.this`" (like pytorch does)
- for methods that are new to QArrays (e.g. `.conj()` or `reshape(*shape)`), have a short one-liner description + link to the full docstring just below in the page.

dynamiqs/qarrays/qarray.py
```diff
@@ -66,10 +66,61 @@ def _dims_from_qutip(dims: list[list[int]]) -> tuple[int, ...]:
...
+        Many functions of the library can be called directly on a qarray rather than
+        through the functional API. For example, you can use `x.dag()` instead of
+        `dq.isdag(x)`. Here is the complete list of these shortcuts:
```
gautierronan, line 99:
```suggestion
        `dq.dag(x)`. Here is the complete list of these shortcuts:
```
typo

gautierronan (changes requested):

pierreguilmin:
@gautierronan if you agree with the overall content I propose to merge this and iterate on top later for styling / minor modifications.

gautierronan (approved):
LGTM!

PR #804, opened 2024-11-21
Nested conversions

gautierronan:
Add support for nested array-like inputs in conversion functions.

For instance, we can now run something like `dq.asqarray([dq.sigmax(), [[1, 0], [0, -1]]])` which yields a QArray of shape (2, 2, 2), or run `dq.to_qutip([[dq.sigmax(), dq.sigmay()], [qutip.sigmax(), qutip.sigmay()]])` which yields a nested list of Qobjs, each of shape (2, 2).

The logic could probably be improved. In particular, we always go through jax arrays even for the sparse format, which is probably not the most efficient. But it seems quite robust at least.


pierreguilmin (changes requested):
Proposal for simplication, maybe I missed something.

pierreguilmin:
Meh, there was probably a new release of ruff, I don't have the time to look at it rn but it should be fairly easy. At least
```shell
pytest dynamiqs/qarrays/utils.py && pytest tests/qarrays/test_utils.py
```
runs fines on my computer, so I think I didn't break your PR. 🤞

pierreguilmin (approved):
Ideally we fix ruff on `main` and merge in `dev-qarray` then rebase this PR on top and merge. But if you want to first merge this one and do the main/dev-qarray merge I'm fine with it.

pierreguilmin (approved):

PR #806, opened 2024-11-23
Fix ruff on dev-qarray branch

gautierronan:
Just runs `ruff check --fix` on dev-qarray, and remove obsolete rule.


pierreguilmin (changes requested):
@gautierronan, we should really do this on `main` and _then_ merge with `dev-qarray`. All these fixes are due to the new ruff release https://astral.sh/blog/ruff-v0.8.0, we should also updage the dependency in `pyproject.toml`.

pierreguilmin:
Closed by #811.

PR #808, opened 2024-11-25
Fix JIT compatibility with state creation

gautierronan:
Fixes the following MWE, which made any function that called `dq.fock`/`dq.basis`/`dq.coherent` incompatible with the JIT.
```python
import dynamiqs as dq
import jax

@jax.jit
def f():
    return dq.fock(2, 0)

f()
```

(CI not passing because of new ruff version)


abocquet (approved):
LGTM 👍

pierreguilmin (approved):

PR #816, opened 2024-12-03
Simplify `dims` in conversion functions

pierreguilmin:
- Simplify `dims` management in conversion functions.
- Extract dims from sequence of objects with native dims, e.g. this returns the correct dims:

  ```python
  >>> dq.asqarray([dq.fock((3, 4), (0, 0)), qt.tensor(qt.fock(3, 1), qt.fock(4, 0))])
  QArray: shape=(2, 12, 1), dims=(3, 4), dtype=complex64, layout=dense
  [...]
  
  ```

### DIFF
--- a/dynamiqs/qarrays/qarray.py
+++ b/dynamiqs/qarrays/qarray.py
@@ -38,6 +38,24 @@ def _to_jax(x: QArrayLike) -> Array:
         return jnp.asarray(x)
 
 
+def _get_dims(x: QArrayLike) -> tuple[int, ...] | None:
+    if isinstance(x, Sequence):
+        sub_dims = [_get_dims(sub_x) for sub_x in x]
+        return sub_dims[0] if all(sd == sub_dims[0] for sd in sub_dims) else None
+    if isinstance(x, QArray):
+        return x.dims
+    elif isinstance(x, Qobj):
+        dims = np.max(x.dims, axis=0)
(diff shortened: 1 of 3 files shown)


dynamiqs/qarrays/qarray.py
```diff
@@ -38,6 +38,24 @@ def _to_jax(x: QArrayLike) -> Array:
...
+    elif isinstance(x, Qobj):
+        dims = np.max(x.dims, axis=0)
+        return tuple(dims.tolist())
+    else:
+        return None
+
+
+def _to_jax_and_dims(x: QArrayLike) -> tuple[Array, tuple[int, ...]]:
+    dims = _get_dims(x)
+    return _to_jax(x), dims
```
gautierronan, line 56: Nit: I don't quite see the need to introduce this function, as it's just making two separate calls to `_get_dims` and `_to_jax`. Might be more readable to keep these calls separated in `_asdense` and `_assparsedia`.

gautierronan (approved):
LGTM, small nit below, but feel free to ignore.

dynamiqs/qarrays/qarray.py
pierreguilmin: I think you're right, that's an artifact from the refactoring process, I'll remove it.

gautierronan (approved):

PR #817, opened 2024-12-03
Fix `ptrace` for qarrays

pierreguilmin:
We can now do `x.ptrace(*keep)` for a `QArray`, this was not yet ported.

### DIFF
--- a/dynamiqs/utils/quantum_utils/general.py
+++ b/dynamiqs/utils/quantum_utils/general.py
@@ -270,16 +270,27 @@ def ptrace(
         (3, 4, 5)
         >>> psi_abc.shape
         (60, 1)
-        >>> rho_a = dq.ptrace(psi_abc, 0, (3, 4, 5))
+        >>> rho_a = dq.ptrace(psi_abc, 0)
         >>> rho_a.dims
         (3,)
         >>> rho_a.shape
         (3, 3)
-        >>> rho_bc = dq.ptrace(psi_abc, (1, 2), (3, 4, 5))
+        >>> rho_bc = dq.ptrace(psi_abc, (1, 2))
         >>> rho_bc.dims
         (4, 5)
         >>> rho_bc.shape
         (20, 20)
+
+        If the input qarray-like object `x` does not hold Hilbert space dimensions, you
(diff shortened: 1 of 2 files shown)


dynamiqs/utils/quantum_utils/general.py
```diff
@@ -264,18 +266,23 @@ def ptrace(x: ArrayLike, keep: int | tuple[int, ...], dims: tuple[int, ...]) ->
...
         (60, 1)
-
-        # todo: temporary fix
-        # >>> rho_a = dq.ptrace(psi_abc, 0, (3, 4, 5))
-        # >>> rho_a.shape
-        # (3, 3)
-        # >>> rho_bc = dq.ptrace(psi_abc, (1, 2), (3, 4, 5))
-        # >>> rho_bc.shape
-        # (20, 20)
+        >>> rho_a = dq.ptrace(psi_abc, 0, (3, 4, 5))
```
gautierronan, line 273:
```suggestion
        >>> rho_a = dq.ptrace(psi_abc, 0))
```

dynamiqs/utils/quantum_utils/general.py
```diff
@@ -264,18 +266,23 @@ def ptrace(x: ArrayLike, keep: int | tuple[int, ...], dims: tuple[int, ...]) ->
...
+        >>> rho_a.dims
+        (3,)
+        >>> rho_a.shape
+        (3, 3)
+        >>> rho_bc = dq.ptrace(psi_abc, (1, 2), (3, 4, 5))
```
gautierronan, line 278:
```suggestion
        >>> rho_bc = dq.ptrace(psi_abc, (1, 2))
```

gautierronan: Might be useful to add an example with the use of explicit `dims` afterwards, but the basic example should be without.

gautierronan (changes requested):
Thanks for the PR. Comment below regarding the docstring example. Otherwise the implementation looks good!

gautierronan (dismissed):
Small bug in CI, but LGTM!

gautierronan (approved):

PR #820, opened 2024-12-04
Simplify documentation management and directory structure

pierreguilmin:
This decouples the structure of the directories from the documentation files generation. We can go from the current
```
dynamiqs/utils
├── __init__.py
├── jax_utils.py
├── operators.py
├── optimal_control.py
├── quantum_utils
│   ├── __init__.py
│   ├── general.py
│   └── wigner.py
├── states.py
└── vectorization.py
```
to
```
dynamiqs/utils
├── __init__.py
├── general.py
├── jax_utils.py
├── operators.py
├── optimal_control.py
├── states.py
├── vectorization.py
└── wigner.py
```
This is also needed in `dev-qarray` (for `isqarraylike` that is defined in the core `qarray.py` file but that we want to expose in the doc in qarray utils). There is a remaining issue: in `index.md` it forces us to use different tables. @gautierronan any idea for a clean fix?


gautierronan:
Right now the call to `::: dynamiqs.utils.general` generates the HTML for both tables using `module.html.jinja` + `table.html.jinja` which gives:
```html
<!-- dynamiqs.utils.general --> 
<div class="doc doc-object doc-module">
  <div class="doc doc-contents first">
    <div class="md-typeset__scrollwrap">
      <div class="md-typeset__table">
        <table>
          <colgroup>
            <col span="1" style="width: 30%;">
            <col span="1" style="width: 70%;">
          </colgroup>
          <tbody>
            <!-- ... --> 
          </tbody>
        </table>
      </div>
    </div>
  </div>
</div>
<!-- dynamiqs.utils.wigner_utils --> 
<div class="doc doc-object doc-module">
  <div class="doc doc-contents first">
    <div class="md-typeset__scrollwrap">
      <div class="md-typeset__table">
        <table>
          <colgroup>
            <col span="1" style="width: 30%;">
            <col span="1" style="width: 70%;">
          </colgroup>
          <tbody>
            <!-- ... --> 
          </tbody>
        </table>
      </div>
    </div>
  </div>
</div>
```

What we can do is modify `module.html.jinja` such that it directly redirects to `table.html.jinja` and such that we only generate the table content, e.g.:
```html
<!-- dynamiqs.utils.general --> 
<tbody>
  <!-- ... --> 
</tbody>
<!-- dynamiqs.utils.wigner_utils --> 
<tbody>
  <!-- ... --> 
</tbody>
```
Then, we would need to write the rest of the divs directly inside `index.md`, e.g. something like:
```markdown
<div class="doc doc-object doc-module">
  <div class="doc doc-contents first">
    <div class="md-typeset__scrollwrap">
      <div class="md-typeset__table">
        <table>
          <colgroup>
            <col span="1" style="width: 30%;">
            <col span="1" style="width: 70%;">
          </colgroup>

::: dynamiqs.utils.general
    options:
        table: true
        members:
        - dag
        - ...

::: dynamiqs.utils.wigner_utils
    options:
        table: true
        members:
        - wigner
 
        </table>
      </div>
    </div>
  </div>
</div>
```
This is rather easy to do, but we need to add this HTML structure throughout `index.md` which is kinda annoying. Another solution is to hard code the whole table directly in markdown, without even using the mkdocs helpers `::: module`. There are maybe cleaner solutions, but can't think of any from the top of my head.

gautierronan:
Actually scratch that, bb058e983d34d2be4dca4577b28d95400e7d165e works in terms of appearance. We simply need to fix the links, which might be less trivial. Looking if that can be done.

gautierronan:
Ok, with 9527bc398bb267c4fdcbd03079d0601b56129df9, it seems to work as intended. I added the possibility to specify a `namespace` option in the mkdocstrings autodoc call, which will override the assumed link extracted from the code structure.

pierreguilmin:
Ha, that's beautiful! Thanks a bunch!

gautierronan (approved):

PR #824, opened 2024-12-05
Introduce `dq.array` and support `copy` argument 

gautierronan:
@pierreguilmin I started a PR to introduce `dq.array` on top of `dq.asarray` (copy vs no copy), but realise it's a bit of a mess (need to add the `copy` argument everywhere and quite a few error checks). Do you think I should push this further or abandon it? Not sure how useful it is in the end.

### DIFF
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -65,6 +65,10 @@ class DenseQArray(QArray):
         data = self.data.mT
         return DenseQArray(self.dims, data)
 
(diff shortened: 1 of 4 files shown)


pierreguilmin:
> @pierreguilmin I started a PR to introduce dq.array on top of dq.asarray (copy vs no copy), but realise it's a bit of a mess (need to add the copy argument everywhere and quite a few error checks). Do you think I should push this further or abandon it? Not sure how useful it is in the end.

If we don't need it now, let's set it aside 😊

PR #827, opened 2024-12-08
Add a sparsedia implementation for computing matrix exponentials

gautierronan:
Derived from the JAX implementation, which in principle doesn't require a specific matrix layout, but only knowledge of how to perform matrix products. As a consequence, adding the JAX license. 

Left to do:

- [ ] Add tests
- [ ] Debug the vectorization (`jnp.vectorize` won't work on a QArray). @pierreguilmin do you have an idea on how to do that? 
- [ ] Check whether the lineax call can be improved
- [ ] (optional) Add support for upper triangular matrix (`upper_triangular: bool` argument)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -200,5 +200,8 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 
-This project also includes code derived from QuTiP, licensed under the BSD 3-Clause
+This project includes code derived from QuTiP, licensed under the BSD 3-Clause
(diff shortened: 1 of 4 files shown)


gautierronan:
After investigation, the LU solve required for `expm` materializes the sparse array to a dense array, even in lineax. So it doesn't make sense to do complicated processing to end up converting to a dense array at some point in the computation.

As such, I am closing this PR and will be opening a new one to remove the todo comment.

PR #832, opened 2024-12-10
Separate the sparsedia logic from the SparseDIAQArray class

gautierronan:
This PR introduces a new file `sparsedia_primitives.py` which handles all the sparse logic. Within this file, all functions take as input `offsets` and `diags` plus eventual additional arguments, and return another set of `offsets` and `diags`. This completely decouples this logic from the actual `SparseDIAQArray` implementation which then handles `dims`, conversions, etc. In the long term, this should simplify maintenance, especially if we want to replace certain of these primitives with e.g. GPU kernels.

The PR itself is rather lengthy, but there's no new logic. It's only displacing functions around and renaming arguments. As such, you probably don't want to review the precise implementations, but rather functions names, typing, signatures, etc. For the logic, all tests are passing so it should be preserved.

### DIFF
--- a/dynamiqs/qarrays/sparsedia_qarray.py
+++ b/dynamiqs/qarrays/sparsedia_qarray.py
@@ -129,7 +129,7 @@ class SparseDIAQArray(QArray):
             stacklevel=2,
         )
         x = sparsedia_to_array(self.offsets, self.diags)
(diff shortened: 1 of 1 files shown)


pierreguilmin (dismissed):
Beautiful! 😍

PR #834, opened 2024-12-11
Rehaul qarray warnings and checks

gautierronan:
- Introduce stricter operations throughout QArrays (both dense and sparse), e.g. disallow `dq.sigmaz() + 3`, `dq.sigmaz() * dq.sigmax()` or `dq.sigmax()**2`
- Introduce alternative methods `addscalar`, `elmul` and `elpow` to replace these functionnalities
- Move all checks to the base `QArray` class: the only remaining warnings are when a sparse qarray is converted to a dense one
- Rename `other` to `y` throughout SparseQArray to match the naming conventions of the other two files


pierreguilmin (approved):
Amazing 😍  I proposed small modifications on top.

PR #836, opened 2024-12-11
Adapt floquet solver to QArrays

gautierronan:
I suggest reviewing commit by commit.

### DIFF
--- a/dynamiqs/integrators/apis/floquet.py
+++ b/dynamiqs/integrators/apis/floquet.py
@@ -14,7 +14,7 @@ from ...options import Options
 from ...qarrays.qarray import QArrayLike
 from ...result import FloquetResult
 from ...solver import Solver, Tsit5
-from ...time_array import TimeArray
+from ...time_array import CallableTimeArray, TimeArray
 from .._utils import _astimearray, cartesian_vmap, catch_xla_runtime_error
 
 __all__ = ['floquet']
@@ -174,10 +174,13 @@ def _check_floquet_args(
     )
 
     # === check that the Hamiltonian is periodic with the supplied period
-    H = eqx.error_if(
-        H,
-        jnp.logical_not(jnp.isclose(H(0), H(T))),
-        'The Hamiltonian H is not periodic with the supplied period T.',
-    )
+    if not isinstance(H, CallableTimeArray):
+        # disable the check for CallableTimeArray as equinox cannot find an
+        # underlying array to attach the error message to.
+        H = eqx.error_if(
+            H,
+            jnp.logical_not(jnp.isclose(H(0.0)._data, H(T)._data)),  # noqa: SLF001
+            'The Hamiltonian H is not periodic with the supplied period T.',
+        )
 
     return H, T, tsave
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -43,6 +43,10 @@ class DenseQArray(QArray):
         data = self.data.mT
         return DenseQArray(self.dims, data)
 
+    @property
+    def _data(self) -> Array:
+        return self.data
+
     def conj(self) -> QArray:
         data = self.data.conj()
         return DenseQArray(self.dims, data)
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -98,6 +98,10 @@ class DenseQArray(QArray):
         else:
             return DenseQArray(self.dims, data)
 
+    def eig(self) -> tuple[Array, QArray]:
+        evals, evecs = jax.lax.linalg.eig(self.data, compute_left_eigenvectors=False)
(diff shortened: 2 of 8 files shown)


dynamiqs/integrators/apis/floquet.py
```diff
@@ -174,10 +174,13 @@ def _check_floquet_args(
...
-        'The Hamiltonian H is not periodic with the supplied period T.',
-    )
+    if not isinstance(H, CallableTimeArray):
+        # disable the check for CallableTimeArray as equinox cannot find an
+        # underlying array to attach the error message to.
+        H = eqx.error_if(
+            H,
+            jnp.logical_not(jnp.isclose(H(0.0)._data, H(T)._data)),  # noqa: SLF001
+            'The Hamiltonian H is not periodic with the supplied period T.',
+        )
```
dkweiss31, line 184: This is a little strange? Even if `H` has type `CallableTimeArray` shouldn't e.g. `H(0.0)` return a `QArray`? What am I missing here?

gautierronan: It's because of the way the `eqx.error_if` runtime error works. The second argument `jnp.logical_not(jnp.isclose(H(0.0)._data, H(T)._data))` works as expected, but the first argument `H` is the faulty one here. Indeed, I think `eqx.error_if` requires `H` to be a PyTree with at least one underlying array so it can "attach" the error message to this array. But a `TimeCallable` only has a callable as a leaf, and it doesn't find the a corresponding array, leading to an error.

There might be workarounds, but I couldn't find one.

dkweiss31: I see, understood. Indeed I see in the `error_if` docs that the PyTree passed to this must have at least one array.

dkweiss31: Otherwise everything else LGTM :)

gautierronan: Actually maybe I can replace `H` with `H(0.0)` here, not sure, will try...

dkweiss31: With the code as is I don't think that will work, since as I understand it the PyTree on the left gets replaced by the PyTree in the first argument to `error_if`. The following also would not work
```suggestion
    if not isinstance(H, CallableTimeArray):
        # disable the check for CallableTimeArray as equinox cannot find an
        # underlying array to attach the error message to.
        H0 = eqx.error_if(
            H(0.0),
            jnp.logical_not(jnp.isclose(H(0.0)._data, H(T)._data)),  # noqa: SLF001
            'The Hamiltonian H is not periodic with the supplied period T.',
        )
```
since as I understand it, as `H0` is not used later, this code is eliminated under dead-code elimination https://docs.kidger.site/equinox/api/errors/. Seems a little tricky and annoying and any workarounds feel hackish...

dynamiqs/qarrays/dense_qarray.py
```diff
@@ -94,6 +98,10 @@ def squeeze(self, axis: int | tuple[int, ...] | None = None) -> QArray | Array:
         else:
             return DenseQArray(self.dims, data)
 
+    def eig(self) -> tuple[Array, QArray]:
```
pierreguilmin, line 101: I think we should have this as private for now, this will cause endless troubles with superoperators for the same reason we had trouble proposing a proper `dq.eig()` method.

dynamiqs/qarrays/qarray.py
```diff
@@ -210,6 +211,10 @@ def mT(self) -> QArray:
     def ndim(self) -> int:
         return len(self.shape)
 
+    @property
+    def _data(self) -> Array:
+        pass
```
pierreguilmin, line 216: If possible can we avoid this? Or give it a weirder name, `data` is not explicit enough IMO.

dynamiqs/integrators/apis/floquet.py
```diff
@@ -174,10 +174,13 @@ def _check_floquet_args(
...
-        H,
-        jnp.logical_not(jnp.isclose(H(0), H(T))),
-        'The Hamiltonian H is not periodic with the supplied period T.',
-    )
+    if not isinstance(H, CallableTimeArray):
+        # disable the check for CallableTimeArray as equinox cannot find an
+        # underlying array to attach the error message to.
+        H = eqx.error_if(
+            H,
+            jnp.logical_not(jnp.isclose(H(0.0)._data, H(T)._data)),  # noqa: SLF001
```
pierreguilmin, line 182: This calls for a proper method to compare sparsedia qarrays with `jnp.isclose`, user might want to do this too. We can merge this as is and think about it.

pierreguilmin (changes requested):

dynamiqs/integrators/apis/floquet.py
gautierronan: Fixed with 8513423786eac60c1157fac97acd7d84d45ca599.

pierreguilmin (commented):
Great!

PR #841, opened 2024-12-16
Temporary fixes for the CI

pierreguilmin:
Two not very nice fixes, I think the second one on dense qarray matmul will not be needed once I properly implement vectorization. Not sure about the floquet fix @gautierronan, is there a better way>


gautierronan (changes requested):
Not very down to merge this. The first fix should be elsewhere, and the second should be fixed after vectorization as you say. I can take care of the first one later today.

PR #842, opened 2024-12-16
Support vectorised qarrays

pierreguilmin:
The work on this is far from over, but I don't think it should block further the merge of `dev-qarray` to `main`. See https://linear.app/dynamiqs/issue/DYN-256/qarray-next for a list of follow-up tasks.

I advise reading commit by commit.

### DIFF
--- a/dynamiqs/utils/vectorization.py
+++ b/dynamiqs/utils/vectorization.py
@@ -41,18 +41,18 @@ def operator_to_vector(x: QArrayLike) -> QArray:
         >>> A
         Array([[1.+1.j, 2.+2.j],
                [3.+3.j, 4.+4.j]], dtype=complex64)
-
-        # todo: temporary fix
-        # >>> dq.operator_to_vector(A)
-        # Array([[1.+1.j],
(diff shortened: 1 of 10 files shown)


dynamiqs/utils/vectorization.py
```diff
@@ -115,7 +116,8 @@ def spre(x: QArrayLike) -> QArray:
     x = asqarray(x)
     check_shape(x, 'x', '(..., n, n)')
     n = x.shape[-1]
-    return eye(n) & x
+    xpre = eye(n) & x
```
gautierronan, line 119: Not really useful but could have been written as:
```suggestion
    xpre = eye(*x.dims) & x
```
I guess we should mostly introduce a fast `__and__` for $I \otimes X$ and $X \otimes I$ as it's a bit dumb to initialize an identity matrix here.

gautierronan (approved):
LGTM! Some side comments below but I guess they should be handled in separate PRs.

gautierronan (approved):

PR #843, opened 2024-12-16
Uncomment floquet example

pierreguilmin:


gautierronan (approved):

gautierronan:
Oh! Didn't see it was on top of the `fix-ci` branch, maybe we should have merged directly in `dev-qarray` if we end up dismissing the `fix-ci` PR. Sorry!

PR #855, opened 2024-12-17
Add dims ndims and ndiags to timeqarray

abocquet:
https://linear.app/dynamiqs/issue/DYN-256/qarray-next

TODO ?
- [ ] addscalar
- [ ] asdense
- [ ] assparsedia
- [ ] block_until_ready
- [ ] elmul
- [ ] elpow
- [ ] expm
- [ ] isbra
- [ ] isdm
- [ ] isherm
- [ ] isket
- [ ] isop
- [ ] norm
- [ ] powm
- [ ] proj
- [ ] ptrace
- [ ] sinm
- [ ] sum
- [ ] to_jax
- [ ] to_numpy
- [ ] to_qutip
- [ ] tobra
- [ ] todm
- [ ] toket
- [ ] trace
- [ ] unit
- [ ] vectorized


dynamiqs/time_array.py
```diff
@@ -673,6 +723,14 @@ def dtype(self) -> jnp.dtype:
     def shape(self) -> tuple[int, ...]:
         return jnp.broadcast_shapes(*[tarray.shape for tarray in self.timearrays])
 
+    @property
+    def dims(self) -> tuple[int, ...]:
+        return self.timearrays[0].dims
+
+    @property
+    def ndiags(self) -> int:
```
abocquet, line 731: `SummedTimeArray` accepts mixed number of diags so it's ill defined

pierreguilmin:
So I think you're right we probably only want:
- [x] dims
- [x] ndim
- [x] ndiags (as you did is perfect)
- [ ] vectorized

And then in another PR we can make a `__repr__` that is more aligned with the one of `QArray` (except for summed time-qarrays where we need to bake smthg nice).

dynamiqs/time_array.py
gautierronan: Not sure it's ill-defined, since when summing sparse qarrays, the final result will also be a sparse qarray with a well-defined `ndiags`. This is what we should return.

However, it is ill-defined if one sub-array is dense, but it's the same for other `TimeQArray` types.

dynamiqs/time_array.py
```diff
@@ -604,6 +638,14 @@ def dtype(self) -> jnp.dtype:
...
         return self.f.shape
 
+    @property
+    def dims(self) -> tuple[int, ...]:
+        return self.f(0.0).dims
+
+    @property
+    def ndiags(self) -> int:
+        return self.f(0.0).ndiags
+
```
gautierronan, line 648: A `jax.eval_shape(f).dims` or `jax.eval_shape(f).ndiags` should do the trick without evaluating the function here.

gautierronan (changes requested):
@pierreguilmin I don't think we should support `ndiags` actually, unless we support it on the whole `QArray` class. Right now, it is only defined on `SparseDIAQArray`, making it ill-defined for a generic `TimeQArray` class.

pierreguilmin:
> @pierreguilmin I don't think we should support `ndiags` actually, unless we support it on the whole `QArray` class. Right now, it is only defined on `SparseDIAQArray`, making it ill-defined for a generic `TimeQArray` class.

Right, in practice we should only have it appear when the underlying `qarray` is sparse. Funnily linked to our discussion on Slack on `TimeTree`. 😋 I just foresee the `__repr__` for which it would be nice to have `ndiags` for sparse `TimeQArray`s.

gautierronan:
Ok let's keep `ndiags` for now then, and we can remove it later if we realize it's raising too many bugs.

dynamiqs/time_qarray.py
```diff
@@ -673,6 +723,14 @@ def dtype(self) -> jnp.dtype:
     def shape(self) -> tuple[int, ...]:
         return jnp.broadcast_shapes(*[tqarray.shape for tqarray in self.timeqarrays])
 
+    @property
+    def dims(self) -> tuple[int, ...]:
+        return self.timearrays[0].dims
+
+    @property
+    def ndiags(self) -> int:
+        raise NotImplementedError
```
gautierronan, line 732:
```suggestion
        return jax.eval_shape(self.__call__, 0.0).ndiags
```
Doesn't this work?

gautierronan (changes requested):

pierreguilmin:
@abocquet can we add `vectorized` as well?

gautierronan (approved):
LGTM. Thanks!

tests/test_time_qarray.py
```diff
@@ -137,7 +137,7 @@ def test_rmul(self):
 
     def test_add(self):
         # test type `ArrayLike`
-        x = self.x + jnp.ones_like(self.x)
+        x = self.x + constant(jnp.ones_like(self.x))
```
pierreguilmin, line 140: Weird that we need this 🤔

PR #858, opened 2024-12-19
Refactor reformat

abocquet:
Solves vectorization bug where checks were too stricts. 
Instead of making them more flexible with weird conditions for internal use only, I created a `_reshape_unchecked` method that skips all checks and gives full power. This avoids weird conditions such as `shape[-2] == shape[-2] or shape[-2] == shape[-2] * shape[-1] and shape[-1] == 1`

Another option was to move the `operator_to_vector` logic inside the classes but it boiled to adding this `_reshape_unchecked` anyway

### DIFF
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -61,7 +61,7 @@ class DenseQArray(QArray):
         data = self.data.conj()
         return self._replace(data=data)
 
-    def reshape(self, *shape: int) -> QArray:
+    def _reshape_unchecked(self, shape: tuple[int, ...]) -> QArray:
         data = jnp.reshape(self.data, shape)
         return self._replace(data=data)
 
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -61,7 +61,7 @@ class DenseQArray(QArray):
         data = self.data.conj()
         return self._replace(data=data)
 
-    def _reshape_unchecked(self, shape: tuple[int, ...]) -> QArray:
+    def _reshape_unchecked(self, *shape: int) -> QArray:
         data = jnp.reshape(self.data, shape)
         return self._replace(data=data)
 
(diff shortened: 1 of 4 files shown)


dynamiqs/qarrays/qarray.py
```diff
@@ -263,6 +262,19 @@ def reshape(self, *shape: int) -> QArray:
...
+                f'not match current shape dimensions, {self.shape}.'
+            )
+        return self._reshape_unchecked(shape)
+
+    @abstractmethod
+    def _reshape_unchecked(self, shape: tuple[int, ...]) -> QArray:
+        """Does the heavy-lifting for `reshape` but skips all checks.
+        This private method allows for more powerful reshapes that
+        are useful for vectorization.
+        """
```
gautierronan, line 277: Should not have a docstring otherwise it will appear in the documentation page.

dynamiqs/qarrays/dense_qarray.py
```diff
@@ -61,7 +61,7 @@ def conj(self) -> QArray:
         data = self.data.conj()
         return self._replace(data=data)
 
-    def reshape(self, *shape: int) -> QArray:
+    def _reshape_unchecked(self, shape: tuple[int, ...]) -> QArray:
```
gautierronan, line 64:
```suggestion
    def _reshape_unchecked(self, *shape: int) -> QArray:
```
We should keep the same API as `reshape`.

gautierronan (changes requested):
The solution looks good to me. Two small comments below.

dynamiqs/qarrays/dense_qarray.py
abocquet: the goal was to remove the syntaxic sugar in order to deter even more using it

gautierronan: It's already a private and undocumented function. If users still want to use it, let's not give them some batons dans les roues. Plus we should keep some uniformity IMO.

abocquet: I don't agree but I won't argue on this

gautierronan: Thank you!

PR #861, opened 2024-12-19
Add dq.hc

abocquet:


abocquet:
Tests fail in CI but it runs on my machine™️

dynamiqs/__init__.py
```diff
@@ -1,4 +1,5 @@
 from . import dark, plot, random
+from .hermitian_conjugate import hc
```
gautierronan, line 2: For imports throughout the library, we use `from .x import *` and define a `__all__ = [...]` in the file. This structure is needed for documentation afterwards.

dynamiqs/hermitian_conjugate.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class HermitianConjugate:
+    def __radd__(self, y: QArrayLike) -> QArrayLike:
+        if isinstance(y, Qobj):
+            y = asqarray(y)
+
+        if isinstance(y, QArray):
+            return y + y.dag()
+        elif isinstance(y, ArrayLike):
```
gautierronan, line 14: I think the CI complains because of this. I'm not sure but we're not supposed to check types using `jaxtyping` (although that's also what we do at other places in the library). @pierreguilmin an idea?

dynamiqs/hermitian_conjugate.py
```diff
@@ -0,0 +1,23 @@
+from jaxtyping import ArrayLike
+from qutip import Qobj
+
+from .qarrays import QArray, QArrayLike, asqarray
+
+
+class HermitianConjugate:
+    def __radd__(self, y: QArrayLike) -> QArrayLike:
+        if isinstance(y, Qobj):
+            y = asqarray(y)
```
gautierronan, line 10: The input object should not be converted silently (or at least, be converted when it's a Qobj but not when it's an ArrayLike). Either we return a `Qobj` or we don't support `Qobj` at all IMO.

dynamiqs/hermitian_conjugate.py
```diff
@@ -0,0 +1,23 @@
...
+
+        if isinstance(y, QArray):
+            return y + y.dag()
+        elif isinstance(y, ArrayLike):
+            return y + y.mT.conj()
+        else:
```
gautierronan, line 16: The QArrayLike type also includes nested sequences of QArrayLike. So I guess there's something a bit subtle to handle here. Maybe we could only support `dq.hc` with QArrays to begin with ?

dynamiqs/hermitian_conjugate.py
```diff
@@ -0,0 +1,23 @@
...
+            raise TypeError(
+                f'`dq.hq` can only be right-added '
+                f'to a QArrayLike, {y.__class__} given.'
+            )
+
+
+hc = HermitianConjugate()
```
gautierronan, line 23: We need to document this object in the Python API. Not sure whether we should document the class or this instance (might not be possible to document an instance actually?).

dynamiqs/hermitian_conjugate.py
```diff
@@ -0,0 +1,23 @@
...
+        if isinstance(y, Qobj):
+            y = asqarray(y)
+
+        if isinstance(y, QArray):
+            return y + y.dag()
+        elif isinstance(y, ArrayLike):
+            return y + y.mT.conj()
+        else:
+            raise TypeError(
+                f'`dq.hq` can only be right-added '
```
gautierronan, line 18: Doesn't Ruff complain that this should not be an f-string if it's not used?

gautierronan (changes requested):
Thanks for the PR. Below are some comments. Also, we should add support (+tests) for the `TimeArray` type, and not just for `QArray`.

dynamiqs/hermitian_conjugate.py
gautierronan: See also the equinox functions: https://docs.kidger.site/equinox/api/manipulation/#equinox.is_array

pierreguilmin: Just need to do: (for Python 3.9 I think)
```
from typing import get_args
isinstance(y, get_args(ArrayLike))
```

abocquet: Thanks for the tip!

dynamiqs/hermitian_conjugate.py
abocquet: There is no `QArrayLike` here I think so we're good

dynamiqs/hermitian_conjugate.py
abocquet: No

dynamiqs/hermitian_conjugate.py
abocquet: I didn't find how to do it, any idea is welcome :)

dynamiqs/hermitian_conjugate.py
pierreguilmin: Not sure what you mean @abocquet? If the input type is `QArrayLike`, the object is not either a `QArray` or an `ArrayLike`, see https://github.com/dynamiqs/dynamiqs/blob/main/dynamiqs/qarrays/qarray.py#L578-L594 for the complete definition of `QArrayLike`.

dynamiqs/hermitian_conjugate.py
pierreguilmin: Let me try smthg.

pierreguilmin: It is possible by just putting a docstring under @gautierronan.

pierreguilmin:
I've added a few proposals and doc. I don't think we should support something else than qarrays and time-qarrays. Three issues with the doc: (i) I'm not sure you guys enjoy the "Magic" section (I like it 😅), (ii) it doesn't appear in the Python API list and (iii) @gautierronan can we get rid of the "module-attribute" next to the title? (we also have the same annoying markups "abstractmethod" for e.g. time-qarrays).

Screenshot 2024-12-26 at 21 57 52

pierreguilmin:
Do we want to use the trick here and there in the other doc examples where appropriate?

gautierronan:
> I've added a few proposals and doc. I don't think we should support something else than qarrays and time-qarrays. Three issues with the doc: (i) I'm not sure you guys enjoy the "Magic" section (I like it 😅), (ii) it doesn't appear in the Python API list and (iii) @gautierronan can we get rid of the "module-attribute" next to the title? (we also have the same annoying markups "abstractmethod" for e.g. time-qarrays).

(ii) and (iii) are fixed with the last two commits. I indeed don't really like the name "Magic" but it would also be a shame to bury these neat features deep down in an existing section. So all in all, I'm ok to keep it. Maybe with the section name "Magic helpers" ?

pierreguilmin:
> > I've added a few proposals and doc. I don't think we should support something else than qarrays and time-qarrays. Three issues with the doc: (i) I'm not sure you guys enjoy the "Magic" section (I like it 😅), (ii) it doesn't appear in the Python API list and (iii) @gautierronan can we get rid of the "module-attribute" next to the title? (we also have the same annoying markups "abstractmethod" for e.g. time-qarrays).
> 
> (ii) and (iii) are fixed with the last two commits. I indeed don't really like the name "Magic" but it would also be a shame to bury these neat features deep down in an existing section. So all in all, I'm ok to keep it. Maybe with the section name "Magic helpers" ?

Great thanks a lot! I changed to "Magic helpers".

pierreguilmin (dismissed):

PR #862, opened 2024-12-20
Fix the Floquet periodicity check using `eqx.tree_at`

gautierronan:
Follow up on https://github.com/dynamiqs/dynamiqs/pull/836 to use `eqx.tree_equal` for the periodicity check, which is much more neat than my previous `_underlying_array` thing. 

The only issue is that I have to hardcode a tolerence for the check. In principe, we could add it to `dq.Options`, but I'm a bit annoyed to have to add such a global option for only one little check. At some point, we should think of how to properly handle such cases.


pierreguilmin (approved):
Much better! 😍

PR #866, opened 2024-12-24
Introduce a class property `out_axes` for Results

gautierronan:
Small little addition that makes life easier when vmapping externally over dynamiqs integration functions. Instead of trying to figure out the correct `out_axes`, we can now just call `jax.vmap(...,  out_axes=dq.MESolveResult.out_axes)`.


pierreguilmin (approved):
Nice!

PR #881, opened 2024-12-27
Add indicative APIs for stochastic solvers

pierreguilmin:
On top of #894.

Adds four new APIs:
- `dq.jssesolve()` for the jump stochastic Schrödinger equation
- `dq.dssesolve()` for the diffusive stochastic Schrödinger equation
- `dq.jsmesolve()` for the jump stochastic master equation
- `dq.dsmesolve()` for the diffusive stochastic master equation


pierreguilmin:
A few remarks:
- Once we have options specific to solvers I propose adding `nmaxclick` to `jsmesolve` options.
- For fixed step size solvers, the `clicktimes` will indicate the interval during which a click occured $[t, t+dt[$ rather than the exact `clicktimes`.
- Once approved, I'll write the corresponding `jssesolve` and `dssesolve` APIs, which will be fairly similar (with no $\theta_k$ or $\eta_k$).

pierreguilmin:
For the jump SSE equations see the _QUANTUM MEASUREMENT AND CONTROL_ book:
- jump SSE: (4.29) + errata ($-iH$ is outside the sum)
- diffusive SSE: (4.76)

gautierronan (dismissed):
Thanks a lot for the PR, it's great. I committed on top with a few changes. I'll let you look at them and see if you agree.

Some comments:
- We should add a reference within the docstring for all these equations, probably the Wiseman book, ideally with the corresponding chapter or equation.
- I'll let you check, but I believe there was a missing factor 1/4 in the DSSE equation (according to Eq. (18.74) from Quantum and Atom Optics by Daniel Steck, at least)
- In both SSEs, I replaced brakets with expectation values to ease readability of the equation
- There's a small rendering bug in the result API admonition, in which the Attributes does not appear on a new line. I was not able to fix it.

I'm happy to merge as is. If we can fix number (1) and number (4) above, I'll be even happier :D

dynamiqs/integrators/apis/dssesolve.py
```diff
@@ -28,18 +28,21 @@ def dssesolve(
         This function has not been implemented yet. The following API is indicative
         of the planned implementation.
 
+    In quantum optics, the diffusive SSE describes the evolution of a quantum
+    system under homodyne or heterodyne detection with unit detection efficiency.
     This function computes the evolution of the state vector $\ket{\psi(t)}$ at time $t$,
     starting from an initial state $\ket{\psi_0}$ according to the diffusive SSE in Itô
-    form (with $\hbar=1$ and where time is implicit(1))
+    form ($\hbar=1$, time is implicit(1))
```
pierreguilmin, line 35: Why not, I propose to do this in another PR for all solvers.

dynamiqs/integrators/apis/dssesolve.py
```diff
@@ -28,18 +28,21 @@ def dssesolve(
...
     $$
         \begin{split}
-            \dd\ket\psi = \Bigg[
+            \dd\\!\ket\psi = \Bigg[
                 &-iH \dt
                 -\frac12 \sum_{k=1}^N \left(
-                    L_k^\dag L_k - \braket{\psi| L_k + L_k^\dag |\psi} L_k + \braket{\psi| L_k + L_k^\dag |\psi}^2 I
+                    L_k^\dag L_k - \braket{L_k + L_k^\dag} L_k
+                    + \braket{L_k + L_k^\dag }^2 \\! / 4
```
pierreguilmin, line 42: I think we should put the back `I`.

dynamiqs/integrators/apis/dssesolve.py
```diff
@@ -28,18 +28,21 @@ def dssesolve(
...
                 \right) \dt \\\\
                 &+ \sum_{k=1}^N \left(
-                    L_k - \frac12 \braket{\psi| L_k + L_k^\dag |\psi}
+                    L_k - \braket{L_k + L_k^\dag} \\! / 2
```
pierreguilmin, line 45: I find it weird to inline some fractions and not some other. I suggest not inlining any fraction.

dynamiqs/integrators/apis/jsmesolve.py
```diff
@@ -64,14 +65,9 @@ def jsmesolve(
...
-    Note-: Jump vs. diffusive SME
-        In quantum optics the _jump_ SME corresponds to photon counting schemes, as
-        opposed to the _diffusive_ SME which corresponds to homodyne or heterodyne
-        detection schemes.
-
     The continuous-time measurements are defined by the point processes $\dd N_k$. The
-    solver returns the times at which the detector clicked:
-    $I_k = \{t \text{ s.t. } \dd N_k(t)=1\}$.
+    solver returns the times at which the detector clicked,
+    $I_k = \{t \in [t_0, t_\text{end}[ \,|\, \dd N_k(t)=1\}$.
```
pierreguilmin, line 70: `t_0`  and `t_\text{end}` are not defined.

dynamiqs/integrators/apis/jssesolve.py
```diff
@@ -28,18 +28,20 @@ def jssesolve(
...
         \begin{split}
             \dd\ket\psi = \Bigg[
                 &-iH \dt
                 - \frac12 \sum_{k=1}^N \left(
-                    L_k^\dag L_k - \braket{\psi| L_k^\dag L_k |\psi}
+                    L_k^\dag L_k - \braket{L_k^\dag L_k}
                 \right) \dt \\\\
                 &+ \sum_{k=1}^N \left(
-                    \frac{L_k}{\sqrt{\braket{\psi| L_k^\dag L_k |\psi}}} - I
+                    \frac{L_k}{\sqrt{\braket{L_k^\dag L_k}}} - 1
```
pierreguilmin, line 44: The `I` is important.

dynamiqs/integrators/apis/dssesolve.py
```diff
@@ -0,0 +1,211 @@
...
+        of the planned implementation.
+
+    In quantum optics, the diffusive SSE describes the evolution of a quantum
+    system under homodyne or heterodyne detection with unit detection efficiency.
+    This function computes the evolution of the state vector $\ket{\psi(t)}$ at time $t$,
+    starting from an initial state $\ket{\psi_0}$ according to the diffusive SSE in Itô
+    form ($\hbar=1$, time is implicit(1))
+    $$
+        \begin{split}
+            \dd\\!\ket\psi = \Bigg[
```
pierreguilmin, line 38: We should either do the `\\!` everywhere in the doc or nowhere. If we choose to do it I suggest doing it in another PR.

gautierronan: What do you mean? There was some very wierd spacing issue in this equation, which is not the case in other equations.

dynamiqs/integrators/apis/dssesolve.py
gautierronan: Ok, I prefer this succint formulation on my end.

dynamiqs/integrators/apis/dssesolve.py
gautierronan: Why? There is no ambiguity here, I'm very much in favor of keeping the equation as simple and readable as possible.

dynamiqs/integrators/apis/dssesolve.py
gautierronan: Ok, this one is a personal preference (I think it looks better), but I'm ok to remove.

dynamiqs/integrators/apis/jsmesolve.py
gautierronan: Sure, but it's quite obvious, no? I don't think we need to be 100% rigorous everywhere.

dynamiqs/integrators/apis/jssesolve.py
gautierronan: Why is it important? It's not defined anywhere, plus it is standard notation to replace identity with a `1`.

gautierronan:
> I disagree with the change introducing "In quantum optics ..." as a main part of the dosctring. SSEs and SMEs are not restricted to quantum optics, and used in many other fields. Maybe we can juste remove this remark altogether? It was mostly here to guide QuTiP users, bc QuTiP defines "homodyne", "photocounting" and "heterodyne" options.

I don't really have a strong opinion on the formulation, but we should definitely have a short (one sentence) introduction on the meaning of these equations. We are not only targetting theoretical physicists that are very knowledgeable about SMEs (0.01% of physicists), but also a more general audience that we want to educate on these topics (experimentalists, but also theoreticians and students). 

> That's great! Should we define what ⟨ ∙ ⟩ means in this context for clarity?

I think the expectation value notation is standard enough for it not to be explicited. My overall goal is to keep the docstrings understandable without them being bloated with rigourous notations.

dynamiqs/integrators/apis/dssesolve.py
pierreguilmin: Many people (including myself and Gaspard for example) bug when mixing matrices with numbers (much like when you can do "a + 2" in QuTip), it's "inhomogeneous". I think it's a notational shortcut that doesn't gain much space and makes the equation harder to understand. But if you strongly prefer this it's fine.

dynamiqs/integrators/apis/jsmesolve.py
pierreguilmin: Yes ok 👍

dynamiqs/integrators/apis/dssesolve.py
pierreguilmin: Ha I don't have the issues on my end. 🤔 I thought you just wanted to have the straight "d" closer to the ket symbol.

pierreguilmin:
> I don't really have a strong opinion on the formulation, but we should definitely have a short (one sentence) introduction on the meaning of these equations. We are not only targetting theoretical physicists that are very knowledgeable about SMEs (0.01% of physicists), but also a more general audience that we want to educate on these topics (experimentalists, but also theoreticians and students).

Right, I think that's the goal of the tutorials, the API should be rather short and strict. I'm a bit afraid that people doing condensed matter (and detecting jumps of fermions daily) avoid using these equations (we're already using the quantum optics notations). That's why I added the remark as a side note.

dynamiqs/integrators/apis/dssesolve.py
gautierronan: I don't quite agree. The equation is perfectly valid here, you're multiplying a constant with a ket, there is no need to introduce an identity operator in between. IMO this additional identity is making the equation more complicated and adds nothing in terms of understanding.

Your example is quite relevant, we always write `a^2 - alpha^2` and not `a^2 - alpha^2 * I` in physics equations, otherwise we start having identity operators all around the place.

dynamiqs/integrators/apis/dssesolve.py
gautierronan: Yes that's exactly what I wanted to do. The "d" looked very far from the ket symbol if I didn't add the `\\!`, for some reason?

dynamiqs/integrators/apis/dssesolve.py
pierreguilmin: Ok, as you prefer :)

dynamiqs/integrators/apis/dssesolve.py
pierreguilmin: Ok, I'm not sure it's needed anymore then.

gautierronan:
I get what you mean. I don't think we should specifically talk about quantum optics though. But having a short introduction to the equation in the docstring is very useful.

Also, I'm not convinced by your argument of the tutorial: the tutorial will probably be written a long time ago from now (if ever), and even if, the Python API should be welcoming for beginners (which, again, are the majority of users regarding stochastic equations).

gautierronan (dismissed):

pierreguilmin:
Great, thanks a lot for the additions @gautierronan!! 😍

PR #882, opened 2024-12-28
Remove abstraction hell

pierreguilmin:
The logic of the solvers remain unchanged, this is simply a more efficient way to manage all the boilerplate.

This was written foreseeing #881.


gautierronan (dismissed):
Great PR! I made a few changes on top, well separated into commits. Let me know what you think. My overarching goal is to be a bit stricter on the way to define Integrators while introducing minimal boilerplate, such that we may write a straightforward tutorial afterwards for solving custom ODEs.

pierreguilmin:
Thanks! I disagree with some of the changes:
- 477da9f: ✅
- 5416ba6: ✅ but meh (a bit weird bc the solver is `floquet` and not `sefloquet`, but why not).
- 960daa1: ❌ mixins should not define interfaces. We can have a solver for `sesolve` which does not use the `SolveSaveMixin`, and thus would not inherit from it, and thus would be missing the `self.Es`. Interfaces define what are an object attributes and methods, mixins are just helpers (you can read online for the difference 😉).
- 1f99403: ❌ we need the `TimeInterface` to use `self.discontinuity_ts` in `DiffraxIntegrator` and `ExpmIntegrator`. I think it makes sense to have it at this level.
- 18ad6bd: ✅ good catch with the missing `PropagatorSaveMixin` for `SEPropagatorDiffraxIntegrator`!
- 059de6a: ✅
- 8faf921: ❌ no, the previous was better, `MESolveRouchon1Integrator` is a subclass of `MESolveDiffraxIntegrator`, no need to redefine it.
- 1105390: ✅ less general but indeed simpler for now, we'll see if we need more later. 👌

PR #886, opened 2024-12-29
Support start and end time for time-qarrays

pierreguilmin:
This was tricky, but it ultimately helped to smooth out some rough edges in the current implementation. I'm happy to get some feedback before finishing the implementation with the todos listed below, I recommend reading commit by commit.

Left todo:
- [x] benchmark new implementation
- [x] add `tstart` and `tend` to `discontinuity_ts`
- [x] document (new constructors optional arguments, new `TimeQArray` attributes and add a section on composing pulses in the tutorial)
- [x] verify that we can jit/grad on `tstart`/`tend` (this could be used for example to fit a delay in an experiment)

Related to https://linear.app/dynamiqs/issue/DYN-147/improve-timearray-support.

### DIFF
--- a/dynamiqs/time_qarray.py
+++ b/dynamiqs/time_qarray.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import functools as ft
 from abc import abstractmethod
-from typing import get_args
 
 import equinox as eqx
 import jax
@@ -265,10 +264,7 @@ class TimeQArray(eqx.Module):
     # Subclasses should implement:
     # - the properties: dtype, shape, dims, ndiags, vectorized, layout, mT, in_axes,
     #                   discontinuity_ts
-    # - the methods: reshape, broadcast_to, conj, __call__, __mul__, __add__
-
-    # Note that a subclass implementation of `__add__` only need to support addition
-    # with `QArray`, `ConstantTimeQArray` and the subclass type itself.
+    # - the methods: reshape, broadcast_to, conj, __call__, __mul__
 
     def _replace(self, **kwargs) -> TimeQArray:
         return type(self)(**kwargs)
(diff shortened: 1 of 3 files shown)


dynamiqs/time_qarray.py
```diff
@@ -679,52 +723,49 @@ def layout(self) -> Layout:
...
+    def _replace(self, timeqarrays: list[TimeQArray] | None = None) -> TimeQArray:
+        if timeqarrays is None:
+            timeqarrays = self.timeqarrays
+        return super()._replace(timeqarrays=timeqarrays)
+
+    def __init__(self, tstart: float, tend: float, timeqarrays: list[TimeQArray]):
+        super().__init__(tstart, tend)
+
+        # dirty way to bypass the check for the implementation of `in_axes`
+        if tstart is not None and tend is not None:
```
gautierronan, line 768: Can't we put this check/shape broadcast inside a `__post_init__` (or maybe `__check_init__`) for this? I don't quite understand the problem of `in_axes` I think.

dynamiqs/time_qarray.py
```diff
@@ -393,6 +413,12 @@ def __call__(self, t: ScalarLike) -> QArray:
         Returns:
             Qarray evaluated at time $t$.
         """
+        return self._indicator(t) * self._call(t)
```
gautierronan, line 416: I believe it should be a tiny-bit faster if you return `self._call(t)` or `dq.zero(*self.dims)` depending on the condition, instead of multiplying by 1 or 0. You save both a multiplication and an evaluation of the array if out-of-time-bounds. Plus we won't have surprises regarding dtype promotions due to the multiplication (or stuff like that).

gautierronan (commented):
It looks good to me. I have added two suggestions below, but I'd be happy to merge as is.

pierreguilmin:
After discussion: let's do an API with `.trim()` and `.shift()`. I'll work on this later.

pierreguilmin:
This will be continued in another PR.

PR #887, opened 2024-12-31
Modify options to add support for local options

gautierronan:
Use tabs to document solver-specific options.

<details>
  <summary>Outdated PR description</summary>
Unfortunately it is not possible to combine `pymdownx` tabs with the mkdocstrings parser, `griffe`. So our idea of documenting the `Options` class with tabs for each quantum solver does not work, unless we use HTML directly in the docstring.

As such, I simply went with a
```
options: Generic options (supported: `save_states`, `progress_meter`, ...).
```
inside each quantum solver docstring, like we do for the `solver` attribute.

IMO this is very unsatisfying and it pushes me back towards the idea of having distinct classes (`dq.MESolveOptions`, ...). Down for your opinions. 
</details>

### DIFF
--- a/docs/generate_python_api.py
+++ b/docs/generate_python_api.py
@@ -16,7 +16,6 @@ doc_sections = {
     'time_qarray': (['dynamiqs/time_qarray.py'], 'dq'),
     'solver': (['dynamiqs/solver.py'], 'dq.solver'),
     'gradient': (['dynamiqs/gradient.py'], 'dq.gradient'),
-    'options': (['dynamiqs/options.py'], 'dq'),
     'result': (['dynamiqs/result.py'], 'dq'),
     'hermitian_conjugate': (['dynamiqs/hermitian_conjugate.py'], 'dq'),
     'utils/operators': (['dynamiqs/utils/operators.py'], 'dq'),
--- a/docs/python_api/options/Options.md
+++ b/docs/python_api/options/Options.md
@@ -1,0 +1,4 @@
+::: dynamiqs.options.Options
+    options:
+        namespace: dq
+        separate_signature: False
--- a/docs/templates/python/material/class.html.jinja
+++ b/docs/templates/python/material/class.html.jinja
@@ -9,16 +9,20 @@
     <code class="doc-symbol doc-symbol-heading doc-symbol-class">
     </code>
   {% endif %}
+  {# same heading regardless of the config.separate_signature value #}
+  <span class="doc doc-object-name doc-class-name">{{ class_full_name }}</span>
+  {#
   {% if config.separate_signature %}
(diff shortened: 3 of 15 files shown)


pierreguilmin (dismissed):
Nice! A few commits on top, happy to merge as is.

Wdyt of this version rather than the current one? (with the signature in the tabs, I think it would be much clearer)
Screenshot 2025-01-03 at 11 42 51

gautierronan:
Thanks for the changes. I disagree with the following two commits:
- https://github.com/dynamiqs/dynamiqs/pull/887/commits/0f36327b3288ad525cc0d73f0e2b1459d194a0d7: I agree the change is useless in its current form, but it's a safety net for future changes, and it makes it clear that the options passed to `propagator` are not necessarely the same as the options passed to `floquet`. Plus it's instantaneous in runtime, so I'd be in favor of keeping it.
- https://github.com/dynamiqs/dynamiqs/pull/887/commits/96a1762e1b6b55acfa8dfa346223a3cb813c6aa0: the formulation is simply not correct, you can pass in `dq.Options(save_propagators=True)` to `mesolve` and the function will run as expected.

> Wdyt of this version rather than the current one? (with the signature in the tabs, I think it would be clearer)

Not sure. It's not correct in terms of API since the object accepts any option as input, but I agree it may be easier to understand.

pierreguilmin:
> https://github.com/dynamiqs/dynamiqs/commit/0f36327b3288ad525cc0d73f0e2b1459d194a0d7: I agree the change is useless in its current form, but it's a safety net for future changes, and it makes it clear that the options passed to propagator are not necessarely the same as the options passed to floquet. Plus it's instantaneous in runtime, so I'd be in favor of keeping it.

Ok, I disagree but don't have a strong opinion, we can keep it. :) You can revert the commit. Can you use an `eqx.tree_at` for the replace rather than the dataclasses one?

Regarding these two comments:
> https://github.com/dynamiqs/dynamiqs/commit/96a1762e1b6b55acfa8dfa346223a3cb813c6aa0: the formulation is simply not correct, you can pass in dq.Options(save_propagators=True) to mesolve and the function will run as expected.
> Not sure. It's not correct in terms of API since the object accepts any option as input, but I agree it may be easier to understand.

I think `dq.Options` is just a glorified dictionnary, and the goal of this doc page is to be very clear that you should only use the options specific to the chosen solver. So I don't think we need to be super precise nor exact on what's happening under the hood, rather precise on what we expect. If someone uses `dq.Options(save_states=True, save_propagators=True)`, it's a bad behaviour even if it's not catched by our exception.

pierreguilmin (approved):
We can do cleaner, but I think we should merge this for now to move fwd. 😊

PR #892, opened 2025-01-13
Add runtime check for non-hermitian density matrix input

gautierronan:
cc @RemiRousseau 

Note that a better solution would be to use a different `vector_field` depending on `rho0` hermitian or not, but a bit harder to implement and maintain. Also note that I make this runtime check for all `mesolve` inputs, but, in principle, we support non-hermitian inputs for the `Expm` solver.


pierreguilmin (changes requested):
Great! Two minor comments.

I'm a bit scared by the default `atol` of `isherm` at `1e-8`. I propose switching to `1e-5` or `1e-6`.

pierreguilmin (approved):
Ha that's even better! 😍

PR #893, opened 2025-01-13
Fix Ruff A005 by making random a folder instead of a module

gautierronan:
Fixes (new with Ruff 0.9.1)
```
>>> ruff check
dynamiqs/random.py:1:1: A005 Module `random` shadows a Python standard-library module
Found 1 error.
```


pierreguilmin (commented):
Why not just ignoring this specific ruff warning for this file?

gautierronan:
> Why not just ignoring this specific ruff warning for this file?

Because if Ruff complains it's probably a bad habit, and indeed @abocquet had issues with it. Plus this solution is very simple so I prefer not ignoring ruff warnings when we can.

pierreguilmin (approved):

PR #894, opened 2025-01-15
Merge documentation of options, results and quantum solvers

pierreguilmin:


gautierronan (commented):
I very much like your proposal. I don't have much to add but:
- https://github.com/dynamiqs/dynamiqs/pull/894/commits/d237d6561cc9c91742451e4de2697074a4bfcccf: small nits regarding formatting
- https://github.com/dynamiqs/dynamiqs/pull/894/commits/3314a879357c5e97a2ea8e8599f09cceae2e03dc: proposal to add a signature to the results admonition, which looks a bit cleaner IMO, but feel free to remove.

Other than this, I'm good to propagate to other functions!

gautierronan (approved):
In the end, I'm not completely satisfied in what we're converging to. I think these pages are a bit too complicated, which makes it difficult to access for beginners.

Still, I don't have time (yet) to propose an alternative, and don't want to be blocking, so I'm happy to merge this PR as is. I will probably open a PR in the next few weeks with a proposal, if that's ok.

PR #898, opened 2025-01-27
Implement the operator sign utility

gautierronan:
This PR implements the operator sign utility function, typically for computing the sign of quadrature, but generally of a hermitian operator. For instance, it is useful for the GKP code to compute $\mathrm{sign}(\mathrm{cos}(2 \pi x))$, or for the cat code to compute $\mathrm{sign}(x)$.

Because it is a duplicate of the `dq.dark.quadrature_sign`, I propose removing the dark namespace altogether, which has not seen much usage for now.


dynamiqs/qarrays/utils.py
```diff
@@ -288,6 +287,8 @@ def to_qutip(x: QArrayLike, dims: tuple[int, ...] | None = None) -> Qobj | list[
          [0. 0. 0. 0. 1. 0.]
          [0. 0. 0. 0. 0. 1.]]
     """  # noqa: E501
+    from .._checks import check_shape
+
```
gautierronan, line 291: For some reason, removing the dark namespace required this to avoid a circular import. The fact that there is a circular import makes sense, but I'm not sure why the dark namespace made it work.

gautierronan:
For some reason, CI runs locally but not on github. Fixed with https://github.com/dynamiqs/dynamiqs/pull/898/commits/9640bf118b57124d52ee246fc5f2381b4eb8c125 (sign inside QArray print). Do you have an idea @pierreguilmin ?

abocquet (dismissed):

dynamiqs/utils/general.py
```diff
@@ -171,6 +172,53 @@ def sinm(x: QArrayLike) -> QArray:
...
+
+    Returns:
+        _(qarray of shape (..., n, n))_ Operator sign function of `x`.
+
+    Note:
+        This function uses [`jnp.linalg.eigh()`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.eigh.html)
+        to compute the sign of a matrix $A$:
+        $$
+            \mathrm{sign}(A) = U \mathrm{sign}(D) U^\dagger
+        $$
```
pierreguilmin, line 192: Can we put the formula in the main description?

dynamiqs/utils/general.py
```diff
@@ -171,6 +172,53 @@ def sinm(x: QArrayLike) -> QArray:
...
+         [ 0.888+0.j -0.   +0.j  0.46 +0.j -0.   +0.j]
+         [ 0.   +0.j  0.46 +0.j -0.   +0.j  0.888+0.j]
+         [-0.46 +0.j -0.   +0.j  0.888+0.j -0.   +0.j]]
+    """
+    x = asqarray(x)
+    x = check_hermitian(x, 'x')
+    L, Q = x.asdense()._eigh()
+    sign_L = jnp.diag(jnp.sign(L))
+    array = Q @ sign_L @ dag(Q)
+    return asqarray(array)
```
pierreguilmin, line 219: Dims of the original qarray will not propagate correctly.

dynamiqs/utils/general.py
```diff
@@ -171,6 +172,53 @@ def sinm(x: QArrayLike) -> QArray:
     return -0.5j * (expm(1j * x) - expm(-1j * x))
 
 
+def signm(x: QArrayLike) -> QArray:
+    r"""Returns the operator sign function of a hermitian qarray.
```
pierreguilmin, line 176: Is "operator sign function" the correct name mathematically? 🤔

dynamiqs/utils/general.py
```diff
@@ -171,6 +172,53 @@ def sinm(x: QArrayLike) -> QArray:
...
+
+    The operator sign is generally dense, and is different from the element-wise
```
pierreguilmin, line 178: We can maybe put this in a "Note:" admonition.

pierreguilmin (changes requested):
Great! Minor comments.

dynamiqs/qarrays/sparsedia_qarray.py
```diff
@@ -203,8 +203,10 @@ def _eigvalsh(self) -> Array:
     def devices(self) -> set[jax.Device]:
         raise NotImplementedError
 
-    def isherm(self) -> bool:
-        raise NotImplementedError
+    def isherm(self, rtol: float = 1e-5, atol: float = 1e-8) -> bool:
+        # TODO: Improve this by using a direct QArray comparison function, once it is
+        # implemented. This will avoid materalizing the dense matrix.
+        return jnp.allclose(self.to_jax(), self.dag().to_jax(), rtol=rtol, atol=atol)
```
pierreguilmin, line 209:
```suggestion
        return self.asdense().isherm(rtol=rtol, atol=atol)
```

dynamiqs/utils/general.py
gautierronan: I'm not sure. How would you call it? I don't want to call it "sign function" only because it's ambiguous with the element-wise sign.

dynamiqs/utils/general.py
gautierronan: It's not really a definition, hence why I put it there, but ok.

dynamiqs/utils/general.py
gautierronan: Good catch

pierreguilmin (approved):
Great!

PR #899, opened 2025-01-27
Fix wrongly implemented __truediv__ in QArrays

gautierronan:
```python
import dynamiqs as dq
import jax.numpy as jnp

dq.sigmaz() / jnp.sqrt(2) # this should work
jnp.sqrt(2) / dq.sigmaz() # this should not work
```

### DIFF
--- a/dynamiqs/qarrays/qarray.py
+++ b/dynamiqs/qarrays/qarray.py
@@ -454,12 +454,12 @@ class QArray(eqx.Module):
     def __rmul__(self, y: QArrayLike) -> QArray:
         return self * y
 
-    def __truediv__(self, y: QArrayLike) -> QArray:
+    def __truediv__(self, y: ArrayLike) -> QArray:
         return self * (1 / y)
 
     def __rtruediv__(self, y: QArrayLike) -> QArray:
         raise NotImplementedError(
-            'Division by a qarray with the `/` operator is not supported. '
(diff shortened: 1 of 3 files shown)


abocquet (approved):

dynamiqs/qarrays/qarray.py
```diff
@@ -454,12 +454,13 @@ def __mul__(self, y: ArrayLike) -> QArray:
     def __rmul__(self, y: QArrayLike) -> QArray:
         return self * y
 
-    @abstractmethod
     def __truediv__(self, y: QArrayLike) -> QArray:
```
pierreguilmin, line 457:
```suggestion
    def __truediv__(self, y: ArrayLike) -> QArray:
```

dynamiqs/qarrays/qarray.py
```diff
@@ -454,12 +454,13 @@ def __mul__(self, y: ArrayLike) -> QArray:
...
-        pass
+        return self * (1 / y)
 
     def __rtruediv__(self, y: QArrayLike) -> QArray:
-        return self * 1 / y
+        raise NotImplementedError(
+            'Division by a qarray with the `/` operator is not supported. '
```
pierreguilmin, line 462:
```suggestion
            'Division by a qarray with the `/` operator is not supported.'
```

pierreguilmin (approved):
Great!

PR #904, opened 2025-02-03
Add `normalize` option to Rouchon solver

pierreguilmin:


dynamiqs/integrators/core/rouchon_integrator.py
```diff
@@ -57,35 +58,65 @@ class MESolveRouchon1Integrator(MESolveDiffraxIntegrator):
...
+                # solve Td @ x = rho => x = Td^{-1} @ rho
+                rho = jax.scipy.linalg.solve_triangular(T, rho, lower=True, trans='C')
+                # solve x @ T = rho => x = rho @ T^{-1}, but we need to reverse the
+                # system for jax.scipy.linalg.solve_triangular:
+                # x @ T = rho => Td @ xd = rhod
+                #            => xd = Td^{-1} @ rhod
+                #            => x = rho @ T^{-1}
+                rhod = rho.conj().T
+                xd = jax.scipy.linalg.solve_triangular(T, rhod, lower=True, trans='C')
+                rho = xd.conj().T
```
gautierronan, line 117: Instead of this, you can use [`jax.lax.linalg.triangular_solve`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.linalg.triangular_solve.html) which has the `transpose_a: bool`, `conjugate_a: bool` and `left_side: bool` options. Same thing for the previous `solve_triangular`, although it should be the same.

gautierronan (changes requested):
I looked at the Kraus map in detail, everything LGTM apart from the small comment below which might speed up a little bit.

pierreguilmin:
The build is failing for some other reason, it seems easy to fix but I'm going to bed. If you have time to look at it in the meantime that would be perfect, otherwise I'll do it tmo. 😊

pierreguilmin:
I'm down for a re-thorough review of 7ef9b76f089b8ab0d7ff7d6eeb7244204e679499.

gautierronan (approved):
LGTM!

PR #906, opened 2025-02-04
Fix `signm` example to avoid the float precision 0 vs -0 sign doctest failure

pierreguilmin:


pierreguilmin:
@gautierronan I don't have a better option for your sign problem 🤷‍♂️ It's the same for me. I'm guessing a subtle hardware-dependent float precision dependency, it's not the first time I change an example to avoid this.

gautierronan (approved):
Ok weird! Thanks for checking it out

PR #909, opened 2025-02-05
Add `dq.plot.xyz()` to plot Pauli operators expectation values for a qubit

pierreguilmin:
I use this all the time. 😇

Screenshot 2025-02-05 at 11 59 16

### DIFF
--- a/dynamiqs/plot/qubit.py
+++ b/dynamiqs/plot/qubit.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+
+import jax.numpy as jnp
+from jax.typing import ArrayLike
+from matplotlib.axes import Axes
+
+from .._checks import check_shape
+from ..qarrays.qarray import QArrayLike
+from ..qarrays.utils import to_jax
+from ..utils.general import expect
+from ..utils.operators import sigmax, sigmay, sigmaz
+from .utils import optional_ax
+
+__all__ = ['xyz']
+
+
+@optional_ax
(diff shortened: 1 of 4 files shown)


dynamiqs/plot/qubit.py
```diff
@@ -0,0 +1,57 @@
...
+@optional_ax
+def xyz(
+    states: QArrayLike,
+    *,
+    ax: Axes | None = None,
+    times: ArrayLike | None = None,
+    hlines: bool = True,
+):
+    r"""Plot the expectation values of the Pauli operators $\sigma_x$, $\sigma_y$ and
+    $\sigma_z$ for a qubit as functions of time.
```
gautierronan, line 26:
```suggestion
    r"""Plot the expectation value of the Pauli operators of a qubit $\sigma_x$,
    $\sigma_y$ and $\sigma_z$ as a function of time.
```

gautierronan (changes requested):
Minor comment on the phrasing of the docstring.

I realize we never put axes labels on our plots. I think we should add sane defaults, e.g. "Time" (for the x-axis) and "Expectation value" (for the y-axis) here.

pierreguilmin:
> I realize we never put axes labels on our plots. I think we should add sane defaults, e.g. "Time" (for the x-axis) and "Expectation value" (for the y-axis) here.

Yes! Let's fix this for all plots in another PR 👍

PR #922, opened 2025-02-25
Various nits and cleanup

gautierronan:
Two main changes:

- Harmonize the internal use of private/public methods. A private method (marked as `_method`) is only used inside the scope where it is defined (i.e. inside its own file or its own class). A public method (marked as `method`) can be used anywhere in the code. It does not need to be in the public API to be public. Of course, this is purely a naming convention that I propose, but it was a bit random for now.
- Move the implementation of `to_jax` and `to_numpy` directly in `qarray.py`. This is similar to what we do with `isqarraylike`, and avoids having both `_to_jax` and `to_jax` functions that do the same thing.

### DIFF
--- a/dynamiqs/time_qarray.py
+++ b/dynamiqs/time_qarray.py
@@ -232,10 +232,6 @@ def timecallable(
     return CallableTimeQArray(f, discontinuity_ts)
 
 
-class Shape(tuple):
-    """Helper class to help with Pytree handling."""
-
-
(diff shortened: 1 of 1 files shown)


pierreguilmin (approved):
Beautiful, thanks a lot! I've added the new conventions to the backlog in https://linear.app/dynamiqs/issue/DYN-254/improve-contributingmd.

PR #924, opened 2025-02-26
Fix matrix multiplication of dense @ sparse

gautierronan:
@pierreguilmin Do you have an idea for a cleaner dispatching inside `DenseQArray.__matmul__`, rather than with the delayed import of `SparseDIAQArray`?

### DIFF
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -169,10 +169,13 @@ class DenseQArray(QArray):
         return self._replace(data=data)
 
     def __matmul__(self, y: QArrayLike) -> QArray | Array:
-        super().__matmul__(y)
+        from .sparsedia_qarray import SparseDIAQArray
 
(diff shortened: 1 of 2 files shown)


pierreguilmin:
Well we could do this with `__array_priority__`, and only for `__matmul__`, that would be a clean way.

gautierronan:
> Well we could do this with `__array_priority__`, and only for `__matmul__`, that would be a clean way.

Ok, I don't think we should rely on `__array_priority__` too much. It doesn't seem well adopted/stable enough.

pierreguilmin:
> > Well we could do this with `__array_priority__`, and only for `__matmul__`, that would be a clean way.
> 
> 
> 
> Ok, I don't think we should rely on `__array_priority__` too much. It doesn't seem well adopted/stable enough.

It's just an easy way for us to avoid the sparse import, we can call the property something else if you want. But I'm fine with the import too.

pierreguilmin (approved):

PR #914, opened 2025-02-14
Add a warning in `dq.wigner()` when the state is not normalized

cyril87885:
-  _checks.py : Add the function "warn_non_normalize" with the decimal precision discussed in Slack
- utils/wigner_utils.py : called the function created in _checks.py
- tests/sesolve/test_checks.py : Tests the warning works correctly with valid and invalid cases

PR #925, opened 2025-02-26
Remove normalization warning in `dq.wigner()` (not jit-compatible)

gautierronan:
Follow-up to https://github.com/dynamiqs/dynamiqs/pull/914 to fix the following MWE:
```python
import dynamiqs as dq
import jax

jax.jit(dq.wigner)(dq.coherent(8, 1.0))
```
cc @cyril87885

### DIFF
--- a/dynamiqs/_checks.py
+++ b/dynamiqs/_checks.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import warnings
 
 import equinox as eqx
+import jax
 import jax.numpy as jnp
 from jax import Array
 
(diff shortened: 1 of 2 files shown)


pierreguilmin (changes requested):
Unfortunately this raises a warning whatever the state norm. 😕

```python
>>> psi = dq.basis(2, 0)
>>> dq.wigner(psi)
UserWarning: Argument state is not normalized (expected norm 1).
```

gautierronan:
> Unfortunately this raises a warning whatever the state norm. 😕

Indeed, this is because both branches of the `lax.cond` are evaluated at trace time. Then I think we should eliminate this check altogether. It is better not to have the warning than to not be jit-compatible. 

Maybe a solution would be something similar to [Equinox runtime errors](https://docs.kidger.site/equinox/api/errors/#equinox.error_if), but it would have to be a runtime warning instead. Probably requires implementing something like [here](https://github.com/patrick-kidger/equinox/blob/8191b113df5d985720e86c0d6292bceb711cbe94/equinox/_errors.py#L283-L344).

pierreguilmin:
I think we should either always warn or never, otherwise the user may wrongly assume that the norm is correct bc he is used to the warning mechanism which doesn't work sometimes.

gautierronan:
Ok, then I removed the warning altogether. That's quite unfortunate :(

pierreguilmin:
Yes it's a bit of a shame. 🥲

pierreguilmin (approved):

pierreguilmin:
CC @CyrilDeloince, we have to revert #914. The warning is not compatible with the JAX JIT mechanism, and we couldn't find a workaround.

PR #926, opened 2025-02-28
Fix alpha channel in gifit

davidschlegel:
If the plot contains non-trivial alpha values (e.g. by setting an transparent figure or axis background using `matplotlib.rcParams`), now this is reflected in `gifit`.

Before, all alpha values were discarded.

### DIFF
--- a/dynamiqs/integrators/apis/dssesolve.py
+++ b/dynamiqs/integrators/apis/dssesolve.py
@@ -92,8 +92,8 @@ def dssesolve(
 
     Args:
         H _(qarray-like or time-qarray of shape (...H, n, n))_: Hamiltonian.
-        jump_ops _(list of qarray-like or time-qarray, each of shape (n, n))_: List of
-            jump operators.
+        jump_ops _(list of qarray-like or time-qarray, each of shape (...Lk, n, n))_:
+            List of jump operators.
         psi0 _(qarray-like of shape (...psi0, n, 1))_: Initial state.
         tsave _(array-like of shape (ntsave,))_: Times at which the states and
             expectation values are saved. The equation is solved from `tsave[0]` to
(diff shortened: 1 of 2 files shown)


pierreguilmin (changes requested):
Thanks a lot! Unfortunately this re-introduces an old color quantization bug (see https://github.com/dynamiqs/dynamiqs/pull/723). If you find a way around we'll be happy to merge it.

You can test with
```python
import dynamiqs as dq
import jax.numpy as jnp

n = 16
a = dq.destroy(n)
H = dq.zeros(n)
jump_ops = [a @ a - 4.0 * dq.eye(n)]  # cat state inflation
psi0 = dq.coherent(n, 0)
tsave = jnp.linspace(0, 1.0, 1001)
dq.plot.wigner_gif(result.states, fps=25, xmax=4.0, ymax=2.0)
```

Result on `main`:
34b40ef2-e34e-4057-b413-52cc9f56ee47

Result for your PR:
1ec1e578-358c-4623-b1f4-a215a478d5f3

davidschlegel:
Okay, I have to see what actually causes this (the `RGBA` attribute or the `disposal` argument)

davidschlegel:
Actually, I see the issue right [here](https://github.com/python-pillow/Pillow/blob/2755e0ffaadc8b29c3e67e223c333c50e197a733/src/PIL/Image.py#L964-L965):
```python
if mode == "P" and self.mode == "RGBA":
      return self.quantize(colors)
```

I am trying to find a workaround.

davidschlegel:
It seems a little more tricky.
Quantizing RGBA values can be done via `libimagequant`, but this is not natively included in Pillow, as it has a different licence. For it to work requires building Pillow from source against libimagequant.
Is this something that we could do?

pierreguilmin:
> It seems a little more tricky. Quantizing RGBA values can be done via `libimagequant`, but this is not natively included in Pillow, as it has a different licence. For it to work requires building Pillow from source against libimagequant. Is this something that we could do?

This would be a bit too advanced for just this specific feature, especially since we'd need to include it in the CI and the installation process. So no this is not something we can do.

pierreguilmin:
Maybe @gautierronan or @abocquet have ideas on how to fix this?

PR #927, opened 2025-03-04
Fix public/private methods convention

pierreguilmin:
Follow-up to #922.


gautierronan (approved):
Thank you ! Had missed those

PR #928, opened 2025-03-04
Add qarray matmul priority mechanism

pierreguilmin:
This avoids importing sparse dia qarray in dense qarray matmul, and is overall a cleaner mechanism.


gautierronan (approved):
Should we add a test for this? I fear we might miss again the silent conversion from sparse to dense in case of future changes.

pierreguilmin:
Yes!

PR #929, opened 2025-03-04
Add qarray `x.mean()` method

pierreguilmin:


gautierronan (approved):
Nice! I realize we don't have documentation for this method, or `.sum()` for instance (maybe others too). We should add it in a separate PR.

PR #934, opened 2025-03-07
Simplify `__add__` in timeqarray

pierreguilmin:
On top of #933. Taken from https://github.com/dynamiqs/dynamiqs/pull/886. This fixes the incorrect `__add__` implementation of `CallableTimeQArray`.

PR #935, opened 2025-03-08
Add `x.clip()` method to timeqarray

pierreguilmin:
On top of #934.


pierreguilmin:
@gautierronan this is now ready for review 🚀

pierreguilmin:
I am then planning a `.shift()` method (even simpler to implement) to change the start time of the pulse. This would be very convenient to fit electrical delays for example.

dynamiqs/time_qarray.py
```diff
@@ -291,6 +302,11 @@ def ndim(self) -> int:
     def ndiags(self) -> int:
         pass
 
+    @property
+    @abstractmethod
+    def offsets(self) -> tuple[int, ...]:
+        pass
```
gautierronan: This is ill-defined for generic QArrays, since DenseQArrays won't have this property. Also, it looks quite unrelated to the PR and you don't use it, so I'd be in favor of removing this here and in children classes.

dynamiqs/time_qarray.py
```diff
@@ -314,10 +330,36 @@ def in_axes(self) -> PyTree[int | None]:
...
+        times = []
+        if self.tstart is not None:
+            times += [self.tstart]
+        if self.tend is not None:
+            times += [self.tend]
+        return jnp.array(times) if len(times) > 0 else jnp.empty(0)
+
+    def clip(
+        self, tstart: float | None = None, tend: float | None = None
+    ) -> TimeQArray:
```
gautierronan: I find the API and the warning a bit misleading: it's hard to understand exactly what's done under the hood. For instance, if one calls `x.clip(0, 1)` and then `x.clip(0.5)`, it's not necessarily evident that only `(0.5, None)` is kept.

I would thus propose to remove default values, which would at least force users to write `x.clip(0.5, None)`, and to update the warning to saying that if `None` is specified, it will override any existing value.

```suggestion
    def clip(self, tstart: float | None, tend: float | None) -> TimeQArray:
```

gautierronan (changes requested):

gautierronan (approved):
Thank you!

PR #937, opened 2025-03-11
Direct indexing support for sparse arrays

abocquet:
Adds the possibility to efficiently index sparse arrays.
TODO:
- [ ] add index checks
- [ ] support stride


pierreguilmin (changes requested):
I think we should either go fully general (e.g. support `x[1:2, (1, 2, 3), None]` which is supported today on JAX/NumPy array), or explicitly restrict what we support today, and maybe mention it in a comment. Moreover, I fear the implementation is overly explicit and complicated, making it hard to review and maintain. I'm sure we can condense this by playing a few tricks. Perhaps being more general would actually help in simplifying the code? I'm happy to give it a try, but I haven't looked into the problem much, so I'd love to have your input before diving down the rabbit hole @abocquet. 😊

pierreguilmin:
Another remark: I believe our current data format for sparse dia is particularly unsuitable, and we should store diagonals and offsets in a dict structure. Would this help simplifying your implementation? I have a draft PR to reimplement sparse dia primitives for the new format, but I haven't finished it yet.

<details>
  <summary>Example on `__add__()`</summary>

**Today's data structure**
```python
def add_sparsedia_sparsedia(
    left_offsets: tuple[int, ...],
    left_diags: Array,
    right_offsets: tuple[int, ...],
    right_diags: Array,
) -> tuple[tuple[int, ...], Array]:
    # compute the output offsets
    out_offsets = np.union1d(left_offsets, right_offsets).astype(int)

    # initialize the output diagonals
    batch_shape = jnp.broadcast_shapes(left_diags.shape[:-2], right_diags.shape[:-2])
    diags_shape = (*batch_shape, len(out_offsets), left_diags.shape[-1])
    dtype = jnp.promote_types(left_diags.dtype, right_diags.dtype)
    out_diags = jnp.zeros(diags_shape, dtype=dtype)

    # loop over each offset and fill the output
    for i, offset in enumerate(out_offsets):
        if offset in left_offsets:
            left_diag = left_diags[..., left_offsets.index(offset), :]
            out_diags = out_diags.at[..., i, :].add(left_diag)
        if offset in right_offsets:
            right_diag = right_diags[..., right_offsets.index(offset), :]
            out_diags = out_diags.at[..., i, :].add(right_diag)

    return _numpy_to_tuple(out_offsets), out_diags
```

**Dict data structure**
```python
def add_sparsedia_sparsedia(left: dict[int, Array], right: dict[int, Array]) -> dict[int, Array]:
    return {k: left.get(k, 0) + right.get(k, 0) for k in left.keys() | right.keys()}
```

</details>

This is an ongoing debate with @gautierronan but I see no performance degradation, and _if_ some functions were much easier to write with the old data structure it's trivial to construct it from the dict data structure. So if it helps simplifying this PR that would be a motivation to finish it. 😇

abocquet:
@pierreguilmin I agree, this is a first shot at this and as you can see it already is somewhat non-trivial. We can probably make the code shorter and definitely make it more general but I didn't want to swamp you in a 400+ lines PR. As for the `dict` I am agnostic. I don't see a reason not to do it.

pierreguilmin:
Ok, but I think this is exactly what I fear: that we reimplement a lot of complex indexing logic for indexing arrays that is already implemented somewhere in JAX/numpy. We "just" need to do something specific for the two last dimensions, I don't think we should reimplement a huge machinery for this. Or at least I hope there's a way around.

(Actually a simple initial solution first is to materialize the dense matrix and use standard indexing. This will only fail for sparse Liouvillians, which we don't support yet.)

abocquet:
Note that all the complex indexing already works for batching dimensions

>(Actually a simple initial solution first is to materialize the dense matrix and use standard indexing. This will only fail for sparse Liouvillians, which we don't support yet.)
This is precisely what we hoped to avoid with this PR I believe

dynamiqs/result.py
```diff
@@ -144,11 +144,12 @@ def _str_parts(self) -> dict[str, str | None]:
 
 class PropagatorResult(Result):
     @property
-    def propagators(self) -> QArray:
+    def propagators(self) -> Array:
```
pierreguilmin, line 147: Can we convert the object back to a qarray in the propagator solver? (and remove the changes in this file)

dynamiqs/qarrays/qarray.py
```diff
@@ -257,10 +257,16 @@ def __check_init__(self):
                 f'Argument `dims` must be a tuple of ints, but is {self.dims}.'
             )
 
+        if self.shape == ():
+            # handles the case where the array is a single scalar. There
+            # is no need to check the shape then
+            return
```
pierreguilmin, line 263: I'm surprised by this, I don't think we should authorize scalar qarrays?

dynamiqs/qarrays/qarray.py
```diff
@@ -654,6 +660,38 @@ def include_last_two_dims(axis: int | tuple[int, ...] | None, ndim: int) -> bool
     )
 
 
+def _is_key_in_batch_dims(key: int | slice | tuple, ndim: int) -> bool:
```
pierreguilmin, line 663: This is not the correct typing for `key` (missing `None`, tuple type specification, and we can also have `x[1:2, (1, 2, 3), None]`).

pierreguilmin (changes requested):
Thanks for the change! 🙏 I think this is easier to start with.

PR #933, opened 2025-03-07
Use `_replace()` to simplify timeqarray implementation

pierreguilmin:
Taken from #886. This lays the ground for future work.

PR #938, opened 2025-03-15
Replace discontinuity times `None` default value by empty array

pierreguilmin:
The typing makes more sense and it simplifies subsequent operations on this object. This follows up from work in #886, #933, #934.


gautierronan (approved):
All good !

PR #942, opened 2025-03-16
Support `jump_ops` batching in jump and diffusive SSE solvers

pierreguilmin:
- Support `jump_ops` batching in `dq.dssesolve()`.
- Fix corresponding doc in `dq.jssesolve()`.


gautierronan (approved):
So it worked already in the end? Lovely

pierreguilmin:
It always did for the SSE, the SME is more annoying because of the etas and thetas 🫠

PR #945, opened 2025-03-19
Adds forward mode support from diffrax

gaspardbb:
We needed forward mode differentiation for a project. This uses `diffrax.ForwardMode`. 

It could be mentioned in the documentation that it's probably the most efficient in most quantum simulation. Or made the default mode. We usually have a Lindbladian parametrized with a few values (<~100) and the output is higher dimensional (a density matrix of dim 100x100 or e.g. >~100 observables in time).

### DIFF
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -74,6 +74,7 @@ The **Dynamiqs** Python API features two main types of functions: solvers of dif
         members:
         - Autograd
         - CheckpointAutograd
+        - ForwardAutograd
 
 ## Utilities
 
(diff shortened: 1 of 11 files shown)


pierreguilmin:
> It could be mentioned in the documentation that it's probably the most efficient in most quantum simulation. Or made the default mode. We usually have a Lindbladian parametrized with a few values (<~100) and the output is higher dimensional (a density matrix of dim 100x100 or e.g. >~100 observables in time).

As discussed, it is only efficient if the output is high dimensional, so AFAIU not beneficial when minimizing a scalar loss function, albeit I haven't tried benchmarking. 😇

pierreguilmin:
I've added a minor update, we can merge if you agree @gaspardbb. Thanks a lot for the change 😍

gaspardbb:
> As discussed, it is only efficient if the output is high dimensional, so AFAIU not beneficial when minimizing a scalar loss function, albeit I haven't tried benchmarking.

And I agree with that. 
Let's merge!

pierreguilmin (approved):

PR #949, opened 2025-03-27
Feature: quantum relative entropy

davidschlegel:
A new function `entropy_relative` is introduced that computes the relative entropy between two states.

It is computed as
$$S(\rho \| \sigma) = \mathrm{Tr}\rho (\log \rho - \log \sigma)$$
Note that $S(\rho \| \sigma)$ is not symmetric.

If two states are classically related, (i.e. $\rho$ and $\sigma$ commute), the quantum relative entropy reduces to the classical Kulback-Leibler divergence.

Example: 
```python
>>> rho = (dq.fock_dm(2, 0) + dq.fock_dm(2, 1)).unit()
>>> sigma = dq.fock_dm(2, 0)
>>> dq.entropy_relative(rho, sigma)
>>> Array(inf, dtype=float64)
>>> dq.entropy_relative(sigma, rho)
>>> Array(0.69314718, dtype=float64)
```

The function should be fully batchable.

Computations for kets are non-ideal as they are converted to density matrices.

### DIFF
--- a/dynamiqs/utils/general.py
+++ b/dynamiqs/utils/general.py
@@ -1085,13 +1085,12 @@ def entropy_relative(rho: QArrayLike, sigma: QArrayLike) -> Array:
     nrvals = jnp.where(rvals < 0, 0, rvals)
     nsvals = jnp.where(svals < 0, 0, svals)
 
-    S = jnp.nan_to_num(
+    return jnp.nan_to_num(
         nrvals
         * (jnp.log(nrvals) - (P * jnp.expand_dims(jnp.log(nsvals), (-2))).sum(-1)),
         posinf=jnp.inf,
         neginf=-jnp.inf,
     ).sum(-1)
-    return S
 
 
 def bloch_coordinates(x: QArrayLike) -> Array:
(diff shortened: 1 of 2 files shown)


pierreguilmin (changes requested):
Looks good, I've added a few minor fixes. Could you explain a bit more (by commenting in the PR) how you implement the formula? Also, are there simpler examples where the relative quantum entropy is known to be sqrt(2) or smthg like this?

pierreguilmin:
Arf, I added a few fixes, but my commits didn't pass, let me see. Next time if you could create a new branch on your remote rather than directly using `main` that would make things easier! 😊

davidschlegel:
Is there any update with regards to this PR?
Is there something I can do on my side?

pierreguilmin:
> Looks good! Could you explain a bit more (by commenting in the PR) how you implement the formula? Also, are there simpler examples where the relative quantum entropy is known to be sqrt(2) or smthg like this?

Hey @davidschlegel, yes these two and then we can merge. 😊

pierreguilmin:
Great, thanks a lot @davidschlegel!

PR #950, opened 2025-03-29
Added the tutorial for using pip and virtual environments

CyrilDeloince:
- Added the tutorial for using pip and virtual environments.


pierreguilmin (approved):
Great thanks! I've also edited the doc website contributing, we have two copies of the same document. 😉

PR #951, opened 2025-04-06
Make `method` a non-static argument for all solvers

gautierronan:
This is a small but important PR, allowing the `method` input not to be static anymore. This is required for the implementation of the `JumpMonteCarlo` method since its argument `keys` is non-hashable (PRNGKeyArray) but stored inside `method`.

Note that it also fixes the "bug" where solvers are jitted again when changing the arguments inside `method`, e.g. when changing from `method=dq.method.Tsit5(atol=1e-6)` to `method=dq.method.Tsit5(atol=1e-5)`.

### DIFF
--- a/dynamiqs/integrators/apis/dsmesolve.py
+++ b/dynamiqs/integrators/apis/dsmesolve.py
@@ -266,7 +266,7 @@ def dsmesolve(
 
 
 @catch_xla_runtime_error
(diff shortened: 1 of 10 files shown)


abocquet (approved):

pierreguilmin (approved):
Very cool!

PR #952, opened 2025-04-06
Implement the `JumpMonteCarlo` method

gautierronan:


dynamiqs/integrators/apis/jssesolve.py
```diff
@@ -85,9 +85,9 @@ def jssesolve(
     $I_k = \{t \in [t_0, t_\text{end}[ \,|\, \dd N_k(t)=1\}$.
 
     !!! Note
-        If you are looking for the equivalent of QuTiP's `mcsolve`, look no further!
-        `jssesolve` computes the very same quantities, albeit under a different name in
-        this library.
+        If you are only interested in an efficient method to compute the master
+        equation, consider using [`dq.mesolve()`][dynamiqs.mesolve] with the
+        [`dq.method.JumpMonteCarlo()`][dynamiqs.method.JumpMonteCarlo] method.
```
dkweiss31, line 90: Do we want to have a `!!! Note` for both of these statements? One is that QuTiP `mcsolve` users are in the right place, and the other is that users only looking for an efficient master-equation solver should use `mesolve` with the appropriate `method`

dkweiss31 (approved):
This LGTM! Looks beautiful and simple. Do we also want to maybe add tests for `mesolve` that use this integrator? Or are we happy with the existing `jssesolve` tests?

dynamiqs/integrators/core/montecarlo_integrator.py
```diff
@@ -0,0 +1,44 @@
...
+    final_state_norm: Array
+
+
+class MESolveJumpMonteCarloIntegrator(BaseIntegrator, MEInterface, SolveInterface):
+    def run(self) -> Result:
+        # call _vectorized_jssesolve to compute the jump SSE results
+        jsse_result = _vectorized_jssesolve(
+            self.H,
+            self.Ls,
+            self.y0,
```
pierreguilmin, line 21: Here `y0` is a density matrix, how can this work? 🤔

dynamiqs/integrators/core/montecarlo_integrator.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+
+        # compute saved with the mean results
+        saved = SolveSaved(
+            ysave=jsse_result.mean_states,
+            Esave=jsse_result.mean_expects,
+            extra=JumpMonteCarloExtra(
+                clicktimes=jsse_result.clicktimes,
+                final_state_norm=jsse_result.final_state_norm,
+            ),
```
pierreguilmin, line 37: If we want to save this it should be in infos, not extra. `result.extra` is for
> Extra data saved with `save_extra()` if specified in `options`.

Now:
- I don't think we want to add these two bits to the infos, then the user should _really_ use `jssesolve`.
- We should add back the `save_extra` mechanism, or document that it doesn't work.

pierreguilmin (changes requested):
Very nice! Small feedback.

dynamiqs/integrators/core/montecarlo_integrator.py
```diff
@@ -0,0 +1,44 @@
...
+        jsse_result = _vectorized_jssesolve(
+            self.H,
+            self.Ls,
+            self.y0,
+            self.ts,
+            self.method.keys,
+            self.Es,
+            self.method.jsse_method,
+            self.gradient,
+            self.method.jsse_options,
```
pierreguilmin, line 27: Sadly, options passed to `mesolve` are completely ignored. Maybe we should do smthg smarter.

dynamiqs/integrators/core/montecarlo_integrator.py
gautierronan: (still need to fix this one, will try to do tomorrow)

dynamiqs/method.py
```diff
@@ -528,3 +530,53 @@ def __init__(
...
+    (ME). These trajectories are then averaged to obtain an approximation of the ME
+    solution.
+
+    Note:
+        This method is solely a wrapper around [`dq.jssesolve()`][dynamiqs.jssesolve].
+        If you are looking for direct access to individual trajectories, use
+        [`dq.jssesolve()`][dynamiqs.jssesolve] instead.
+
+    Warning:
+        This method does not support the `progress_meter` option.
```
pierreguilmin: We should remove this IMO, many of our methods don't support some of the options, depending on the solvers. I find it weird to specify it only for this one.

dynamiqs/integrators/apis/jssesolve.py
```diff
@@ -78,9 +78,12 @@ def jssesolve(
...
     !!! Note
-        If you are looking for the equivalent of QuTiP's `mcsolve`, look no further!
-        `jssesolve` computes the very same quantities, albeit under a different name in
-        this library.
+        If you are only interested in simulating trajectories to solve the Lindblad
+        master equation, consider using [`dq.mesolve()`][dynamiqs.mesolve] with the
+        [`dq.method.JumpMonteCarlo`][dynamiqs.method.JumpMonteCarlo] method.
+
+    !!! Note
+        This function is the Dynamiqs counterpart of QuTiP's `mcsolve()` function.
```
pierreguilmin, line 86: Visual bug in the doc for some reason. 🤷‍♂️
Screenshot 2025-06-05 at 15 45 19

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
```diff
@@ -227,7 +227,7 @@ def save(self, y: PyTree) -> Saved:
         return super().save(y.state)
 
     def postprocess_saved(self, saved: Saved, ylast: PyTree) -> Saved:
-        saved = super().postprocess_saved(saved, ylast.state)
+        saved = super().postprocess_saved(saved, ylast.state[None, :])
```
pierreguilmin, line 229: This change (and the next one) break the saved state shape for existing stochastic solvers no?

gautierronan: Yes, but it's rather a fix for the stochastic solvers. When save_states=False, the output shape of states is wrong, and missing the nsave=1 dimension.

pierreguilmin: Ha nice! Where is the equivalent for sesolve/mesolve? I'm confused I can't find it.

gautierronan: Yes I was confused too. When using Diffrax with an additional SaveAt for the final state, it already returns the final state with shape (1, n, n)!

pierreguilmin: 🤯 Good catch!

pierreguilmin (approved):
Looks good, let's ship! 🥳

PR #954, opened 2025-04-15
Add an optional offset argument to dq.asqarray to make it jit-compatible

gautierronan:
Similar to e.g. [`jnp.nonzero`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.nonzero.html).

### DIFF
--- a/dynamiqs/qarrays/dense_qarray.py
+++ b/dynamiqs/qarrays/dense_qarray.py
@@ -136,10 +136,10 @@ class DenseQArray(QArray):
     def asdense(self) -> DenseQArray:
         return self
 
-    def assparsedia(self) -> SparseDIAQArray:
+    def assparsedia(self, offsets: tuple[int, ...] | None = None) -> SparseDIAQArray:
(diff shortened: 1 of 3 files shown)


abocquet (dismissed):

pierreguilmin:
Nice! We need this for all functions that create sparse DIA qarrays though. 🫠 (see e.g. in #935) What about an API like `dq.dia` or `dq.dia(offsets=...)` to lighten the function signatures?

gautierronan:
> Nice! We need this for all functions that create sparse DIA qarrays though. 🫠 (see e.g. in #935) What about an API like `dq.dia` or `dq.dia(offsets=...)` to lighten the functions signatures?

Sorry for the late reply. Not sure I get your comment @pierreguilmin ? In principle it's only for conversion functions: do we have other functions where this would be relevant? Note that once we've implemented dynamic offsets, we'll be able to replace `offsets` by `noffsets`, which is quite a bit lighter already.

PR #956, opened 2025-04-16
Add Euler-Jump solver to jump SSE and SME and fix some issues with Event solver

pierreguilmin:
Sorry for this massive PR. 🙈 I suggest directly reading the new files rather than the diffs.

**Changes**
- Added an Euler-Jump solver to `dq.jssesolve()` and `dq.jsmesolve()` (by generalising and reusing the existing fixed-step stochastic integrator for diffusive SSE/SME). I temporarily removed the `Event` solver for this change and then re-added it; see details just below.
- Added a tutorial on continuous jump measurement.
- Simplified the vectorization over trajectories by separating vectorization over `H`, `jump_ops`, or `psi0/rho0` and vectorization over `keys`.
- Re-added the `Event` solver. The logic remains mostly the same, albeit with some modifications, and I fixed a few issues:
  - Fixed the `result` object API (we now return `noclick_prob` of the good shape, and we don't return `final_state_norm`).
  - Fixed the progress meter (previously displaying multiple times during the solve and breaking the vmap).
  - The logic of the `Event` solver is now well separated from the `dq.jssesolve()` API, and isolated in the `event_integrator.py` file. This allows someone to implement a Gillespie solver without having to change everything, for example.
  - Instead of being a diffrax integrator, the event integrator is just using diffrax integrators (multiple, at different times). I believe this better decouples the logic.
  - When `smart_sampling=True`, we now solve the no-click trajectory without using any event handling, we don't consume a random key (the trajectory is deterministic), and the result is not part of `result.states` but is stored as something specific to the `Event` method, in `result.infos`.

**Points of attention for the review**
- I updated the doc for supported gradients. Correct me if I'm wrong, but I think the code only supports bwd checkpointed, not fwd mode.
- I believe there was a bug in the previous implementation, because `nclicks` was shared by all jump channels, but we should have one index per channel for `clicktimes` of shape `(..., ntrajs, len(jump_ops), nmaxclick)`. See for example [here](https://github.com/dynamiqs/dynamiqs/blob/fdc2e26b9d862ca49184b99d21a7930973f1bc58/dynamiqs/integrators/core/event_integrator.py#L95-L97) in the previous code.

**What's next?**
I have noted several possible improvements and simplifications we could make. I'll write a linear issue at some point. The most important IMO is that the method is unusable without a root finder now (when `root_finder=None`). The reason is that if the no jump evolution is trivial to solve, the adaptive solver will take gigantic steps, leading to extremely imprecise jump times. We should add a way to specify the stepsize controller `dtmax`, so users can control the precision of the click times in that scenario.

**Final words**
Thanks again @dkweiss31 and @gautierronan for the massive work on the `Event` method (#748 and #943), this would have never happened without it! Happy reviewing. 😊

### DIFF
--- a/docs/documentation/advanced_examples/continuous-jump-measurement.md
+++ b/docs/documentation/advanced_examples/continuous-jump-measurement.md
@@ -1,0 +1,198 @@
+# Continuous jump measurement
+
+!!! Warning "Work in progress."
+    This tutorial is under construction, this is a draft version.
+
+In this example, we simulate stochastic trajectories of quantum systems that are continuously measured by a jump detector. We explain how to use [`dq.jssesolve()`][dynamiqs.jssesolve] to simulate trajectories modelled by the jump SSE, and [`dq.jsmesolve()`][dynamiqs.jsmesolve] to simulate trajectories modelled by the jump SME.
+
+```python
+import jax
+import jax.numpy as jnp
+import numpy as np
+from matplotlib import pyplot as plt
+import dynamiqs as dq
+
+dq.plot.mplstyle(dpi=150)  # set custom matplotlib style
+```
+
+## Qubit
+
+We consider a qubit starting from $\ket{\psi_0}=\ket1$, with no unitary dynamic $H=0$ and two loss operators $L_-=\sigma_-$ and $L_+=\sigma_+$ which are continuously measured by jump detectors. We expect stochastic trajectories jumping between the excited and ground state.
+
+Let's begin with a perfectly efficient detector. We start with a pure state, and we measure all loss channels with perfect efficiency, so we don't loose any information about the system's state. As a result, it remains pure at all times. We use [`dq.jssesolve()`][dynamiqs.jssesolve] to simulate these quantum trajectories.
+
+### Simulation
+
+```python
+# define Hamiltonian, jump operators, initial state
+H = dq.zeros(2)
+jump_ops = [dq.sigmam(), dq.sigmap()]
+psi0 = dq.excited()
+
+# define save times
+tsave = np.linspace(0, 1.0, 101)
+
+# define a certain number of PRNG key, one for each trajectory
+key = jax.random.PRNGKey(20)
+ntrajs = 3
+keys = jax.random.split(key, ntrajs)
+
+# simulate trajectories
+method = dq.method.EulerJump(dt=1e-3)
+result = dq.jssesolve(H, jump_ops, psi0, tsave, keys, method=method)
+print(result)
+```
+
+```text title="Output"
+==== JSSESolveResult ====
+Method     : EulerJump
+Infos      : 1000 steps | infos shape (3,)
+States     : QArray complex64 (3, 101, 2, 1) | 7.9 Kb
+Clicktimes : Array float32 (3, 2, 10000) | 390.6 Kb
+```
+
+```pycon
+>>> result.states.shape  # (ntrajs, ntsave, n, 1)
+(3, 101, 2, 1)
+>>> result.clicktimes.shape  # (ntrajs, nLm, nmaxclick)
+(3, 2, 10000)
+```
+
+### Individual state trajectories
+
+```python
+_, axs = dq.plot.grid(ntrajs, ntrajs, w=5, h=3, sharexy=True)
+
+for i in range(ntrajs):
+    ax = next(axs)
+    dq.plot.xyz(result.states[i], times=tsave, ax=ax)
+    ax.set(
+        title=f'Trajectory {i+1}: {result.nclicks[i].sum()} clicks',
+        xlabel=r'time $t$',
+        ylabel='Pauli expectation values',
+    )
+
+renderfig('jump-monitored-qubit-trajs')
+```
+
+![plot_jump_monitored_qubit_trajs](../../figs_docs/jump-monitored-qubit-trajs.png){.fig}
+
+### Imperfect detection
+
+If the detection is imperfect, the system state is a density matrix. We use [`dq.jsmesolve()`][dynamiqs.jsmesolve] to simulate these quantum trajectories. For jump measurement we can model two classes of imperfections: false clicks (accounted by the dark count rate $\theta$) and missed clicks (accounted by the efficiency $\eta$).
+
+```python
+# define dark count rates
+thetas = [0.0, 0.0]
+# define efficiencies
+etas = [0.5, 0.5]
+
+# simulate trajectories
+result = dq.jsmesolve(H, jump_ops, thetas, etas, psi0, tsave, keys, method=method)
+print(result)
+```
+
+```text title="Output"
+==== JSMESolveResult ====
+Method     : EulerJump
+Infos      : 1000 steps | infos shape (3,)
+States     : QArray complex64 (3, 101, 2, 2) | 9.5 Kb
+Clicktimes : Array float32 (3, 2, 10000) | 234.4 Kb
+```
+
+```pycon
+>>> result.states.shape  # (ntrajs, ntsave, n, n)
+(3, 101, 2, 2)
+>>> result.clicktimes.shape  # (ntrajs, nLm, nmaxclick)
+(3, 2, 10000)
+```
+
+```python
+_, axs = dq.plot.grid(ntrajs, ntrajs, w=5, h=3, sharexy=True)
+
+for i in range(ntrajs):
+    ax = next(axs)
+    dq.plot.xyz(result.states[i], times=tsave, ax=ax)
+    ax.set(
+        title=f'Trajectory {i+1}: {result.nclicks[i].sum()} clicks',
+        xlabel=r'time $t$',
+        ylabel='Pauli expectation values',
+    )
+
+renderfig('jump-monitored-qubit-eta-trajs')
+```
+
+![plot_jump_monitored_qubit_eta_trajs](../../figs_docs/jump-monitored-qubit-eta-trajs.png){.fig}
+
+
+## Quantum harmonic oscillator
+
+We consider a quantum harmonic oscillator starting from the Fock state $\ket{n_0}$, with Hamiltonian $H=0$ and a single loss operator $L=\sqrt\kappa a$ which is continuously measured by photodetection.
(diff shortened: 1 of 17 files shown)


dynamiqs/integrators/apis/jssesolve.py
```diff
@@ -305,42 +293,16 @@ def _vectorized_clicks_jssesolve(
...
+        pass
     else:
-        click_args = (keys, False, 0.0)
-        return f(*core_args, *click_args, *other_args)
+        # vectorize input over keys
+        in_axes = (None, None, None, None, 0, None, None, None, None)
+        out_axes = JSSESolveResult(None, None, None, None, 0, 0, 0)
+        f = jax.vmap(f, in_axes, out_axes)
+
+    return f(H, Ls, psi0, tsave, keys, exp_ops, method, gradient, options)
```
dkweiss31, line 305: On line 242 we check instead for `isinstance(method, EulerJump)`, maybe we can do the same here to symmetrize?

dynamiqs/integrators/core/diffrax_integrator.py
```diff
@@ -147,6 +149,66 @@ def infos(self, stats: dict[str, Array]) -> PyTree:
             )
 
 
+def basic_diffeqsolve(
```
dkweiss31, line 152: I'm confused by this naming? What does basic mean?

dkweiss31: I'm also confused about the purpose of this function: it seems that some code from the `run` function on lines 127-139 is duplicated below. Is there a way to combine the functions to prevent this duplication?

pierreguilmin: > I'm confused by this naming? What does basic mean?

Yes it's not well chosen haha. Maybe `diffrax_diffeqsolve` woudl be better?

> I'm also confused about the purpose of this function: it seems that some code from the run function on lines 127-139 is duplicated below. Is there a way to combine the functions to prevent this duplication?

Yes, to be honest this is a quick dirty fix. I think we should revamp a bit the call to diffeqsolve, in a later PR. I just wanted to have an event integrator which calls an ODE solving routing without _being_ the ODE solving routine. I think it's important because the event method solves two different ODEs: the no-click evolution with no event, and the one with the event. In the long term, I have the feeling the inheritance structure causes more harm than help in the integrators in terms of ease of writing the code and comprehension. I think we want to move more towards using composition than inheritance, this is a first small step in that direction. :)

dkweiss31: I think I definitely agree that composition could be better than inheritance in many instances for the library. Nevertheless I am still very confused about what `basic_diffeqsolve` is doing, and why it needs to define the class `BasicDiffraxIntegrator` within it (is this pattern used anywhere else in the library?). This issue of `hav[ing] an event integrator which calls an ODE solving routing without being the ODE solving routine` is a bit of a subtlety that I'm not sure I'm fully grasping, and it isn't clear to me why this is a problem. Moreover this solution seems to have created a (more concrete in my mind!) problem in duplication of code from `DiffraxIntegrator`. Happy for this to be addressed in a future PR! Just wanted to register my confusion, and maybe suggest that it might be useful to add comments about why this function is needed.

dynamiqs/result.py
```diff
@@ -208,63 +213,45 @@ def clicktimes(self) -> Array:
     def nclicks(self) -> Array:
         return jnp.count_nonzero(~jnp.isnan(self.clicktimes), axis=-1)
 
-    @property
-    def final_state_norm(self) -> Array:
-        return self._saved.final_state_norm
-
-    def noclick_states(self) -> QArray:
-        return self.states[..., 0, :, :, :]
```
dkweiss31, line 216: Is it useful to add a simple helper method

```python
 def noclick_states(self) -> QArray:
        return self.infos.noclick_states
```
? It just seems strange for the user to have to go through `infos` to get to the `noclick_states`. Said another way, this API treats the noclick and click states on a different footing, treating one simply as "additional info" and the other as the true "states" which doesn't feel appropriate to me.

dynamiqs/integrators/core/event_integrator.py
```diff
@@ -1,219 +1,213 @@
...
+            save_body,
+            y,
+            max_steps=len(self.ts),
+            buffers=lambda x: x.saved,
+            kind='checkpointed',
+        )
+
+        return y, click_occurred
+
+    def simulate_single_trajectory(
```
gautierronan, line 145:
```suggestion
    def _solve_single_trajectory(
```

dynamiqs/result.py
gautierronan: I agree that the API is misleading in the proposed form. It is one of these cases for which we might want to have different result classes, although the library is really not built this way for now.

As an intermediate solution, I would definitely like to have these `self.infos` properties promoted to `self` with helper functions that can return an appropriate error message if the method is not `Events`.

dynamiqs/result.py
```diff
@@ -208,63 +213,45 @@ def clicktimes(self) -> Array:
...
+        return d | {'Clicktimes': _array_str(self.clicktimes)}
 
-    @property
     def mean_states(self) -> QArray:
+        # todo: document
         if self.method.smart_sampling:
-            noclick_prob = self.final_state_norm[..., 0, None, None, None] ** 2
-            states_noclick = self.states[..., 0, :, :, :].todm()
-            states_click = self.states[..., 1:, :, :, :].todm().mean(axis=-4)
+            noclick_prob = 0
```
gautierronan, line 223: What is this for?

dynamiqs/result.py
```diff
@@ -208,63 +213,45 @@ def clicktimes(self) -> Array:
...
-            'No click probability': _array_str(self.noclick_prob),
-        }
-
-    @classmethod
-    def out_axes(cls) -> SolveResult:
-        return cls(None, None, None, None, 0, 0, 0)
+        return d | {'Clicktimes': _array_str(self.clicktimes)}
 
-    @property
     def mean_states(self) -> QArray:
```
gautierronan, line 220: There is a mismatch between these two things. Either it's a property and it can be called `mean_states`, or it's a method and it should be something like `compute_mean_states` (but I find that verbose).

dynamiqs/result.py
```diff
@@ -208,63 +213,45 @@ def clicktimes(self) -> Array:
...
-                noclick_prob * states_noclick + (1 - noclick_prob) * states_click
+                noclick_prob * self.infos.noclick_states.todm()
+                + (1 - noclick_prob) * self.states.todm().mean(axis=-4)
             )
         else:
             return self.states.todm().mean(axis=-4)
 
-    @property
-    def mean_expects(self) -> Array:
+    def mean_expects(self) -> Array | None:
```
gautierronan, line 232: Same comment here

docs/documentation/advanced_examples/continuous-jump-measurement.md
```diff
@@ -0,0 +1,198 @@
+# Continuous jump measurement
+
+!!! Warning "Work in progress."
+    This tutorial is under construction, this is a draft version.
```
gautierronan, line 4: We both know that we'll never come back to make it a not-draft version :D

So I'd propose removing this warning, and just accept that it's good enough for our standard of examples for now. And we can come back to it for improvements when we decide to generally improve the examples documentation.

gautierronan (changes requested):
Great work @pierreguilmin. These changes are really useful, and I really like the way you've reworked the Events integrator. I have a bunch of comments below for suggestions of improvements, but the overall structure is perfect :)

gautierronan (commented):
Don't forget to update `docs/index.md` also with your new tutorial! :)

dynamiqs/integrators/core/diffrax_integrator.py
pierreguilmin: I think one way to understand the role of this function is as a "custom" diffeqsolve that takes as input Dynamiqs parameters (for gradient, options, solvers) and converts them to Diffrax's parameters. But confusion registered @dkweiss31, I'll do my best to make this cleaner/more natural in upcoming PRs! 😇

dynamiqs/result.py
pierreguilmin: Ok, I propose we do this in another PR (as this one is quite big and old), it'll make it easier to review.

dynamiqs/result.py
pierreguilmin: Old code 😅

docs/documentation/advanced_examples/continuous-jump-measurement.md
pierreguilmin: I think even if we never come back to it it's good to have different level of documentation readiness. Even for us, to keep track of which tutorials are drafty, which are not. To me it's rather a good way to ship tutorials before we are satisfied of how they look like. 😇

dynamiqs/integrators/apis/jsmesolve.py
```diff
@@ -214,4 +227,219 @@ def jsmesolve(
...
+        raise ValueError(
+            'Argument `thetas` should be of the same length as argument `jump_ops`, but'
+            f' len(thetas)={len(thetas)} and len(jump_ops)={len(Ls)}.'
+        )
+
+    if jnp.any(thetas < 0):
+        raise ValueError(
+            'Argument `thetas` should only contain values greater than 0, but'
+            f' is {thetas}.'
+        )
```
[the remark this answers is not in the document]
pierreguilmin: Arf, we have the same mistake for `eta` then. Lemme add it to the follow-up issue.

pierreguilmin:
Hey guys! As this PR is rather large and old, I decided to go with only the minor modifications, and dump the other proposed changes in a follow-up issue > https://linear.app/dynamiqs/issue/DYN-259/improve-event-solver.

dynamiqs/integrators/core/diffrax_integrator.py
dkweiss31: Sounds great! Thanks so much for the hard work on improving the event functionality!!

PR #957, opened 2025-04-18
Add support for Hermitian matrix in `dq.norm()` and fix documentation

AurelienGauffre:
This PR introduces the `trace_norm` function, which computes the trace norm (also known as the nuclear norm) of a Hermitian matrix, such as a quantum density matrix. This norm can be used in particular to compute the trace distance between two density matrices.

### Motivation
In most practical quantum applications, the input to `trace_norm` should be a density matrix, which is Hermitian by construction, and thus allows using the internal `_eigvalsh()` method from `QArray` instead of `_eigvals()`. This avoids computing eigenvectors and results in a significant speedup, especially for large matrices. See benchmark comparison here: [Colab : eigvals vs eigvalsh benchmark](https://colab.research.google.com/drive/1adAj7IYUV1FruMuyBKknmjwNoxvnmBup#scrollTo=m4qi6E8O38yN) (~x50 less time for 1024x1024 matrices) . If needed, this function can be extended to support general (non-Hermitian) matrices by adding a check for Hermiticity and applying a different norm accordingly.


pierreguilmin (approved):
Nice! Thanks a lot for your first contribution. 🤓 I've made a few minor fixes (you can check how the documentation renders locally with `task docserve`) and added the `trace_distance` function while we're at it. I'm happy to merge, let's wait for @gautierronan's approval.

gautierronan (changes requested):
Thanks @AurelienGauffre for the PR! I hope @pierreguilmin did not force you into it :grin:

I hope I'm not being a wet blanket, but I don't really see the point of these two functions :sweat_smile::see_no_evil:

- `trace_norm` seems very niche to me in the context of quantum mechanics (I've never seen it used). I would thus propose merging it with `dq.norm` with an additional optional argument to avoid overloading the utils API. This is how both numpy and pytorch handle this, see [`numpy.linalg.norm`](https://numpy.org/devdocs/reference/generated/numpy.linalg.norm.html) and [`torch.norm`](https://pytorch.org/docs/stable/generated/torch.norm.html).
- `trace_distance` seems even more niche to me. Since there's no speedup whatsoever associated to it (it just calls `trace_norm(x-y)`), I don't think we should keep it.

gautierronan:
So, correct me if I'm wrong, but the trace norm reduces to the trace when the operator is positive semi-definite. So, our function `dq.norm` indeed computes a norm, but only for density matrices. Since we also have a function `dq.trace`, it is our mistake that `dq.norm` is simply an alias for `dq.trace`.

Therefore, I stand to my point, your function should be merged/replace `dq.norm`. However, there should be a side-case for when the input object is psd, in which case it should reduce to just computing the trace for efficiency.

pierreguilmin:
> Therefore, I stand to my point, your function should be merged with/replace dq.norm. However, there should be a side-case for when the input object is psd, in which case it should reduce to just computing the trace for efficiency.

Yes, I 100% agree with this! To know if a matrix is psd we need to do eigh anyway. So `dq.norm` becomes "norm of a vector (p=2 by default) or of a Hermitian matrix (p=1 by default)"? And the user can set `p`?

PR #958, opened 2025-04-21
Reorder utilities and methods by alphabetic order in docs

gautierronan:
Since we're beginning to have quite a few functions in the utils API, it can be complicated to find certain functions. I propose reordering them in the public API by alphabetical order instead of the current arbitrary/logical order. Happy to go for a different ordering too, maybe some in-between, but I think some change from the current ordering would be useful.

See e.g. the pytorch documentation, [here](https://pytorch.org/docs/stable/torch.html#indexing-slicing-joining-mutating-ops). Note that it's not done systematically, only for longer lists of functions.

### DIFF
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -55,16 +55,16 @@ The **Dynamiqs** Python API features two main types of functions: solvers of dif
     options:
         table: true
         members:
-        - Tsit5
         - Dopri5
         - Dopri8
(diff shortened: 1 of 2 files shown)


gautierronan:
Actually it's not so bad currently, and I'm not sure alphabetical order is better. I'll close this PR for now, unless you agree with me.

PR #960, opened 2025-04-28
Add steady-state solver

derekeverett:
This commit adds an option to halt on detecting a steady state event during mesolve. It modifies the OCavity OpenSystem with a steady_state method, and adds a unit test checking that the solver halts. Because diffrax returns infs for all tsave after a steady state event, the last valid value is copied forward in time.


derekeverett:
Closing this PR and will make a new PR with the requested API changes.

PR #964, opened 2025-05-14
Use `dataclasses.replace` to simplify code

pierreguilmin:
~~@gautierronan if you have an idea to make this work for static fields, we could also save a lot of code repetition by removing the qarrays `_replace()` method.~~ -> we use `dataclasses.replace`.


gautierronan (approved):
LGTM!

PR #966, opened 2025-05-14
Expose `dq.plot.wigner_data()` to plot pre-computed wigner functions

pierreguilmin:
I added a safeguard for people who are looking for `dq.plot.wigner()`.

```python
>>> import dynamiqs as dq
>>> rho = dq.coherent_dm(4, 1.0)
>>> dq.plot.wigner_data(rho, 2.0, 2.0)
TypeError: Wigner data must be of type `float`, not `complex64`. Consider using `dq.plot.wigner()` directly to pass a quantum state.
>>> x, y, w = dq.wigner(rho)
>>> dq.plot.wigner_data(w, x[-1], y[-1])
```

### DIFF
--- a/dynamiqs/plot/wigner.py
+++ b/dynamiqs/plot/wigner.py
@@ -30,10 +30,19 @@ def wigner_data(
     cross: bool = False,
     clear: bool = False,
 ):
-    r"""Plot the Wigner function for pre-computed Wigner data.
+    r"""Plot a pre-computed Wigner function.
+
+    Warning:
+        Documentation redaction in progress.
 
     Note:
-        See [`dq.plot.wigner()`][dynamiqs.plot.wigner] for more details.
+        Choose a diverging colormap `cmap` for better results.
+
+    See also:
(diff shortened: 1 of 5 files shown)


dynamiqs/plot/wigner.py
```diff
@@ -30,8 +30,18 @@ def _plot_wigner_data(
     cross: bool = False,
     clear: bool = False,
 ):
+    r"""Plot the Wigner function for pre-computed Wigner data.
```
gautierronan, line 33:
```suggestion
    r"""Plot the Wigner function of a pre-computed Wigner distribution.
```

dynamiqs/plot/wigner.py
```diff
@@ -30,8 +30,18 @@ def _plot_wigner_data(
...
+
+    Note:
+        See [`dq.plot.wigner()`][dynamiqs.plot.wigner] for more details.
```
gautierronan, line 36: How about the "See Also" admonition here? Could be quite relevant to link to both `dq.plot.wigner` and `dq.wigner`.

gautierronan (changes requested):
Small comments, otherwise looks good :)

PR #967, opened 2025-05-20
Add `Rouchon2` and `Rouchon3` method for `dq.mesolve()`

pierreguilmin:
Based on https://arxiv.org/abs/2503.01712 + code shared by @RemiRobin.

PR #968, opened 2025-05-20
Add adaptive rouchon 1-2 and 2-3 methods for `dq.mesolve()`

pierreguilmin:
On top of #967. These solvers are generally (much) faster than existing solvers whenever there is a stiff term in the Hamiltonian (e.g. $a^{\dagger3}a^3$) or in one of the jump operators (e.g. $a^2$). For example, at fixed rtol/atol precision per step on a CPU in dense layout for a stiff $H=a^{\dagger 3}a^3$, x15 faster than `Dopri5` for `n=32` and x130 faster for `n=64`! 🏎️

scaling

```python
import dynamiqs as dq
import jax.numpy as jnp

dq.set_precision('double')

n = 64
a = dq.destroy(n)

H = a.dag().powm(3) @ a.powm(3)
jump_ops = [0.1 * a]
psi0 = dq.coherent(n, 1.0)

method = dq.method.Dopri5(max_steps=int(1e7))
result1 = dq.mesolve(H, jump_ops, psi0, tsave, method=method)
print(result1.infos)
# |██████████| 100.0% ◆ elapsed 02m12s ◆ remaining 0.00ms  
# 236695 steps (236675 accepted, 20 rejected)

method = dq.method.AdaptiveRouchon12()
result2 = dq.mesolve(H, jump_ops, psi0, tsave, method=method)
print(result2.infos)
# |██████████| 100.0% ◆ elapsed 1.39s ◆ remaining 0.00ms     
# 1437 steps (1434 accepted, 3 rejected)

method = dq.method.AdaptiveRouchon23()
result3 = dq.mesolve(H, jump_ops, psi0, tsave, method=method)
print(result3.infos)
# |██████████| 100.0% ◆ elapsed 1.03s ◆ remaining 0.00ms     
# 928 steps (926 accepted, 2 rejected)
```

### DIFF
--- a/dynamiqs/integrators/apis/mesolve.py
+++ b/dynamiqs/integrators/apis/mesolve.py
@@ -11,6 +11,8 @@ from jaxtyping import ArrayLike
 from ..._checks import check_hermitian, check_qarray_is_dense, check_shape, check_times
 from ...gradient import Gradient
 from ...method import (
+    AdaptiveRouchon12,
+    AdaptiveRouchon23,
     Dopri5,
     Dopri8,
     Euler,
@@ -45,6 +47,8 @@ from ..core.diffrax_integrator import (
 )
 from ..core.expm_integrator import mesolve_expm_integrator_constructor
 from ..core.rouchon_integrator import (
+    mesolve_adaptive_rouchon12_integrator_constructor,
+    mesolve_adaptive_rouchon23_integrator_constructor,
     mesolve_rouchon1_integrator_constructor,
     mesolve_rouchon2_integrator_constructor,
     mesolve_rouchon3_integrator_constructor,
@@ -104,6 +108,8 @@ def mesolve(
             [`Rouchon1`][dynamiqs.method.Rouchon1],
             [`Rouchon2`][dynamiqs.method.Rouchon2],
             [`Rouchon3`][dynamiqs.method.Rouchon3],
+            [`AdaptiveRouchon12`][dynamiqs.method.AdaptiveRouchon12],
+            [`AdaptiveRouchon23`][dynamiqs.method.AdaptiveRouchon23],
             [`Expm`][dynamiqs.method.Expm]).
         gradient: Algorithm used to compute the gradient. The default is
             method-dependent, refer to the documentation of the chosen method for more
@@ -294,6 +300,8 @@ def _mesolve(
         Rouchon1: mesolve_rouchon1_integrator_constructor,
         Rouchon2: mesolve_rouchon2_integrator_constructor,
         Rouchon3: mesolve_rouchon3_integrator_constructor,
+        AdaptiveRouchon12: mesolve_adaptive_rouchon12_integrator_constructor,
+        AdaptiveRouchon23: mesolve_adaptive_rouchon23_integrator_constructor,
         Dopri5: mesolve_dopri5_integrator_constructor,
         Dopri8: mesolve_dopri8_integrator_constructor,
         Tsit5: mesolve_tsit5_integrator_constructor,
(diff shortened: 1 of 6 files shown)


pierreguilmin:
We need this last temporary fix commit in the tests for the combination `AdaptiveRouchon23` and `otdqubit`, I don’t know why. If @rgautier you find the bug while reviewing I’d be delighted. 😍

rgautier:
@pierreguilmin I am not working on your project - please check your userid for review

pierreguilmin:
> @pierreguilmin I am not working on your project - please check your userid for review

Sorry, I meant @gautierronan. ☺️

dynamiqs/integrators/apis/mesolve.py
```diff
@@ -294,6 +300,8 @@ def _mesolve(
         Rouchon1: mesolve_rouchon1_integrator_constructor,
         Rouchon2: mesolve_rouchon2_integrator_constructor,
         Rouchon3: mesolve_rouchon3_integrator_constructor,
+        AdaptiveRouchon12: mesolve_adaptive_rouchon12_integrator_constructor,
+        AdaptiveRouchon23: mesolve_adaptive_rouchon23_integrator_constructor,
```
gautierronan, line 304: I'm not a big fan of this API: if we were to implement a non-adaptive `Dopri5` (which Diffrax supports), we wouldn't rename the current `Dopri5` to `AdaptiveDopri5`. It's also quite heavy in the doc.

In other words, `AdaptiveRouchon12` is just a `Rouchon2` in an adaptive time-stepping setting. Therefore, I would be in favor of merging both under the name `Rouchon2`, and either:

- have an additional option `adaptive_stepping: bool` (or similar naming)
- make `dt: float | None = None` which activates adaptive stepping if `dt` is not specified

dynamiqs/integrators/core/rouchon_integrator.py
```diff
@@ -208,3 +214,75 @@ def _kraus_ops(self, t: float, dt: float) -> list[QArray]:
...
+            Ms_1 = MESolveRouchon1Integrator.Ms(H, L, dt, self.method.exact_expm)
+            rho_1 = cholesky_normalize(Ms_1, rho) if self.method.normalize else rho
+            rho_1 = sum([M @ rho_1 @ M.dag() for M in Ms_1])
+
+            # === second order
+            Ms_2 = MESolveRouchon2Integrator.Ms(H, L, dt, self.method.exact_expm)
+            rho_2 = cholesky_normalize(Ms_2, rho) if self.method.normalize else rho
+            rho_2 = sum([M @ rho_2 @ M.dag() for M in Ms_2])
+
+            return rho_2, rho_2 - rho_1
```
gautierronan: Note that Patrick uses `0.5 * (rho_2 - rho_1)` for the error estimate in diffrax. See e.g. https://github.com/patrick-kidger/diffrax/blob/d3c1430b76e7d0158536823724371d7b26f5db8f/diffrax/_solver/implicit_euler.py#L96. Of course, it will only change how restrictive or not you are with `dt`, so I don't think it really matters.

gautierronan (changes requested):
My main comment is on the API. For the quick fix of the test, I couldn't find anything in this PR, but I suspect the previous PR implementing `Rouchon2` and `Rouchon3` instead, cf this comment https://github.com/dynamiqs/dynamiqs/pull/967#discussion_r2131682812.

dynamiqs/integrators/apis/mesolve.py
pierreguilmin: Ok, let's fix this in another PR, it's not trivial.

gautierronan: Oh, I thought we forbade ourselves from force-pushing to main on non-trivial PRs like this one?

In any case, it'd be great to fix this before the release you were planning then. Otherwise this will stay as is for weeks/months before a new release that will break this API.

PR #911, opened 2025-02-10
Replace matrix-matrix mul with matrix-vector mul in `dssesolve`

gaspardbb:
Previous implementation of `dssesolve` was doing an unnecessary matrix-matrix multiplication. 
It can be replaced with two matrix-vector multiplication, which is more efficient. 
A small optimization can be carried out for the computation of the `exp` but I don't understand how all those trailing dimension should be handled.

PR #969, opened 2025-05-27
Fix significant performance issue of fixed-step stochastic jump solver

pierreguilmin:
This fixes a major performance issue with our fixed-step jump solvers. On CPU, the replacement of `jax.lax.cond` by `jax.lax.select` (see e.g. https://github.com/jax-ml/jax/issues/7934) leads to a x100 speedup for trajectories simulation of a qubit.

| Number of trajectories | Duration before the PR         | Duration after the PR | Speedup |
|---------------|---------------------|----------------|-|
| 1,000    |      12.7 s ± 1.36 s  | 119 ms ± 8.87 ms    | x106            |
| 10,000           | 158 s (1 run)      |       1.1 s ± 186 ms          |  x144 |
| 100,000           | ??     | 13.1 s ± 1.54 s per loop        | ??|

The new perfs on GPU are quite impressive:
| Number of trajectories | Duration after the PR (GPU) |
|---------------|---------------------|
| 1,000 | 27.5 ms ± 258 μs per loop |
| 10,000 | 50 ms ± 137 μs per loop |
| 100,000 | 185 ms ± 609 μs per loop |

<details>
  <summary>Benchmark code</summary>

```python
import dynamiqs as dq
import numpy as np
import jax

H = dq.zeros(2)
jump_ops = [dq.sigmam(), dq.sigmap()]
psi0 = dq.excited()
tsave = np.linspace(0, 2.0, 101)
key = jax.random.PRNGKey(42)
method = dq.method.EulerJump(dt=1e-3)
for ntrajs in [int(1e3), int(1e4)]:
    keys = jax.random.split(key, ntrajs)

    def foo():
        return dq.jssesolve(
            H, jump_ops, psi0, tsave, keys, method=method
        ).states.block_until_ready()

    %timeit -r1 -n1 -q foo()
    %timeit foo()
```

</details>

There is another speedup to implement for `jssesolve` vs `jsmesolve`, the same as #911. I can do it once this PR is merged.

### DIFF
--- a/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
+++ b/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
@@ -93,7 +93,7 @@ class StochasticSolveFixedStepIntegrator(
         return self.method.dt
 
     @abstractmethod
-    def sample_rv(self, key: PRNGKeyArray, nsteps: int) -> Array:
+    def sample_rv(self, key: PRNGKeyArray, nsteps: int) -> PyTree:
         pass
 
     @abstractmethod
@@ -103,7 +103,7 @@ class StochasticSolveFixedStepIntegrator(
         pass
 
     @abstractmethod
-    def sde_y0(self, key: PRNGKeyArray) -> SDEState:
+    def sde_y0(self) -> SDEState:
         # define initial SDE state
         pass
 
@@ -166,8 +166,7 @@ class StochasticSolveFixedStepIntegrator(
 
         # === initial state
         # define initial SDE state
-        key0, key = jax.random.split(self.key)
-        sde_y0 = self.sde_y0(key0)
+        sde_y0 = self.sde_y0()
         # save initial SDE state at time 0
         saved0 = self.save(sde_y0)
 
@@ -179,7 +178,7 @@ class StochasticSolveFixedStepIntegrator(
             return (t, y), self.save(y)
 
         # split the key for each save interval
-        keys = jax.random.split(key, nsave)
+        keys = jax.random.split(self.key, nsave)
         ylast, saved = jax.lax.scan(outer_step, (self.t0, sde_y0), keys)
         ylast = ylast[1]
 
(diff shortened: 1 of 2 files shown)


dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
```diff
@@ -214,21 +207,36 @@ class JumpSolveFixedStepIntegrator(StochasticSolveFixedStepIntegrator):
...
         saved = super().postprocess_saved(saved, ylast.state)
-        return JumpSolveSaved(saved.ysave, saved.extra, saved.Esave, ylast.clicktimes)
+
+        # convert array of click value at each time to array of clicktimes, for example:
+        #   times = [0, 10, 20, 30, 40, 50]
+        #   clicks = [0, 1, 0, 1, 0, 2]
+        #   => clicktimes = [[10, 30, nan, ...], [50, nan, nan, ...]]
+        clicktimes = jnp.full((self.nmeas, self.options.nmaxclick), jnp.nan)
+        for jump_idx in jnp.arange(self.nmeas):
+            times = jnp.arange(self.t0, self.t1, self.dt)
```
gautierronan: There's sometimes weird stuff with the length of `jnp.arange` when `(self.t1 - self.t0) / self.dt` is exactly an integer. Might be safer to write `times = self.t0 + jnp.linspace(nsteps) * self.dt` here, to make sure the length is always `nsteps`.

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
```diff
@@ -214,21 +207,36 @@ class JumpSolveFixedStepIntegrator(StochasticSolveFixedStepIntegrator):
...
         return super().save(y.state)
 
     def postprocess_saved(self, saved: Saved, ylast: PyTree) -> Saved:
         saved = super().postprocess_saved(saved, ylast.state)
-        return JumpSolveSaved(saved.ysave, saved.extra, saved.Esave, ylast.clicktimes)
+
+        # convert array of click value at each time to array of clicktimes, for example:
+        #   times = [0, 10, 20, 30, 40, 50]
+        #   clicks = [0, 1, 0, 1, 0, 2]
+        #   => clicktimes = [[10, 30, nan, ...], [50, nan, nan, ...]]
```
gautierronan, line 235: I don't quite understand this: how can `clicks[-1] == 2`? Shouldn't it always be 0 or 1? And how do you know which jump operator the click is associated to?

gautierronan (approved):
That's a crazy speedup indeed. Definitely something to be very careful about when using jax.lax.cond!

Two comments below, but mostly related to the fact that I don't understand the code of this fixed step stochastic class very well. Not blocking merging if you think they're not relevant.

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
pierreguilmin: For this specific implementation, it's easier to have a single `clicks` array which labels the jump channel that clicked (0 = no click, 1 = first jump operator detector clicked, 2 = second jump operator detector clicked, etc.). We convert it back to `clicktimes` in the end.

PR #974, opened 2025-06-04
Add `prefactor()` method to time-qarray

pierreguilmin:
This helps debugging the `clip()` API and the incoming `shift()` API. It also makes the doc a bit more lively. 😊


gautierronan (approved):
Very nice! Nothing to add

PR #975, opened 2025-06-04
Add CompositeQArray structure

Piwakk:


pierreguilmin:
Very cool! I've worked a bit on top for simple things. The following example now works 🚀

```python
import jax.numpy as jnp
import dynamiqs as dq

terms = [
    [jnp.ones((3, 2, 2)), jnp.ones((3, 3))],
    [jnp.zeros((4, 3, 2, 2)), jnp.zeros((3, 3))],
]

dq.CompositeQArray((2, 3), False, terms).shape
```

Issue #978, opened 2025-06-06
`eye_like` does not respect the `layout` of the passed `QArray`


dkweiss31:
```python
dq.eye_like(dq.sigmax(layout=dq.dense)).layout # dia
```
Of course if I pass a layout it should override the type of the passed object, but otherwise it should respect the type of that object.

I'm happy to submit a PR to fix this issue, though I'm not an expert on the QArray internals so I thought I'd flag this before trying something :)

abocquet:
Hello @dkweiss31 , thank you for the headsup!

https://github.com/dynamiqs/dynamiqs/pull/979 should fix it soon :)

gautierronan:
I'm not convinced that's actually the behaviour we want. I don't quite see a use-case for which users would like the identity matrix explicitly not to be sparse, and even if that's the case, there's still the optional `layout` argument that can be passed. At least that's the reasoning behind the way we wrote `eye_like`: it only copies the `dims` of the input, but does not necessarily respect its `layout`.

Could you share your use-case @dkweiss31 ?

dkweiss31:
I was expecting the behavior of this function to be similar to e.g. numpy's `zeros_like` or `ones_like`. Indeed, the word `like` implies to me same type and layout, unless explicitly told otherwise. For instance, the following code fails

```python
import jax
import dynamiqs as dq

sigmas = dq.asqarray([dq.sigmax(layout=dq.dense), dq.sigmay(layout=dq.dense)])

def step(carry, x):
    x_next = x @ carry
    return x_next, x_next

init = dq.eye_like(sigmas[0])
result, accum_result = jax.lax.scan(step, init, sigmas)
```
Replacing the `init` declaration with the following allows it to work
```python
init = dq.eye_like(sigmas[0], layout=dq.dense)
```
However, it does not feel appropriate to have to specify the layout at this point in the code: it should be implicit from the matrices that are passed in. In particular, I should be able to use the same code with instead sparse `sigmas`. I could do something like `layout=sigmas[0].layout`, but a solution where the layout is implied from the input feels nicer.

PR #979, opened 2025-06-06
Fix `dq.eye_like()` and `dq.zeros_like()` default layout

abocquet:
Fixes https://github.com/dynamiqs/dynamiqs/issues/978 by forcing same layout as argument by default


dynamiqs/utils/operators.py
```diff
@@ -104,7 +104,8 @@ def eye_like(
         dims _(tuple of ints or None)_: Dimensions of each subsystem in the composite
             system Hilbert space tensor product. Defaults to `None` (`x.dims` if
             available, single Hilbert space `dims=(n,)` otherwise).
-        layout _(dq.dense, dq.dia or None)_: Matrix layout.
+        layout _(dq.dense, dq.dia or None)_: Matrix layout. By default, the layout is
+            the same is `x`
```
gautierronan: I quite like the phrasing of [`np.zeros_like`](https://numpy.org/doc/2.2/reference/generated/numpy.zeros_like.html) here, something like the following maybe?
```
Overrides the matrix layout of the result. If `None`, the layout is the same as `x`.
```

gautierronan (changes requested):
Thanks for the fix! Since we're changing the behaviour of the function, there's a few places in the backend that need updating too, namely [here](https://github.com/dynamiqs/dynamiqs/blob/b3a2ad9ef4b8d6315e2753690335c0afbf2129c2/dynamiqs/integrators/core/rouchon_integrator.py#L113), [here](https://github.com/dynamiqs/dynamiqs/blob/b3a2ad9ef4b8d6315e2753690335c0afbf2129c2/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py#L245) and [here](https://github.com/dynamiqs/dynamiqs/blob/b3a2ad9ef4b8d6315e2753690335c0afbf2129c2/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py#L496).

gautierronan:
> @gautierronan I think using the `x` matrix layout is the sane default in the backend too no? (otherwise what would be the good choice?).

I would have argued that we should always use DIA for identities in the backend, no? I don't think there's any advantage in doing dense, although it's only a small overhead anyway. Let's keep it like that for now, it doesn't change much and will be changed anyway as you say.

gautierronan (approved):

gautierronan:
Thanks @abocquet @pierreguilmin !

PR #983, opened 2025-06-11
Add tracing tests for general utilities

gautierronan:
I think we should also write similar tests for all other utilities, to systematically check that we're writing JIT-compatible utility functions. I can propose a PR after this one is merged.

I advise reviewing commit by commit.


dynamiqs/qarrays/dense_qarray.py
```diff
@@ -62,6 +62,7 @@ def ptrace(self, *keep: int) -> QArray:
 
         return ptrace(self.data, keep, self.dims)
 
+    @partial(jax.jit, static_argnums=(1,))
```
pierreguilmin: Why doing this here? We don't jit functions by default for now. Same question for the other `powm` implementations.

pierreguilmin (changes requested):
Amazing! Great PR, thanks. Small question.

pierreguilmin:
Refactoring our tests has been a long-awaited task on my to-do list! Regarding the general organization, I think we should ideally:
- Ensure that the `tests/` directory structure matchess the `dynamiqs/` directory structure
- Have one test function for each function in a file. For example, `tests/utils/test_general.py` should have a `test_dag()` function.
- test the tracing in function testing, e.g. `jax.jit(dq.dag).trace(x)` should be in `test_dag()`, not gathered with all the other functions (because tests should fail independently, and because when one adds a function and a corresponding test it should be at one specific place, not having to remember to add some test elsewhere)
- Include tracing test in function-specific tests. For instance, `jax.jit(dq.dag).trace(x)` should be in `test_dag()`, rather than grouped with other functions. This is because tests should fail independently, and when adding a test for a new function, one should not have to remember to modify another test function.

pierreguilmin:
> I think we should also write similar tests for all other utilities, to systematically check that we're writing JIT-compatible utility functions.

100% agree with this!

dynamiqs/qarrays/dense_qarray.py
gautierronan: Yes, I initially thought that adding this here would force the argument to be static. But if there's an outer jax.jit call (just like in the tests), then this is essentially ignored and still fails. I'll remove it.

gautierronan:
Yes, I quite agree with the test refactoring you're proposing. I'll start by writing the same tracing tests for other utilities, and then in a second PR I'll write this refactoring at least for utility functions to start with.

PR #990, opened 2025-06-17
Add `dtmax` to fix `Event` method click precision

pierreguilmin:
Example of the error on the first click time statistics between `0` and `1` when not specifying `dtmax` (sorry the plot is a bit disgusting 😅):

5ca7d1b2-151e-4ba1-9be1-5867d6a0cd73

@gautierronan @dkweiss31 should we make this an error rather than a warning?

<details>
  <summary>Code</summary>

```python
import dynamiqs as dq
import jax.numpy as jnp
import jax
from matplotlib import pyplot as plt
import optimistix

dq.plot.mplstyle(dpi=150)
dq.set_precision('double')

H = dq.zeros(2)
jump_ops = [dq.sigmam()]
psi0 = dq.excited()

keys = jax.random.split(jax.random.PRNGKey(42), 1000)

method = dq.method.Event(root_finder=optimistix.Newton(rtol=1e-5, atol=1e-5))
result = dq.jssesolve(H, jump_ops, psi0, tsave, keys, method=method)
first_click_time = result.clicktimes[:, 0, 0]
plt.hist(first_click_time, bins=50, alpha=0.2, label=f'ground truth')

for dtmax in [None, 1e-2, 1e-3]:
    method = dq.method.Event(dtmax=dtmax)
    result = dq.jssesolve(H, jump_ops, psi0, tsave, keys, method=method)
    first_click_time = result.clicktimes[:, 0, 0]
    plt.hist(first_click_time, bins=50, alpha=0.2, label=f'dtmax={dtmax}')

plt.xlabel('click time')
plt.ylabel('count')
plt.legend()
plt.ylim(0, 30)
```
</details>

### DIFF
--- a/tests/jssesolve/test_batching.py
+++ b/tests/jssesolve/test_batching.py
@@ -32,7 +32,8 @@ def test_cartesian_batching(nH, npsi0, nL):
     H, Ls, psi0, Es, kmc = rand_jssesolve_args(n, nH, nLs, npsi0, nEs)
     keys = jax.random.split(kmc, num=ntrajs)
     tsave = jnp.linspace(0.0, 0.1, ntsave)
-    result = dq.jssesolve(H, Ls, psi0, tsave, keys=keys, exp_ops=Es)
+    method = dq.method.Event(dtmax=1e-1)
+    result = dq.jssesolve(H, Ls, psi0, tsave, keys=keys, exp_ops=Es, method=method)
 
     # check result shape
     assert result.states.shape == (*nH, *nL, *npsi0, ntrajs, ntsave, n, 1)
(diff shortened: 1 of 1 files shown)


gautierronan (changes requested):
Yes, I think it should be an error and not a warning. It seems too big of a footgun otherwise. The PR looks good apart from this!

dkweiss31:
I think I'm mildly in favor of what you have now with a warning? This way we alert the user to the possible issue without raising the bar for the number of things they have to pass to `jssesolve` in order to get results. Either way this is a great idea and I am happy to be overruled and to have this be an error :)

dkweiss31 (dismissed):

PR #998, opened 2025-06-30
Avoid nesting composite qarray terms when using `__and__`

Piwakk:
Do not nest terms in `CompositeQArray.__and__()`


pierreguilmin (approved):
Great!

PR #1005, opened 2025-07-14
Rename gradients

pierreguilmin:


pierreguilmin:
> No problem with the PR, but could you explain why it's a better naming? Direct is also a backward method, right?

No, AFAIK direct is JAX's automatic differentiation ([DirectAdjoint](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.DirectAdjoint) in Diffrax), so it supports both fwd and bwd differentiation (needed to compute Hessians for example). Autograd is the name of specific libraries (https://github.com/hips/autograd). It doesn't appear in JAX's doc (https://docs.jax.dev/en/latest/automatic-differentiation.html), but it is mentionned in the advanced autodiff tutorial (https://docs.jax.dev/en/latest/advanced-autodiff.html)

> Interestingly, the [Autograd](https://github.com/hips/autograd) library couldn’t do this. The [implementation](https://github.com/HIPS/autograd/blob/96a03f44da43cd7044c61ac945c483955deba957/autograd/differential_operators.py#L60) of reverse-mode jacobian in Autograd had to pull back one vector at a time with an outer-loop map. Pushing one vector at a time through the computation is much less efficient than batching it all together with [jax.vmap()](https://docs.jax.dev/en/latest/_autosummary/jax.vmap.html#jax.vmap).

> Another thing that Autograd couldn’t do is [jax.jit()](https://docs.jax.dev/en/latest/_autosummary/jax.jit.html#jax.jit). Interestingly, no matter how much Python dynamism you use in your function to be differentiated, we could always use [jax.jit()](https://docs.jax.dev/en/latest/_autosummary/jax.jit.html#jax.jit) on the linear part of the computation.

So I'm convinced `Autograd` is a bad naming, and `Direct` I'm not too sure, I'm open to suggestions. 😊

PR #1006, opened 2025-07-14
Composite qarray symbolic data representation when printing

pierreguilmin:
```python
import jax.numpy as jnp
import dynamiqs as dq

na, nb = 16, 8
a, b = dq.destroy(na, nb)
H = dq.CompositeQArray((na, nb), False, [(a @ a, b.dag()), (a.dag() @ a.dag(), b)])
H
```
```
QArray: shape=(16384, 16384), dims=(16, 8), dtype=complex64, type=composite, nterms=2
O = Oa1 ⊗ Ob1 + Oa2 ⊗ Ob2
```

### DIFF
--- a/dynamiqs/qarrays/composite_qarray.py
+++ b/dynamiqs/qarrays/composite_qarray.py
@@ -221,14 +221,13 @@ class CompositeQArray(QArray):
     def __repr__(self) -> str:
         # create string representation, e.g. O = Oa1 ⊗ Ob1 + Oa2 ⊗ Ob2
         alphabet = 'abcdefghijklmnopqrstuvwxyz'
-        factors_str = lambda idx_term: [
-            f'O{alphabet[i]}{idx_term + 1}' for i in range(len(self.dims))
-        ]
-        terms_str_list = [' ⊗ '.join(factors_str(i)) for i in range(self.nterms)]
+        term_to_str = lambda idx: ' ⊗ '.join(
+            f'O{alphabet[i]}{idx + 1}' for i in range(len(self.dims))
+        )
         if len(self.terms) <= 4:
-            terms_str = 'O = ' + ' + '.join(terms_str_list)
+            terms_str = ' + '.join(term_to_str(i) for i in range(self.nterms))
         else:
-            terms_str = f'O = {terms_str_list[0]} + ... + {terms_str_list[-1]}'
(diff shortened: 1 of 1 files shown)


gautierronan (approved):
For now, I think this representation is a bit simple to actually be useful (essentially, there's as much information in `nterms` + `ndims` as there is in your representation). However, once we introduce the Identity operator in CompositeQArrays and we update this representation accordingly, it will be a bit more interesting.

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -219,9 +219,17 @@ def block_until_ready(self) -> QArray:
         return self
 
     def __repr__(self) -> str:
+        # create string representation, e.g. O = Oa1 ⊗ Ob1 + Oa2 ⊗ Ob2
+        alphabet = 'abcdefghijklmnopqrstuvwxyz'
+        factor_str = lambda idx_term: [
+            f'O{alphabet[i]}{idx_term + 1}' for i in range(len(self.dims))
+        ]
+        terms_str = ' + '.join([' ⊗ '.join(factor_str(i)) for i in range(self.nterms)])
```
gautierronan: Suggestion: for `nterms` or `ndims` above a certain threshold (4? 5?), it would be nice to not print the whole representation, but only the first and last terms, with three dots (`...`) in between. This is much like the printing of numpy arrays, and would be more readable.

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -219,9 +219,22 @@ def block_until_ready(self) -> QArray:
...
+        factors_str = lambda idx_term: [
+            f'O{alphabet[i]}{idx_term + 1}' for i in range(len(self.dims))
+        ]
+        terms_str_list = [' ⊗ '.join(factors_str(i)) for i in range(self.nterms)]
+        if len(self.terms) <= 4:
+            terms_str = 'O = ' + ' + '.join(terms_str_list)
+        else:
+            terms_str = f'O = {terms_str_list[0]} + ... + {terms_str_list[-1]}'
+
+        data_str = f'O = {terms_str}'
```
gautierronan, line 232: Looks good! But the `O` appears twice now :D

Also, if I really want to be annoying, we could compute the `terms_str_list` only for the first and last terms when `len(self.terms) > 4`.

```suggestion
        data_str = f'{terms_str}'
```

PR #1007, opened 2025-07-29
Restrict JAX version to <0.7

pierreguilmin:
See https://github.com/jax-ml/jax/releases/tag/jax-v0.7.0. Especially
> The minimum Python version is now 3.11.


pierreguilmin:
> LGTM. Is it indeed breaking basic examples for us? Or rather more advanced stuff?

I guess it's breaking diffrax and/or equinox. I haven't tested yet though.

PR #1008, opened 2025-07-31
Update the style of dq.plot.mplstyle

gautierronan:
Proposal to update the style of dq.plt.mplstyle. I usually end up doing all these changes by hand everytime I want to make a clean-looking plot to share. I think the proportions are better, and I really like having a thin grid by default to see how the curves compare (easier to read).

I also changed the default colorset (only left the first three colors intact) because I found the previous colorset hard to work with (sometimes not good-looking, sometimes hard to distinguish between darker colors).

Before/after:
before2
after2


pierreguilmin (changes requested):
Very nice! I've added a few fixes on top.

Could we choose different colors for 4 (gray) and 5 (black-ish)? I have the feeling these colors are often used in contrast to more colorfull ones, see e.g. the plot in jump tutorial. Ofc we could change this specific plot, but generically I think it is useful not to include gray and black in the base color palette, so one can add lines that stand out.

Image

gautierronan:
Here's a new iteration of colors. I tried to have contrasted colors that still go reasonably well together, and to arrange them such that they fit 2 by 2 and 3 by 3. I updated the plots in the initial PR description. Let me know what you think!

PR #1014, opened 2025-09-03
Fix dq.vectorize and dq.unvectorize from DIA inputs

gautierronan:
Fixes
```python
dq.vectorize(dq.destroy(3))
```
returning a square matrix instead of a vector.

### DIFF
--- a/dynamiqs/utils/vectorization.py
+++ b/dynamiqs/utils/vectorization.py
@@ -5,6 +5,7 @@ from dataclasses import replace
 import numpy as np
 
 from .._checks import check_shape
+from ..qarrays.layout import dense
 from ..qarrays.qarray import QArray, QArrayLike
 from ..qarrays.utils import asqarray
 from .general import dag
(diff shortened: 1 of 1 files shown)


abocquet (approved):

dynamiqs/utils/vectorization.py
```diff
@@ -87,10 +88,10 @@ def unvectorize(x: QArrayLike) -> QArray:
         [[1.+1.j 3.+3.j]
          [2.+2.j 4.+4.j]]
     """
-    x = asqarray(x)
+    x = asqarray(x, layout=dense)
     check_shape(x, 'x', '(..., n^2, 1)')
     bshape = x.shape[:-2]
-    n = int(np.sqrt(x.shape[-2]))
+    n = round(np.sqrt(x.shape[-2]))
```
pierreguilmin, line 94: `round` returns a float.

abocquet: It returns an int
```python
>>> round(3.4).__class__
<class 'int'>
```

pierreguilmin: Wouuups, indeed sorry 😅

PR #1015, opened 2025-09-18
Support static tsave to JIT stochastic solvers

rsnegrin:
Fix the issue where `tsave` was converted to an array and then to a tuple, which made stochastic solvers non-JIT compatible. The temporary fix is to bypass this conversion when `tsave` is explicitly passed as a tuple. Note that if a tuple is passed, we do not perform checks on `tsave`.

<details>
  <summary>Old description</summary>

## Summary
Make `tsave` a **static** argument (accepting `tuple[float]`) so `dssesolve` can be JIT-compiled without hash errors.

## What changed
- Accept `tuple[float]` for `tsave`.
- New static validation `check_times_static(tsave)`.
- Minor typing/doc cleanups.

## Why
Prevents `ValueError: Non-hashable static arguments` and allows JIT’ing `dssesolve`.

## Backwards compatibility
- **Non-JIT calls still accept `ArrayLike`** for `tsave`; it’s converted internally (e.g., `jnp.ndarray`/`list` → `tuple`) before calling the jitted wrapper.
- No behavioral change to the solver.

## Usage
```python
# jit compile with tsave marked static
jitted_dssesolve = jax.jit(dq.dssesolve, static_argnames=("tsave",))
jitted_eqx_dssesolve = equinox.filter_jit(dq.dssesolve) # No static argnames
```

</details>

### DIFF
--- a/dynamiqs/integrators/apis/dssesolve.py
+++ b/dynamiqs/integrators/apis/dssesolve.py
@@ -6,7 +6,7 @@ import jax
 import jax.numpy as jnp
 from jaxtyping import ArrayLike, PRNGKeyArray
 
-from ..._checks import check_shape, check_times, check_times_static
+from ..._checks import check_shape, check_times_static
 from ...gradient import Gradient
 from ...method import EulerMaruyama, Method, Rouchon1
 from ...options import Options, check_options
--- a/dynamiqs/integrators/apis/dssesolve.py
+++ b/dynamiqs/integrators/apis/dssesolve.py
@@ -4,7 +4,6 @@ from functools import partial
 
 import jax
 import jax.numpy as jnp
-from jax import Array
 from jaxtyping import ArrayLike, PRNGKeyArray
 
 from ..._checks import check_shape, check_times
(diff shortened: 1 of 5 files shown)


dynamiqs/integrators/apis/dssesolve.py
```diff
@@ -258,35 +258,37 @@ def dssesolve(
...
     _check_dssesolve_args(H, Ls, psi0, exp_ops)
-    tsave = check_times(tsave, 'tsave')
-    check_options(options, 'dssesolve')
+    check_times(jnp.asarray(tsave), "tsave")  # keep?
+    check_options(options, "dssesolve")
     options = options.initialise()
 
+    # allows ArrayLike objects when non jitted
+    if not isinstance(tsave, tuple):
+        tsave = tuple(tsave.tolist())
```
rsnegrin: Maybe 
`tsave = tuple(jnp.array(tsave).tolist())`

pierreguilmin (commented):
Thanks a lot @rsnegrin! Here's a proposal for a less complete but simpler fix, as the missing support for non-static tsave is just a temporary bug anyway. 😊

gautierronan (approved):
LGTM! Thanks @rsnegrin and @pierreguilmin

PR #1073, opened 2026-03-23
Cache decompositions in low rank integrator

p-pietro:
This PR caches the QR decomposition or pseudoinverse in `MESolveLowRankIntegrator`, so that these expensive operations are performed only once per evaluation of the field. This leads to a significant performance improvement when many jump operators are present.

To accomplish this while keeping AD compatibility, the dependency on `lineax` has been dropped, while reimplementing the relevant rules directly.

### DIFF
--- a/dynamiqs/integrators/core/low_rank_integrator.py
+++ b/dynamiqs/integrators/core/low_rank_integrator.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import warnings
 from functools import partial
+from typing import Any
 
 import diffrax as dx
 import equinox as eqx
 import jax
 import jax.numpy as jnp
-import lineax as lx
+import jax.scipy.linalg as jsp_linalg
 from jax import Array
 from jaxtyping import PyTree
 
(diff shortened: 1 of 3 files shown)


dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -22,20 +23,74 @@
...
+    primals: tuple[Any, ...], tangents: tuple[Any, ...]
+) -> tuple[Array, Array]:
+    A, cache, B = primals
+    dA, _dcache, dB = tangents
+    x = _qr_solve_cached(A, cache, B)
+    # Pseudoinverse JVP formula (independent columns):
+    # dX = A^+ (dB - dA X + (A^+)^H (dA^H (B - A X))).
+    res = B - A @ x
+    dvec = dB - dA @ x + _qr_pinv_h_apply(cache, dA.conj().T @ res)
+    dx = _qr_pinv_apply(cache, dvec)
```
Nicolas-Lepage: this has the same name as `import diffrax as dx`, even if this causes no issue in this file we should rename it

dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -22,20 +23,74 @@
...
+@jax.custom_jvp
+def _cholesky_solve_cached(A: Array, A_pinv: Array, B: Array) -> Array:
+    del A
     return A_pinv @ B
 
 
+@_cholesky_solve_cached.defjvp
+def _cholesky_solve_cached_jvp(
+    primals: tuple[Any, ...], tangents: tuple[Any, ...]
+) -> tuple[Array, Array]:
```
Nicolas-Lepage, line 99: i think the typing can be tightened here with something like 

```
def _cholesky_solve_cached_jvp(
    primals: tuple[Array, Array, Array],
    tangents: tuple[Array, Array, Array],
) -> tuple[Array, Array]: ...
```

dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -22,20 +23,74 @@
...
+@jax.custom_jvp
+def _qr_solve_cached(A: Array, cache: tuple[Array, Array], B: Array) -> Array:
+    del A
+    return _qr_pinv_apply(cache, B)
+
+
+@_qr_solve_cached.defjvp
+def _qr_solve_cached_jvp(
+    primals: tuple[Any, ...], tangents: tuple[Any, ...]
+) -> tuple[Array, Array]:
```
Nicolas-Lepage, line 74: I think the typing can be tightened here: 

```
def _qr_solve_cached_jvp(
    primals: tuple[Array, tuple[Array, Array], Array],
    tangents: tuple[Array, tuple[Array, Array], Array],
) -> tuple[Array, Array]: ...
```

Nicolas-Lepage (changes requested):
Hi @p-pietro! Thanks for the contribution, and sorry for taking a bit of time to review :)

The method looks good, I have a few comments before merging.

Style (non-scientific):
- `dx` is used as a local variable in both JVP rules, shadowing `import diffrax as dx`. Harmless here, but please rename it.
- Could you rename some of the shorthand notations , e.g. `res`, `dvec`, `pinv_h`, `Lm`... 
-`Any` can be replaced with stricter typing. 

tests : 
The existing low-rank tests do cover gradient accuracy against analytic solutions, but only for the default linear solver. 
- Can you arametrize `test_correctness` and `test_gradient` over `LinearSolver.QR` and `LinearSolver.CHOLESKY`

AD :
We're adding higher-order differentiation support to the library and will want to extend it to the low-rank integrator eventually. The cached rules are correct at first order but not at second. could you add a TODO comment to later fix compatibility ?

p-pietro:
Hello @Nicolas-Lepage, thanks for the comments! I should have now addressed all of them.

Nicolas-Lepage (approved):
Thanks @p-pietro for the changes! Everything looks good to me.

@gautierronan, would you like to take a look as well before we merge?

gautierronan (approved):
LGTM! Thanks @p-pietro

PR #1075, opened 2026-04-12
Add typing overloads for destroy, create, and number

### LINKED ISSUE #1003: Typing overloads for methods with single/multiple args
Currently, methods (like `destroy`) have a general return type:

```python
def destroy(*dims: int) -> Array | tuple[Array, ...]:
    ...
````

This causes a typing error when using the result:

```python
dq.tensor(
    dq.eye(10),
    dq.destroy(10),  # typing error here
)
```

Add two overloads for precise typing:

```python
from typing import overload

@overload
def destroy(dims: int, /) -> Array: ...
@overload
def destroy(*dims: int) -> tuple[Array, ...]: ...

def destroy(*dims: int) -> Array | tuple[Array, ...]:
    ...
```

Ideally, place these stubs in a `.pyi` file.

hijirii:
## Summary

Add @overload decorators to destroy(), create(), and number() functions so that type checkers understand the return type depends on the number of arguments passed.

## Fixes
Fixes https://github.com/dynamiqs/dynamiqs/issues/1003


abocquet:
Hello @hijirii 

Thank you for this proposal! The modification looks good to me. I'll wait for a second opinion from @gautierronan and then we can merge

EDIT: I see the CI fails, probably because of new checks introduced in the latest `ty` version. We'll fix them to make this MR pass first

gautierronan (approved):
LGTM as well! Thanks for fixing!

gautierronan:
Can merge after https://github.com/dynamiqs/dynamiqs/pull/1076 is merged.

PR #1084, opened 2026-05-19
Optimize PWC prefactor evaluation 

Nicolas-Lepage:
## Summary

This PR optimizes `PWCTimeQArray._prefactor()` by replacing the `jnp.searchsorted`-based interval lookup with a vectorized interval-mask reduction.

The public `times` representation is unchanged: `times` is still stored as a flat array of shape `(N + 1,)`, so existing user code that accesses `x.times` keeps the same behavior. Internally, the interval view is reconstructed through a private `_times_reshaped` property and used to compute the active interval mask.

## What changed

- Added a private `_times_reshaped` property returning consecutive time intervals from the flat `times` array.
- Replaced the previous `searchsorted` + indexed value lookup in `_prefactor()` with:
  - a boolean mask selecting the active interval,
  - a reduction over interval values.
- Kept the external `times` storage format unchanged for backward compatibility.

## Results

The benchmark compares the current `pwc` implementation against the new interval-mask implementation on CPU and GPU, for both single and 100-interval pulses. All cases agree numerically within tolerance, with maximum observable differences of order `1e-9`.

| Hardware / case | Implementation | Compile time (s) | Execution time (s) | Result |
|---|---:|---:|---:|---|
| GPU, 1 interval | `pwc` | 5.9449 | 1.1839 | baseline |
| GPU, 1 interval | new | **4.0352** | **1.1798** | slightly faster compile, same runtime |
| GPU, 100 intervals | `pwc` | 5.0725 | 3.9727 | baseline |
| GPU, 100 intervals | new | **5.0406** | **1.2530** | **~3.2x faster execution** |
| CPU, 1 interval | `pwc` | 3.1095 | **7.8539** | baseline |
| CPU, 1 interval | new | **2.5600** | 8.0198 | similar runtime |
| CPU, 100 intervals | `pwc` | 3.0377 | 9.6702 | baseline |
| CPU, 100 intervals | new | **2.8817** | **7.9995** | **~17% faster execution** |
| GPU batch 128, 1 interval | `pwc` | 19.7940 | 27.9118 | baseline |
| GPU batch 128, 1 interval | new | **17.7853** | **25.9049** | modest improvement |
| GPU batch 128, 100 intervals | `pwc` | 18.3970 | 33.4496 | baseline |
| GPU batch 128, 100 intervals | new | **18.2105** | **27.1799** | **~19% faster execution** |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ dev = [
     "flit",
     "mike",
     "optax",  # for gradient descent tutorials
-    "ty==0.0.17",
+    "ty",
 ]
 
 [tool.ruff]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -139,60 +138,23 @@ unused-ignore-comment = "error"
...
+docbuild = 'mkdocs build'
+docserve = 'mkdocs serve'
+all = 'task clean && task test && task doctest'
+ci = '''echo "\n>>> ruff check" && ruff check &&
+        echo "\n>>> ruff format --check" && ruff format --check &&
+        task codespell &&
+        task type &&
+        task test &&
+        task doctest &&
+        task docbuild'''
```
gautierronan: Why did you remove the `help` variables throughout tasks? I don't think it's related to the problem you're trying to solve, so it should live in a separate PR

gautierronan (dismissed):
Great PR, great benchmarks, great work! LGTM!

I think we have an issue with the CI related to `ty` however. It probably needs to be fixed in a separate PR before we can merge this.

gautierronan (approved):

PR #1086, opened 2026-05-20
Fix ClipStepSizeController replace() crash with discontinuous Hamiltonians

### LINKED ISSUE #1085: Rouchon3() (adaptive) crashes when Hamiltonian has discontinuities
### Description
`Rouchon3()` adaptive crashes with `TypeError: ClipStepSizeController.__init__() got an unexpected keyword argument 'callback_on_reject'` when Hamiltonian has discontinuities.

`dq.mesolve` with `method=Rouchon3()` (adaptive step size) raises a `TypeError` whenever the Hamiltonian has discontinuities.
With a constant Hamiltonian it works fine.

**Environment**

| package  | version |
|----------|---------|
| dynamiqs | 0.3.5   |
| diffrax  | 0.7.1   |
| equinox  | 0.13.4  |
| jax      | 0.10.0  |

## MWE
```
import jax
import jax.numpy as jnp
import equinox as eqx
import dynamiqs as dq

rho0  = dq.fock_dm(2, 0)
tsave = jnp.linspace(0.0, 1.0, 50)
L     = [0.1 * dq.sigmam()]

# Piecewise-constant Hamiltonian with a discontinuity at t = 0.5
H = dq.pwc(
    times=jnp.array([0.0, 0.5, 1.0]),
    values=jnp.array([1.0, 0.5]),
    qarray=dq.sigmax(),
)

@eqx.filter_jit
def run():
    return dq.mesolve(H, L, rho0, tsave, method=dq.method.Rouchon3())

run()
# TypeError: ClipStepSizeController.__init__() got an unexpected keyword
# argument 'callback_on_reject'

# Both of these work fine:
dq.mesolve(dq.sigmaz(), L, rho0, tsave, method=dq.method.Rouchon3())  # constant H
dq.mesolve(H,           L, rho0, tsave, method=dq.method.Tsit5())     # different solver
```

rsnegrin:
Fixes #1085

In diffrax 0.7, `ClipStepSizeController` renamed its constructor parameter from
`callback_on_reject` to `_callback_on_reject`, but the dataclass field kept the old
name. This caused `dataclasses.replace(stepsize_controller, step_ts=...)` in
`MESolveAdaptiveRouchonIntegrator.stepsize_controller` to raise a `TypeError` whenever
the Hamiltonian had discontinuities (which wraps the controller in a
`ClipStepSizeController`).

**Fix:** detect `ClipStepSizeController` and reconstruct it directly instead of using
`replace()`.

**Test:** added `test_pwc_hamiltonian` to verify `Rouchon2`/`Rouchon3` no longer crash
with a PWC Hamiltonian.

There may be cleaner ways of fixing this.

### DIFF
--- a/dynamiqs/integrators/core/rouchon_integrator.py
+++ b/dynamiqs/integrators/core/rouchon_integrator.py
@@ -546,6 +546,13 @@ class MESolveAdaptiveRouchonIntegrator(
         stepsize_controller = super().stepsize_controller
         # fix incorrect default linear interpolation by stepping exactly at all times
         # in tsave, so interpolation is bypassed
+        if isinstance(stepsize_controller, dx.ClipStepSizeController):
+            return eqx.tree_at(
+                lambda c: c.step_ts,
+                stepsize_controller,
+                self.ts,
+                is_leaf=lambda x: x is None,
+            )
         return replace(stepsize_controller, step_ts=self.ts)
 
 
--- a/dynamiqs/integrators/core/diffrax_integrator.py
+++ b/dynamiqs/integrators/core/diffrax_integrator.py
@@ -113,20 +113,19 @@ class DiffraxIntegrator(BaseIntegrator, AbstractSaveMixin, AbstractTimeInterface
             # === prepare stepsize controller
             stepsize_controller = self.stepsize_controller
             if dtmax is not None:
-                if isinstance(stepsize_controller, dx.ClipStepSizeController):
-                    stepsize_controller = eqx.tree_at(
-                        lambda c: c.controller.dtmax,
-                        stepsize_controller,
-                        dtmax,
-                        is_leaf=lambda x: x is None,
-                    )
-                else:
-                    stepsize_controller = eqx.tree_at(
-                        lambda c: c.dtmax,
-                        stepsize_controller,
-                        dtmax,
-                        is_leaf=lambda x: x is None,
-                    )
+                # Diffrax's ClipStepSizeController and PIDController have different
+                # APIs for setting dtmax, so we need to handle them separately
+                get_dtmax = (
+                    (lambda c: c.controller.dtmax)
+                    if isinstance(stepsize_controller, dx.ClipStepSizeController)
+                    else (lambda c: c.dtmax)
+                )
+                stepsize_controller = eqx.tree_at(
+                    get_dtmax,
+                    stepsize_controller,
+                    dtmax,
+                    is_leaf=lambda x: x is None,
+                )
 
             # === solve differential equation with diffrax
             return dx.diffeqsolve(
--- a/dynamiqs/integrators/core/diffrax_integrator.py
+++ b/dynamiqs/integrators/core/diffrax_integrator.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import warnings
 from abc import abstractmethod
-from dataclasses import replace
 from functools import partial
 
 import diffrax as dx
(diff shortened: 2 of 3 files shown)


dynamiqs/integrators/core/rouchon_integrator.py
```diff
@@ -546,6 +546,12 @@ def stepsize_controller(self) -> dx.AbstractStepSizeController:
         stepsize_controller = super().stepsize_controller
         # fix incorrect default linear interpolation by stepping exactly at all times
         # in tsave, so interpolation is bypassed
+        if isinstance(stepsize_controller, dx.ClipStepSizeController):
+            return dx.ClipStepSizeController(
+                controller=stepsize_controller.controller,
+                step_ts=self.ts,
+                jump_ts=stepsize_controller.jump_ts,
+            )
         return replace(stepsize_controller, step_ts=self.ts)
```
gautierronan, line 556: Instead of using this `isinstance` logic, you can simply use `eqx.tree_at` instead of `replace`. This should do the trick.

More specifically, you can use `return eqx.tree_at(lambda c: c.step_ts, stepsize_controller, self.ts)`.

gautierronan (changes requested):
Thanks for investigating and fixing!

Do you mind also fixing the same occurence inside `diffrax_integrator.py`? Thanks!

Nicolas-Lepage:
### Main bug review

Thanks @rsnegrin for the bug hunting and solution!

@gautierronan :
  What I understand is that the `replace()` method is not very robust to internal attribute changes, and is the cause of the issue. I did not know about the `eqx.tree_at` method suggested but it seems like a good fix.
  
  ### `diffrax_integrator` question
  
  `diffrax_integrator` does not have the same `replace()` for the `PIDController`, because the `Rouchon()` use of `replace()` is an edge case to force the solver to pass exactly through the `self.ts` points for CPTP-correct, uninterpolated outputs.
  
  That being said, `replace()` also appears in `diffrax_integrator` for a completely different reason, linked to `Event` and `dtmax` affecting the step controller. I do not understand that specific path yet, but should this `replace()` also be changed to use `eqx.tree_at`?
  
  ### Merging
  
  If you agree, I can make the changes, rebase onto the new CI changes from yesterday to make everything pass, and then approve the PR on my side.

gautierronan:
> diffrax_integrator does not have the same replace() for the PIDController, because the Rouchon() use of replace() is an edge case to force the solver to pass exactly through the self.ts points for CPTP-correct, uninterpolated outputs.
> That being said, replace() also appears in diffrax_integrator for a completely different reason, linked to Event and dtmax affecting the step controller. I do not understand that specific path yet, but should this replace() also be changed to use eqx.tree_at?

It seems that `replace` is simply less robust than `eqx.tree_at` because it takes the dataclass fields and reruns an `__init__` with these field names. So I'm guessing we'll run in the same issue inside `diffrax_integrator.py` if the `stepsize_controler` is also of type `ClipStepSizeController`. The actual field we're overriding (`dtmax` or `step_ts`) does not matter: the bug will still occur.

rsnegrin:
Hey! 

I went with `eqx.tree_at` as suggested. One thing though, `PIDController` has not `step_ts` attribute, so the `eqx.tree_at` only works when the controller is already a `ClipStepSizeController`.

I had to keep the `isinstance()` to handle this case.
I also fixed the same` replace()` in `diffrax_integrator.py` for the `dtmax` case.

Let me know if you'd like a different approach!

gautierronan (dismissed):
Thanks for fixing @rsnegrin. The behavior of diffrax is actually very strange:

- If I have a step size controller of type `PIDController` and I try to change `step_ts` through `eqx.tree_at` it doesn't work because the field doesn't exist. But if I use `replace` instead, it works and is upgraded to a `ClipStepSizeController`.
- If I have a step size controller of type `ClipStepSizeController` and I try to change `step_ts` through `eqx.tree_at` it works as intended (your fix). But if I use `replace` instead, it fails according to your bug.

That's kind of poor design by diffrax IMO.

Anyway, your fix LGTM

gautierronan (approved):

PR #1087, opened 2026-05-21
Fix CI dependency resolution for taskipy and ty typing issues

Nicolas-Lepage:
## Summary

This PR fixes two small issues that were preventing CI from running properly:

- pins/fixes the `taskipy` version issue during `uv` environment resolution
- fixes formatting issues in the `ty` task instructions

Both problems were encountered while working on #1084

## Notes

Once CI is able to run again, this also includes minor `ty`-related fixes.

### DIFF
--- a/dynamiqs/method.py
+++ b/dynamiqs/method.py
@@ -628,7 +628,7 @@ class Event(_DEMethod):
         (default).
     """
 
-    noclick_method: Method = Tsit5()  # ty: ignore[invalid-assignment]
+    noclick_method: Method = Tsit5()
     dtmax: float | None = eqx.field(static=True, default=None)
     root_finder: AbstractRootFinder | None = eqx.field(static=True, default=None)
     smart_sampling: bool = eqx.field(static=True, default=False)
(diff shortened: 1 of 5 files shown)


Nicolas-Lepage:
@abocquet and @gautierronan after trying a couple of “no free lunch” fixes, I wasn’t able to find a clean typing solution here, so I reverted the code to the original structure and only added localized `ty: ignore` comments.

The two alternatives I explored were:

* allowing `None` and `int` in the `Result` attribute types so that the `out_axes()` constructor becomes type-valid, but this weakens the actual `Result` typing quite a lot for normal usage,
* creating a separate PyTree object for the `out_axes` structure, but this does not work properly with JAX/Equinox because `out_axes` needs to match the exact `Result` PyTree structure/treedef.

So for now I kept everything as it was before the attempted typing fixes, with only narrow `# ty: ignore[invalid-argument-type]` comments on the `out_axes()` constructors, since this is really a special JAX axis-spec pattern rather than a real `Result` object.

gautierronan (approved):
Thanks for investigating. To be fair, as long as the CI passes, I'm fine with it. And it's also one of the unitaryhack issues to solve this typing issues, right?

PR #1088, opened 2026-05-27
API: flatten solver options into keyword arguments

Nicolas-Lepage:
## Summary

This PR flattens solver options in the public API.

Instead of passing an `Options` object via `options=...`, users now pass solver settings directly as keyword arguments.

This change applies to the codebase only. Documentation updates will be handled separately.

This modification only affects the public API. Internally, each solver still constructs a `dq.Options` object before passing it to lower-level integrators. This PR can therefore serve as a first step toward fully removing the `dq.Options` class in a future change, if we decide to go in that direction.

## Changes

- Removed the `options=` argument from the public API of all 9 main solver functions.
- corrected 3 documentation expales to run with the new functions 
- corrected all tests where options were passed to the solver

## Affected solvers

- `sesolve`
- `mesolve`
- `dssesolve`
- `dsmesolve`
- `jssesolve`
- `jsmesolve`
- `floquet`
- `sepropagator`
- `mepropagator`

## Additional fixes

This PR also fixes two inconsistencies in solver option handling:

- Added `progress_meter` to `mepropagator`, which was already supported by the underlying diffrax integrator.

## Breaking change

Code using `options=dq.Options(...)` with these solvers will now raise a `TypeError`.

## Examples

Before:

```
dq.mesolve(..., options=dq.Options(save_states=False))
```

After:

```
dq.mesolve(..., save_states=False)
```

### DIFF
--- a/dynamiqs/integrators/apis/mesolve.py
+++ b/dynamiqs/integrators/apis/mesolve.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
 import warnings
+from collections.abc import Callable
 
 import jax
 import jax.numpy as jnp
 from jax import Array
-from jaxtyping import ArrayLike
+from jaxtyping import ArrayLike, PyTree, ScalarLike
 
 from ..._checks import check_qarray_is_dense, check_shape, check_times
 from ...gradient import Gradient
@@ -27,6 +28,7 @@ from ...method import (
     Tsit5,
 )
 from ...options import Options, check_options
+from ...progress_meter import AbstractProgressMeter
 from ...qarrays.qarray import QArray, QArrayLike
 from ...qarrays.utils import asqarray
 from ...result import MESolveResult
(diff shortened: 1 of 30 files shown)


dynamiqs/integrators/apis/mesolve.py
```diff
@@ -284,6 +292,17 @@ def mesolve(
...
+    # === build options
+    options = Options(
+        save_states=save_states,
+        cartesian_batching=cartesian_batching,
+        progress_meter=progress_meter,
+        t0=t0,
+        save_extra=save_extra,
+        vectorized=vectorized,
+        assume_hermitian=assume_hermitian,
+    )
```
gautierronan, line 304: Because the Options dataclass is now built internally, it would be better to have a different Options class per solver, namely a `MESolveOptions`, `SESolveOptions` and so on, all inheriting from Options. This will make the internals a bit more strict, which is simpler to maintain.

This can be done in a separate PR however.

gautierronan (approved):
LGTM! I only have a comment below, but which can (probably should) be handled in a separate PR.

PR #1089, opened 2026-05-27
Docs : flatten solver options into keyword arguments

Nicolas-Lepage:
## Summary

must be reviewer after #1088

This PR updates the solver documentation to reflect the new flattened solver options API introduced in the first PR.

Solver settings are now documented as explicit keyword arguments instead of being grouped under an `Options` object.

## Changes

- Updated the API documentation for all 9 main solver functions:
  - `sesolve`
  - `mesolve`
  - `dssesolve`
  - `dsmesolve`
  - `jssesolve`
  - `jsmesolve`
  - `floquet`
  - `sepropagator`
  - `mepropagator`

- Reworked the solver docstrings to follow a clearer structure inspired by SciPy documentation:
  - Parameters
  - Results
  - Other Parameters
  - Examples

## Notes

With this change, `Options` is almost entirely removed from the public API.

One exception remains: solver result objects still expose `options` in both the code and documentation. This PR leaves that behavior unchanged, as changing it would require additional work and it is not clear yet whether removing it is necessary.

### DIFF
--- a/dynamiqs/integrators/apis/sesolve.py
+++ b/dynamiqs/integrators/apis/sesolve.py
@@ -112,21 +112,21 @@ def sesolve(
                 - **`options`** _(Options)_ - Options used.
 
     Other Parameters:
-        save_states (bool, optional): If `True`, the state is saved at every time in
+        save_states: If `True`, the state is saved at every time in
             `tsave`, otherwise only the final state is returned. Defaults to `True`.
-        cartesian_batching (bool, optional): If `True`, batched arguments are treated
+        cartesian_batching: If `True`, batched arguments are treated
             as separated batch dimensions, otherwise the batching is performed over a
             single shared batch dimension. Defaults to `True`.
-        progress_meter (AbstractProgressMeter or bool or None, optional): Progress
+        progress_meter: Progress
             meter indicating how far the solve has progressed. Defaults to `None`
             which uses the global default progress meter (see
             [`dq.set_progress_meter()`][dynamiqs.set_progress_meter]). Set to `True`
             for a [tqdm](https://github.com/tqdm/tqdm) progress meter, and `False`
             for no output. If gradients are computed, the progress meter only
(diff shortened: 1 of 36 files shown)


dynamiqs/integrators/apis/sesolve.py
```diff
@@ -147,6 +111,26 @@ def sesolve(
                 - **`gradient`** _(Gradient)_ - Gradient used.
                 - **`options`** _(Options)_ - Options used.
 
+    Other Parameters:
+        save_states (bool, optional): If `True`, the state is saved at every time in
```
gautierronan: We have this internal rule that we do not document the typing of objects when it doesn't add information compared to the function signature. The complete typing can already be found in the function signature at the top of the documentation page. So you can probably remove all the typing throughout these other parameters.

In addition, the typing inside `Other Parameters` doesn't render similarly as for the rest of the documentation. It is rendered in code instead of italic.

gautierronan (changes requested):
Thanks a lot @Nicolas-Lepage. I have a small comment below about the typing, otherwise it looks good to me

dynamiqs/integrators/apis/sesolve.py
Nicolas-Lepage: i have removed the redundant type annotations in all 9 files.

Nicolas-Lepage:
Hi Ronan 🙂
I've updated the docstring, but I also needed to add a new html.jinja template for the sections. By default, MkDocs was rendering type annotations and using code-style formatting for the section titles/content, so the custom template was required to achieve the desired rendering.

gautierronan (approved):
LGTM. Thanks!

Did you check the rendering live with a `docserve` ?

Nicolas-Lepage:
yes, i ran locally the `mkdocs serve` command and gives me this : 

Screenshot 2026-06-03 at 14 19 26

PR #1090, opened 2026-05-27
Api/expose progress meter

Nicolas-Lepage:
## Summary

This PR exposes progress meter classes in the public API and adds them to the generated API documentation.

Users can now access progress meters directly from `dynamiqs`, for example `dq.NoProgressMeter()` or `dq.TqdmProgressMeter()`.

## Changes

- Exported progress meter classes from the top-level `dynamiqs` API.
- Added docstrings for the 4 public progress meter classes:
  - `AbstractProgressMeter`
  - `NoProgressMeter`
  - `TextProgressMeter`
  - `TqdmProgressMeter`
- Added `progress_meter` to the Python API generation script.
- Added generated API pages for progress meters.
- Added a new “Progress meters” section to the API index and navigation.

### DIFF
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -95,7 +95,6 @@ The **Dynamiqs** Python API features two main types of functions: solvers of dif
         extra:
             table: true
         members:
-        - AbstractProgressMeter
         - NoProgressMeter
         - TextProgressMeter
         - TqdmProgressMeter
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -88,6 +88,18 @@ The **Dynamiqs** Python API features two main types of functions: solvers of dif
         - BackwardCheckpointed
         - Forward
 
+### Progress meters
+
+::: dynamiqs.progress_meter
+    options:
+        extra:
+            table: true
+        members:
+        - AbstractProgressMeter
+        - NoProgressMeter
+        - TextProgressMeter
+        - TqdmProgressMeter
+
 ## Utilities
 
 ### Operators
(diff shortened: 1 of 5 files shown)


docs/python_api/index.md
```diff
@@ -88,6 +88,18 @@ The **Dynamiqs** Python API features two main types of functions: solvers of dif
...
         - Forward
 
+### Progress meters
+
+::: dynamiqs.progress_meter
+    options:
+        extra:
+            table: true
+        members:
+        - AbstractProgressMeter
```
gautierronan: This one should not be in the public API, or at least not in the documented API. By definition, it is abstract and thus only used internally.

dynamiqs/progress_meter.py
gautierronan: Same comment, we should remove it from `__all__` now that we export the module

gautierronan (changes requested):

dynamiqs/progress_meter.py
gautierronan: Actually, I see that `diffrax` does export this abstractprogressmeter. We can thus decide to do it as well or not, WDYT?

Nicolas-Lepage: The difference with diffrax is that our `AbstractProgressMeter` only exposes to_diffrax(), so users can't actually define meaningful custom meters ( at east i think, to be verified ).

That said, if we wanted to support custom progress meters properly, we'd need to rework the API so that AbstractProgressMeter mirrors diffrax's interface (init, step, close), making it a real extension point. That could be a follow-up PR.

gautierronan: Ok, I agree. Then let's not bother too much with this for now because it's very niche, and we can keep it hidden from the public API.

gautierronan (approved):

PR #1091, opened 2026-05-27
feat: add CompositeQArray and CompositeTerm skeleton

Nicolas-Lepage:
## Summary

This PR introduces the initial data structures for representing composite / separable-sum qarrays.

It adds:

- `CompositeTerm`: one separable product term
- `CompositeQArray`: a lazy sum of separable product terms

No algebra, evaluation, or solver integration is implemented yet. This PR only declares the core containers, exposes them publicly, and adds API documentation.

### `CompositeTerm`

Represents one separable term:

c A₀ ⊗ A₁ ⊗ ... ⊗ Aₙ₋₁

with:

- `operators: tuple[QArray | TimeQArray, ...]`  
  One local operator per subsystem.
- `coeff: ArrayLike`  
  Scalar or batched prefactor, defaulting to `1`.

### `CompositeQArray`

Represents a lazy separable sum:

H = Σⱼ cⱼ Aⱼ,₀ ⊗ ... ⊗ Aⱼ,ₙ₋₁

with:

- `dims: tuple[int, ...]`  
  Hilbert-space dimensions of each subsystem.
- `terms: tuple[CompositeTerm, ...]`  
  The separable terms.

Both classes are exposed at the `dq.` level and documented under `Core > Composite qarrays`.

## Design note: static and time-dependent composite terms

The main design question was how to represent time dependence in composite qarrays.

Two options were considered:

### Option 1: separate static and time-dependent classes

This would mirror the current `QArray` / `TimeQArray` split, for example with:

- `CompositeTerm`
- `TimeCompositeTerm`
- possibly `CompositeQArray`
- possibly `TimeCompositeQArray`

This gives a clean type-level distinction between static and time-dependent objects, but it adds more classes and requires tensor/construction logic to choose the correct container.

### Option 2: single composite container

This PR chooses a single `CompositeTerm` / `CompositeQArray` representation where time dependence is carried by the local operators themselves:

`operators: tuple[QArray | TimeQArray, ...]`

This means a composite qarray can contain either static or time-dependent local factors. A future implementation can then expose:

`CompositeQArray.is_time_dependent() -> bool`

`CompositeQArray.__call__(t) -> CompositeQArray`

where `__call__(t)` evaluates all `TimeQArray` factors at time `t`.

The motivation is ergonomics: users should be able to write code such as:

`dq.tensor(H1, H2)`

without needing to care whether this returns a static or time-dependent composite container. The resulting `CompositeQArray` can determine internally whether it contains time-dependent terms.

Solver dispatch can then rely on `is_time_dependent()`, similarly to how static and time-dependent qarrays are currently distinguished.

### DIFF
--- a/dynamiqs/qarrays/composite_qarray.py
+++ b/dynamiqs/qarrays/composite_qarray.py
@@ -16,24 +16,14 @@ __all__ = []
 
 
 class CompositeTerm(eqx.Module):
-    r"""One separable term in a :class:`CompositeQArray`.
-
-    Represents a single term of the form
-
-    $$
-        c \, A_0 \otimes A_1 \otimes \cdots \otimes A_{N-1}
-    $$
-
-    where $c$ is a scalar coefficient and each $A_k$ is a local operator acting
-    on subsystem $k$.
+    r"""One separable term $c \, A_0 \otimes \cdots \otimes A_{N-1}$ in a
+    :class:`CompositeQArray`.  Holds the bulk of the lazy logic; most ``LAZY``
+    methods on :class:`CompositeQArray` delegate to a corresponding method here.
 
     Attributes:
-        operators: Local operators $(A_0, \ldots, A_{N-1})$, one per subsystem.
-            Each is a square :class:`MaterializedQArray` of shape
-            $(\ldots, d_k, d_k)$.
-        coeff: Scalar coefficient $c$ multiplying the full tensor-product operator.
-            Can be a Python scalar or a broadcastable JAX array for batched use.
-            Defaults to $1$.
+        operators: Per-subsystem local operators (one square :class:`MaterializedQArray`
+            per subsystem).
+        coeff: Scalar coefficient; may be a JAX array for batched use. Defaults to 1.
     """
 
     operators: tuple[MaterializedQArray, ...]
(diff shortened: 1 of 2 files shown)


dynamiqs/qarrays/composite_qarray.py
```diff
@@ -0,0 +1,64 @@
...
+            Each is a square :class:`QArray` or :class:`TimeQArray` of shape
+            $(\ldots, d_k, d_k)$. Use the factory functions :func:`composite_term`
+            or :func:`composite` to construct, which validate dimensions automatically.
+        coeff: Scalar coefficient $c$ multiplying the full tensor-product operator.
+            Can be a Python scalar or a broadcastable JAX array for batched use.
+            Defaults to $1$.
+    """
+
+    operators: tuple[QArray | TimeQArray, ...]
+    coeff: ArrayLike = 1.0
```
gautierronan: The `CompositeTerm` class could also carry the `dims` information, no? Although I guess it can be a property extracted from the `operators` attribute

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -0,0 +1,64 @@
...
+import equinox as eqx
+from jaxtyping import ArrayLike
+
+from ..time_qarray import TimeQArray
+from .qarray import QArray
+
+__all__ = ['CompositeTerm', 'CompositeQArray']
+
+
+class CompositeTerm(eqx.Module):
```
gautierronan, line 20: In this PR, could you also list all the methods that need to be implemented inside `CompositeTerm` (without actually implementing them) ? Same comment for `CompositeQArray`

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -0,0 +1,64 @@
...
+        coeff: Scalar coefficient $c$ multiplying the full tensor-product operator.
+            Can be a Python scalar or a broadcastable JAX array for batched use.
+            Defaults to $1$.
+    """
+
+    operators: tuple[QArray | TimeQArray, ...]
+    coeff: ArrayLike = 1.0
+
+
+class CompositeQArray(eqx.Module):
```
gautierronan: `CompositeQArray` should inherit from `QArray` which will then become a base class for all qarrays. We may need to change the `QArray` class to do so, and we might want to introduce a second children class for standard qarrays, e.g. `NotCompositeQArray` (need to find a better name of course).

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -0,0 +1,64 @@
...
+        operators: Local operators $(A_0, \ldots, A_{N-1})$, one per subsystem.
+            Each is a square :class:`QArray` or :class:`TimeQArray` of shape
+            $(\ldots, d_k, d_k)$. Use the factory functions :func:`composite_term`
+            or :func:`composite` to construct, which validate dimensions automatically.
+        coeff: Scalar coefficient $c$ multiplying the full tensor-product operator.
+            Can be a Python scalar or a broadcastable JAX array for batched use.
+            Defaults to $1$.
+    """
+
+    operators: tuple[QArray | TimeQArray, ...]
```
gautierronan:
```suggestion
    operators: tuple[NotCompositeQArray, ...]
```

This should not accept `TimeQArray` as input. The `TimeQArray` object is really a different kind of class that handles time-dependence using static qarrays. For instance, today it is not possible to compute the tensor product between a `TimeQarray` and a `QArray`:
```python
H1 = dq.modulated(lambda t: 1.0, dq.sigmaz())
H2 = dq.sigmax()
H = dq.tensor(H1, H2)
# TypeError: must be real number, not ModulatedTimeQArray
```
(by the way, would be nice to improve this error message)

gautierronan (changes requested):
Thanks for the PR @Nicolas-Lepage. It's very nice to be able to discuss API like this before diving into the implementation.

Regarding the time-dependence, we should really think of `CompositeQArray` as a children class of a base `QArray` class (which may need to be reworked a bit), which handles everything static. Then, time-dependence is fully handled by the `TimeQArray` class which, although it has a similar name, has nothing to do with static qarrays. This should greatly simplify the implementation, although it does slightly restrict the use-cases (not possible to take tensor products with time-qarrays).

What do you think?

Nicolas-Lepage:
Hi @gautierronan,
I've reworked this PR based on the abstraction introduced in #1101.
Key changes:
`CompositeQarray` now fully satisfies the base class contract and adds a `materialize` method. Each method includes a comment explaining its laziness strategy : whether it must materialize the full array or can be computed lazily (which sometimes depends on whether the `CompositeQarray` consists of a single term in the sum).
`CompositeTerm` retains nearly the same surface area of methods even in cases where the solution materializes, since these methods can be useful to call directly from within materialize itself.

gautierronan (approved):
Thanks for the changes @Nicolas-Lepage. I like the proposal as is, and don't have further comments!

Issue #1083, opened 2026-05-06
unitaryHACK - Build a benchmark suite for Dynamiqs solvers


Nicolas-Lepage:
### Context

Dynamiqs provides several numerical methods for solving quantum dynamics, but there is currently no standardized benchmark suite to compare them across representative problems.

This makes it difficult to:

- compare performance, accuracy, and stability between methods;
- choose the best solver for a given physical problem;
- understand scaling with Hilbert-space size, batching, time dependence, and dissipation;
- track performance and accuracy regressions as the library evolves.

This issue aims to build a reproducible benchmark suite with two goals:

1. understand the behavior of Dynamiqs numerical methods across meaningful quantum systems;
2. create a benchmark log that can be used over time to monitor changes in timing, accuracy, and solver stability.

An MVP branch with proposed benchmark problems is available here:

https://github.com/Nicolas-Lepage/dynamiqs/tree/feat/benchmark_MVP/tests/benchmark

### What we want

Design and implement a benchmark pipeline for Dynamiqs solvers.

The benchmark suite should define a small set of representative problems and run multiple numerical methods on each problem, collecting standardized metrics.

The benchmark should be reproducible and easy to run locally. It should produce a leaderboard-style summary, or at minimum save results to CSV files that can be compared across runs.

The benchmark suite should be useful both for one-off solver comparison and for long-term regression tracking.

### Scope

The benchmark should primarily cover deterministic ODE solvers and `Expm`.

Primary methods:

- `Tsit5`
- `Dopri5`
- `Dopri8`
- `Kvaerno3`
- `Kvaerno5`
- `Euler`
- `Rouchon1`
- `Rouchon2`
- `Rouchon3`
- `Expm`

Secondary methods:

- `JumpMonteCarlo`
- `DiffusiveMonteCarlo`

The `Event` sub-method is out of scope.

Stochastic/SDE solvers are a secondary objective. They are welcome if the benchmark design supports them cleanly, but the core benchmark should first work well for deterministic solvers.

### Benchmark problems

Use the MVP benchmark branch as the starting point.

The benchmark suite should include several complementary problems covering:

- closed-system Schrödinger dynamics;
- Lindblad/master-equation dynamics;
- one-mode, two-mode, and larger systems;
- time-dependent Hamiltonians;
- batched or vmapped workloads;
- realistic larger examples from Dynamiqs use cases.

Each benchmark case should define:

- Hamiltonian;
- jump operators when relevant;
- initial state;
[...]

PR #1097, opened 2026-06-04
Add opt-in solver benchmark suite

JAMA4666:
Addresses #1083.

Adds an opt-in benchmark suite for Dynamiqs solvers with representative deterministic problems, CSV output, leaderboard summary, runtime/nsteps/error metrics, and a lightweight smoke test for CI.

Author: Juan Medina <alberto4002@gmail.com>

Validation:
- py -m ruff check tests\benchmark
- py -m pytest tests\benchmark\test_smoke.py
- py -m tests.benchmark.runner --suite full --methods Tsit5 --out benchmark-full-tsit5.csv

### DIFF
--- a/tests/benchmark/metrics.py
+++ b/tests/benchmark/metrics.py
@@ -1,8 +1,27 @@
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import Any
 
 import jax.numpy as jnp
+
+import dynamiqs as dq
+
+
+@dataclass(frozen=True)
+class MetricStats:
+    mean: float
+    minimum: float
+    maximum: float
+
+
+def aggregate_stats(values: Any) -> MetricStats:
+    values = jnp.asarray(values)
+    return MetricStats(
+        mean=float(jnp.mean(values)),
+        minimum=float(jnp.min(values)),
+        maximum=float(jnp.max(values)),
+    )
 
 
 def relative_l2_error(value: Any, reference: Any) -> float:
@@ -19,7 +38,15 @@ def max_abs_error(value: Any, reference: Any) -> float:
     return float(jnp.max(jnp.abs(value - reference)))
 
 
-def extract_nsteps(result: Any) -> float | None:
+def state_infidelity_stats(states: Any, reference_states: Any) -> MetricStats:
+    """Return batch statistics of the mean trajectory infidelity."""
+    fidelities = jnp.clip(dq.fidelity(states, reference_states), 0.0, 1.0)
+    infidelities = 1.0 - fidelities
+    mean_time_infidelity = jnp.mean(infidelities, axis=-1)
+    return aggregate_stats(mean_time_infidelity)
+
+
+def extract_nsteps_stats(result: Any) -> MetricStats | None:
     infos = getattr(result, 'infos', None)
     if infos is None:
         return None
@@ -29,4 +56,10 @@ def extract_nsteps(result: Any) -> float | None:
     nsteps = jnp.asarray(nsteps)
     if nsteps.size == 0:
         return None
-    return float(jnp.mean(nsteps))
+    return aggregate_stats(nsteps)
+
+
+def extract_nsteps(result: Any) -> float | None:
+    """Return the mean step count, preserving the previous helper API."""
+    stats = extract_nsteps_stats(result)
+    return None if stats is None else stats.mean
--- a/tests/benchmark/metrics.py
+++ b/tests/benchmark/metrics.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from typing import Any
+
+import jax.numpy as jnp
+
+
+def relative_l2_error(value: Any, reference: Any) -> float:
+    value = jnp.asarray(value)
+    reference = jnp.asarray(reference)
+    numerator = jnp.linalg.norm(value - reference)
+    denominator = jnp.maximum(jnp.linalg.norm(reference), 1e-14)
+    return float(numerator / denominator)
+
+
+def max_abs_error(value: Any, reference: Any) -> float:
+    value = jnp.asarray(value)
+    reference = jnp.asarray(reference)
+    return float(jnp.max(jnp.abs(value - reference)))
+
+
+def extract_nsteps(result: Any) -> float | None:
+    infos = getattr(result, 'infos', None)
+    if infos is None:
+        return None
+    nsteps = getattr(infos, 'nsteps', None)
+    if nsteps is None:
+        return None
+    nsteps = jnp.asarray(nsteps)
+    if nsteps.size == 0:
+        return None
+    return float(jnp.mean(nsteps))
--- a/tests/benchmark/problems.py
+++ b/tests/benchmark/problems.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from collections.abc import Callable, Sequence
+from abc import ABC, abstractmethod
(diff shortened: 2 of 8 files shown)


tests/benchmark/metrics.py
```diff
@@ -0,0 +1,32 @@
...
+    reference = jnp.asarray(reference)
+    numerator = jnp.linalg.norm(value - reference)
+    denominator = jnp.maximum(jnp.linalg.norm(reference), 1e-14)
+    return float(numerator / denominator)
+
+
+def max_abs_error(value: Any, reference: Any) -> float:
+    value = jnp.asarray(value)
+    reference = jnp.asarray(reference)
+    return float(jnp.max(jnp.abs(value - reference)))
```
Nicolas-Lepage, line 38: both those metrics are good but i would like to add an error metric which is the fidelity of the state vs reference state or 1-fidelity.

tests/benchmark/metrics.py
```diff
@@ -0,0 +1,32 @@
...
+    infos = getattr(result, 'infos', None)
+    if infos is None:
+        return None
+    nsteps = getattr(infos, 'nsteps', None)
+    if nsteps is None:
+        return None
+    nsteps = jnp.asarray(nsteps)
+    if nsteps.size == 0:
+        return None
+    return float(jnp.mean(nsteps))
```
Nicolas-Lepage: This wasn't part of the initial issue, but it came to mind while I was reviewing it.

For batched problems, I think it would be useful for metrics such as the number of steps and the number of errors to return aggregated statistics across the batch. In particular, having the mean, minimum, and maximum values for each metric would provide a better overview of batch performance and variability.

tests/benchmark/problems.py
```diff
@@ -0,0 +1,396 @@
...
+    description: str
+    deterministic_methods: Sequence[str]
+    reference_name: str
+    run: Callable[[Method], Any]
+    reference: Callable[[], Any]
+    error: Callable[[Any, Any], float]
+
+
+def _solver_options() -> dict[str, Any]:
+    return {'save_states': False, 'progress_meter': NoProgressMeter()}
```
Nicolas-Lepage: Saving states is important for benchmarking purposes. Since the state is the direct output of the simulation, it provides the richest information for computing metrics.

Expectation values already involve a post-processing step, which may hide some errors or behaviors present in the raw output. Saving states would make metric computation both more flexible and more informative.

tests/benchmark/problems.py
```diff
@@ -0,0 +1,396 @@
...
+            tsave,
+            exp_ops=exp_ops,
+            method=method,
+            gradient=Direct(),
+            **_solver_options(),
+        )
+
+    def reference() -> jnp.ndarray:
+        result = run(
+            Dopri8(rtol=1e-9, atol=1e-9, safety_factor=0.75, max_steps=1_000_000)
```
Nicolas-Lepage: Particular care is needed for problems where the reference is computed with Dopri8. Since Dynamqis uses single precision by default, generating reference solutions with tolerances below `1e-7` can be problematic.

The standard way to enable double precision, `dq.set_precision("double")`, applies globally and must be called right after the imports. As a result, we should ensure that only the reference computation runs in double precision (optionally casting the result back to single precision afterward), while the rest of the benchmark remains unchanged.

Nicolas-Lepage (commented):
Thanks @JAMA4666 for the great PR ! For some reason it did not appear in the PRs linked to the issue so i took a bit of time to review. 

the PR is very good and does what we want but there are a few style / philosophy changes i would like you to do : 

- I would prefer the problem structure to stay closer to what we use in `tests/`: a base class defining the contract, and one child class per problem implementing it. you can take inspiration from `test/systems`

- For fixed-step methods, I do not think we should benchmark only one hardcoded `dt`. For `Euler` and `Rouchon1`, please sweep a few values, e.g. `1e-2`, `1e-3`, `1e-4`, so we can see the accuracy / cost tradeoff. Keeping adaptive methods on their defaults sounds good.

- For the error metric, I would prefer computing it on `states` rather than expectation values. The state is the direct output of the simulation and gives the most information. So I would save around 100 equally spaced states, compute the error at each time, and report the mean.

- For batched problems, such as Kerr and two-mode PWC, please do not collapse the batch to a single scalar. For both the step count and error metrics, report the mean, min, and max over the batch.

- I have left also a comment on reference vs the rest precision. on top of that comment it would be useful to have an option on running the problem with single or double precision. 

- For the method matrix, I would suggest:
`sesolve`: `Euler` with the 3 `dt` values, `Dopri5`, `Dopri8`, `Tsit5`, `Kvaerno3`, `Kvaerno5`
`mesolve`: same list, plus `Rouchon1` with the 3 `dt` values, `Rouchon2`, `Rouchon3`
One open question is `Expm`, which was one of the main methods in the issue. Since each problem now declares its own supported methods, I think constant-Hamiltonian problems, such as the Ising chain, can opt into `Expm`, while time-dependent problems leave it out.

Nicolas-Lepage:
Hi @JAMA4666 :) did you get a chance to finish the PR? I believe you have until one week after last Wednesday's deadline to wrap up the unitaryHACK PRs and still receive the bounty. Your implementation was the closest to what we're looking for, so we'd be glad to award you the bounty if you can address the last few issues this week.

JAMA4666:
Hi @Nicolas-Lepage, thank you so much for the detailed review and for the encouraging feedback on the PR!​I apologize for the delay in getting back to you; I've been quite busy with work lately and haven't had the chance to improve the solver as requested. I plan to address your comments and implement the suggested changes this week.​Honestly, more than the bounty, I'm really excited to know that the implementation is working as intended. I've applied some specific formulations that I designed myself, so receiving your technical feedback and seeing them validated through this process is truly the most rewarding part for me. Thanks for your time and guidance!Enviado desde mi Galaxy
-------- Mensaje original --------De: Nicolas-Lepage ***@***.***> Fecha: 22/6/2026  3:36 p.m.  (GMT+01:00) A: dynamiqs/dynamiqs ***@***.***> CC: JAMA4666 ***@***.***>, Mention ***@***.***> Asunto: Re: [dynamiqs/dynamiqs] Add opt-in solver benchmark suite (PR #1097) Nicolas-Lepage left a comment (dynamiqs/dynamiqs#1097)
Hi @JAMA4666 :) did you get a chance to finish the PR? I believe you have until one week after last Wednesday's deadline to wrap up the unitaryHACK PRs and still receive the bounty. Your implementation was the closest to what we're looking for, so we'd be glad to award you the bounty if you can address the last few issues this week.

JAMA4666:
Hi @Nicolas-Lepage! I've just pushed the updates addressing all your
feedback:

   -

   *Structure:* Reorganized the benchmarks to use the base class and child
   classes pattern (inspired by tests/systems).
   -

   *Metrics & Batches:* Added the state infidelity metric and updated the
   statistics (mean, min, max) so batch dimensions are no longer collapsed.
   -

   *Precision:* Configured the reference computation (Dopri8) to switch
   dynamically to double precision without affecting the rest of the benchmark.

The commits are up. Let me know if everything looks good on your end or if
any further adjustments are needed!


Regards.

gautierronan:
Thank you for your contribution. Unfortunately, we will be going for an in-house implementation (see https://github.com/dynamiqs/dynamiqs/pull/1146) as none of the open contributions were meeting our needs. Thanks again for taking the time to work on this.

PR #1098, opened 2026-06-04
Add Hessian-compatible gradient mode

### LINKED ISSUE #1082: unitaryHACK - Enable Hessian computation
### Context 

Linked to [issue](https://github.com/dynamiqs/dynamiqs/issues/1021) 

Dynamiqs currently does not reliably support Hessian computation through `dq.sesolve()` and `dq.mesolve()`.

This is useful because users may want to compute second-order derivatives of simulation outputs with respect to model parameters, for example for optimization, sensitivity analysis, curvature information, or uncertainty quantification.

At the moment, applying `jax.hessian(...)` to functions that call Dynamiqs solvers can fail because the current solver differentiation path is not compatible with higher-order autodiff in this setting.

[Diffrax Hessian documentation](https://docs.kidger.site/diffrax/examples/hessian/)

### What we want

Implement Hessian support for `dq.sesolve()` and, if possible, `dq.mesolve()`, following the approach used in Diffrax.

A successful contribution should:

- **Create a new higher-order gradient method in Dynamiqs**
  - add a new `gradient` mode specifically for higher-order autodiff / Hessian support;
  - keep this mode distinct from the existing first-order gradient options.

- **Use `DirectAdjoint` internally for this mode**
  - map the new higher-order gradient mode to a `DirectAdjoint`-style differentiation path through the Diffrax solve;
  - ensure the solve remains compatible with higher-order autodiff.

- **Construct Diffrax methods with `scan_kind="bounded"`**
  - when this new higher-order gradient mode is selected, instantiate the chosen Diffrax method with `scan_kind="bounded"` where required;
  - follow the Diffrax-recommended configuration for Hessian-compatible solves.

- **Handle saved states correctly**
  - ensure Hessian computation works with the Dynamiqs save path used by `sesolve()` and `mesolve()`;
  - support the usual saved outputs, rather than only working on a stripped-down internal solve.

- **Expose this through the existing solver API**
  - activate the behavior through the solver `gradient=` argument;
  - keep the user workflow JAX-native: users should be able to define a scalar-valued function that calls `dq.sesolve()` or `dq.mesolve()`, then apply `jax.hessian(...)`.

- **Validate the implementation with analytical tests**
  - add at least one `sesolve()` test where the second derivative is known analytically;
  - add a `mesolve()` test as well if `mesolve()` support is implemented;
  - verify that the computed Hessian matches the expected analytical result.

A good outcome for this issue would be:

- Hessians can be computed successfully for at least one standard `sesolve()` workflow;
- `mesolve()` is supported if feasible;
- the required solver configuration is handled cleanly inside Dynamiqs;
- tests are added to prevent regressions.

- A new Hessian-compatible gradient mode is added to Dynamiqs.
- The new mode is available through the existing solver `gradient=` argument.
- The implementation uses a Diffrax configuration compatible with higher-order autodiff.
- `dq.sesolve()` supports Hessian computation with this mode.
- `dq.mesolve()` supports Hessian computation with this mode if feasible.
- Saved outputs used by the public solver API are handled correctly.
- At least one analytical Hessian test is added for `sesolve()`.
- If `mesolve()` support is implemented, at least one analytical Hessian test is added for `mesolve()`.
- Tests verify the value returned by `jax.hessian(...)`, not only that the solve runs.
- Existing first-order gradient behavior is not broken.
- Existing tests continue to pass.

### Focus

Balanced, leaning engineering.

Focus on autodiff support, Diffrax solver configuration, public API integration, and validation against simple analytical quantum systems.

### Difficulty

Medium

---

### Proposed setup for Test 1

A simple `sesolve()` test can be built from a one-parameter qubit evolution where the final observable has a known second derivative.

Use a qubit Hamiltonian:

$$
H(\omega) = \frac{\omega}{2}\sigma_x
$$

with initial state:

$$
|\psi_0\rangle = |0\rangle
$$

and evolution time \(t\).

Under this evolution, the expectation value of \(\sigma_z\) is:

$$
\langle \sigma_z \rangle(\omega) = \cos(\omega t)
$$

so the first derivative is:

$$
\frac{d}{d\omega}\langle \sigma_z \rangle = -t\sin(\omega t)
$$

and the second derivative is:

$$
\frac{d^2}{d\omega^2}\langle \sigma_z  \rangle = -t^2\cos(\omega t)
$$

The test should:

- define a scalar function of \(\omega\) that calls `dq.sesolve()`;
- return the final expectation value of \(\sigma_z\);
- apply `jax.hessian(...)` to this function;
- compare the result against the analytical second derivative.

### Possible extension: two-parameter Hessian test

An optional stronger test is to use a two-parameter Hamiltonian, so that the Hessian is a full matrix rather than a scalar second derivative.

For example, use parameters controlling two Hamiltonian terms and compare the computed Hessian matrix against an analytical or high-precision reference result.

### Possible setup for `mesolve()`

If `mesolve()` support is included, add an analogous density-matrix test using a simple closed-system evolution or a dissipative model with a known analytical observable.

The test should again define a scalar-valued function of one or more parameters, call `dq.mesolve()`, apply `jax.hessian(...)`, and compare against the known second derivative.

jamilahmadzai:
## Summary
- Add `dq.gradient.HigherOrder()` for higher-order autodiff through Diffrax-based ODE solvers.
- Use `DirectAdjoint` with bounded solver scans for the new mode.
- Add analytical Hessian regression tests for `sesolve()` and `mesolve()`, plus docs API entries.

Closes #1082

## Validation
- `uv run --python 3.11 --with pytest --with-editable . pytest tests/sesolve/test_hessian.py tests/mesolve/test_hessian.py tests/sesolve/test_adaptive.py tests/mesolve/test_adaptive.py -q`
- `uv run --python 3.11 --with ruff ruff check dynamiqs/gradient.py dynamiqs/method.py dynamiqs/integrators/core/diffrax_integrator.py tests/sesolve/test_hessian.py tests/mesolve/test_hessian.py`
- `uv run --python 3.11 --with ruff ruff format --check dynamiqs/gradient.py dynamiqs/method.py dynamiqs/integrators/core/diffrax_integrator.py tests/sesolve/test_hessian.py tests/mesolve/test_hessian.py`
- `git diff --check`

### DIFF
--- a/docs/python_api/index.md
+++ b/docs/python_api/index.md
@@ -87,6 +87,7 @@ The **Dynamiqs** Python API features two main types of functions: solvers of dif
         - Direct
         - BackwardCheckpointed
         - Forward
+        - HigherOrder
 
 ### Progress meters
 
--- a/dynamiqs/gradient.py
+++ b/dynamiqs/gradient.py
@@ -4,7 +4,7 @@ import equinox as eqx
 
 from ._utils import tree_str_inline
 
-__all__ = ['Direct', 'BackwardCheckpointed', 'Forward', 'Gradient']
+__all__ = ['Direct', 'BackwardCheckpointed', 'Forward', 'HigherOrder', 'Gradient']
 
 
 class Gradient(eqx.Module):
@@ -103,6 +103,24 @@ class Forward(Gradient):
         For Diffrax-based methods, this falls back to the
         [`diffrax.ForwardMode`](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.ForwardMode)
         option.
+    """
+
+    # dummy init to have the signature in the documentation
+    def __init__(self):
+        pass
+
+
+class HigherOrder(Gradient):
+    """Higher-order automatic differentiation.
+
+    Enables support for higher-order automatic differentiation
+    (like [`jax.hessian`](https://docs.jax.dev/en/latest/_autosummary/jax.hessian.html))
+    through Diffrax-based solvers.
+
+    Note:
+        For Diffrax-based methods, this uses
+        [`diffrax.DirectAdjoint`](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.DirectAdjoint)
+        and configures compatible solver scans when required by Diffrax.
     """
 
     # dummy init to have the signature in the documentation
--- a/dynamiqs/integrators/core/diffrax_integrator.py
+++ b/dynamiqs/integrators/core/diffrax_integrator.py
@@ -6,12 +6,13 @@ from functools import partial
 
 import diffrax as dx
 import equinox as eqx
+from diffrax import AbstractRungeKutta
 from jax import Array
 from jaxtyping import PyTree, Scalar
 
 from ..._checks import check_hermitian
 from ..._utils import obj_type_str
-from ...gradient import BackwardCheckpointed, Direct, Forward, Gradient
+from ...gradient import BackwardCheckpointed, Direct, Forward, Gradient, HigherOrder
 from ...method import Dopri5, Dopri8, Euler, Kvaerno3, Kvaerno5, Method, Tsit5
 from ...options import Options
 from ...result import MESolveResult, Result, Saved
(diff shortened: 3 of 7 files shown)


Nicolas-Lepage:
Thanks for the PR! Unfortunately, we ended up merging another contribution that addressed the same issue, so I'll be closing this PR. Thanks again for taking the time to work on it.

PR #1100, opened 2026-06-05
Improve type check coverage

### LINKED ISSUE #1081: unitaryHACK - Expand type coverage across Dynamiqs
### Context

Dynamiqs already uses Python typing in parts of the codebase, and type checking is part of the development workflow through `task type`.

Contributors are also expected to type new functions, as described in the [contributing guide](https://github.com/dynamiqs/dynamiqs/blob/main/CONTRIBUTING.md).

However, type coverage is still incomplete. Several important modules are currently excluded from `ty` checking:

```
[tool.ty.src]
exclude = [
    "dynamiqs/integrators",
    "dynamiqs/plot",
    "dynamiqs/qarrays",
    "dynamiqs/random",
    "dynamiqs/utils",
    "dynamiqs/time_qarray.py"
]
```

The current CI does not use the latest version of `ty`, and updating `ty` is known to reveal new typing issues in files that are already type-checked today.

This issue therefore has two equally important parts:

1. make the currently type-checked parts of the library pass with the latest supported version of `ty`;
2. expand type coverage by typing currently excluded files and removing them from the exclusion list.

Improving type coverage will make the codebase easier to maintain, improve IDE/autocomplete support, and make future refactors safer.

### What we want

Update the type-checking setup so Dynamiqs works with the latest supported version of `ty`, then progressively expand type coverage across the library.

The first step is to run `task type` with the latest supported `ty` version and fix the issues reported in files that are already covered by type checking.

The second step is to type currently excluded files or modules. Any file or module that is correctly typed should be removed from the `ty` exclusion list.

The work should follow the existing development workflow:

- add or refine type annotations;
- run `task type`;
- fix reported typing issues;
- run `task all` locally before submitting.

### Important constraints

Typing issues should be fixed properly in the code or annotations.

Do not silence broad classes of errors using global ignores, blanket exclusions, or configuration changes that make `ty` stop checking the relevant code. Suppression comments may be used only when narrowly scoped and clearly justified.

Ignoring typing errors is not a fix. The goal is to increase real type coverage, not to make the checker pass by hiding errors.

- The project uses the latest supported version of `ty`.
- The parts of the library that are currently type-checked pass with that version.
- Newly typed files or modules are removed from the `ty` exclusion list.
- All files that are no longer excluded pass type checking.
- Type errors are fixed with real annotations or code changes, not broad ignores or exclusions.
- `task type` passes locally.
- `task all` passes locally.
- The full CI passes.
- Existing behavior is not changed except where needed to fix typing issues.

### Focus

Mostly engineering.

Focus on maintainability, API clarity, developer experience, and static correctness.

### Difficulty

Medium

Mayank447:
FIxes #1081 

Couple of notes to the reviewer:

1. I noticed `a = f(a)`, if the data type of `a` previously was different from when the function `f` was applied. The type checker is not able to infer that. Half of the changes are essentially converting `a` to `a_` on the LHS.

2. In time qarray, t is taken to be `ScalarLike` but this includes `complex` which is not feasible for Open Quantum System (unless we are doing Quantum Field Theory calculations or Wick's rotation). This should be replaced with `RealScalarLike` in the entire codebase I believe.

3. Default types like `QArray`, `QArrayLike`  lack proper documentation. Also overload should be used in dunder methods like `__mul__`, `__matmul__` etc. to handle interoperability between `QArray` and `Array`.

4. Not related to this PR, but a general note about ty. Unlike ruff, uv by astral, ty isn't yet ready and there is no stable beta yet (0.1 version). Also there are a lot of functionality missing in ty and its not really accurate in type checking as of now (check first comparison table at this [link](https://pyrefly.org/blog/typing-conformance-comparison/)). It might be better to switch to Pyrefly (its also written in rust and is stable) or pyright (low false positives and speed should not be an issue given that this codebase is not too big).

### DIFF
--- a/dynamiqs/integrators/_utils.py
+++ b/dynamiqs/integrators/_utils.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import math
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from functools import wraps
 from typing import Any
 
@@ -48,7 +48,7 @@ def ispwc(x: TimeQArray) -> bool:
         return False
 
 
-def catch_xla_runtime_error(func: callable) -> callable:
+def catch_xla_runtime_error(func: Callable) -> Callable:
     # Decorator to catch `XlaRuntimeError`` exceptions, and set a more friendly
     # exception message. Note that this will not work for jitted function, as the
     # exception code will be traced out.
@@ -91,8 +91,8 @@ def assert_method_supported(method: Method, supported_methods: Sequence[type[Met
 
 
 def multi_vmap(
-    f: callable, in_axes: int | None | Sequence[Any], out_axes: Any, nvmap: int
-) -> callable:
+    f: Callable, in_axes: int | None | Sequence[Any], out_axes: Any, nvmap: int
+) -> Callable:
     """Vectorize a function multiple time over multiple shared axes (similar to
     jnp.vectorize).
 
@@ -142,8 +142,8 @@ def multi_vmap(
 
 
 def cartesian_vmap(
-    f: callable, in_axes: int | None | Sequence[Any], out_axes: Any, nvmap: PyTree[int]
-) -> callable:
+    f: Callable, in_axes: int | None | Sequence[Any], out_axes: Any, nvmap: PyTree[int]
+) -> Callable:
     """Vectorize a function multiple time over distinct axes.
 
     The function `f` is mapped multiple time over on each input specified by `nvmap`
--- a/dynamiqs/integrators/_utils.py
+++ b/dynamiqs/integrators/_utils.py
@@ -81,7 +81,7 @@ def catch_xla_runtime_error(func: callable) -> callable:
     return wrapper
 
 
-def assert_method_supported(method: Method, supported_methods: Sequence[Method]):
+def assert_method_supported(method: Method, supported_methods: Sequence[type[Method]]):
     if not isinstance(method, tuple(supported_methods)):
         supported_str = ', '.join(f'`{x.__name__}`' for x in supported_methods)
         raise TypeError(
--- a/dynamiqs/integrators/_utils.py
+++ b/dynamiqs/integrators/_utils.py
@@ -27,7 +27,7 @@ class HasShape(Protocol):
     shape: tuple[int, ...]
 
 
-T = TypeVar('T', bound=HasShape)
+HasShapeType = TypeVar('HasShapeType', bound=HasShape)
 
 
 def astimeqarray(x: QArrayLike | TimeQArray) -> TimeQArray:
@@ -194,7 +194,9 @@ def cartesian_vmap(
     return f
 
 
-def attach_batch_indices(x: T, ndim_suffix: int = 2) -> tuple[T, Array]:
+def attach_batch_indices(
+    x: HasShapeType, ndim_suffix: int = 2
+) -> tuple[HasShapeType, Array]:
     """Bundle an array with an array of batch indices for key-folding through vmap.
 
     Returns ``(x, indices)`` where *indices* is an arange over the batch
--- a/dynamiqs/integrators/_utils.py
+++ b/dynamiqs/integrators/_utils.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import math
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from functools import wraps
-from typing import Any
+from typing import Any, Protocol, TypeVar
 
 import jax
 import jax.numpy as jnp
@@ -21,6 +21,13 @@ from ..time_qarray import (
     SummedTimeQArray,
     TimeQArray,
 )
+
+
+class HasShape(Protocol):
+    shape: tuple[int, ...]
+
+
+T = TypeVar('T', bound=HasShape)
 
 
 def astimeqarray(x: QArrayLike | TimeQArray) -> TimeQArray:
@@ -81,7 +88,7 @@ def catch_xla_runtime_error(func: Callable) -> Callable:
     return wrapper
 
 
-def assert_method_supported(method: Method, supported_methods: Sequence[type[Method]]):
+def assert_method_supported(method: Method, supported_methods: Iterable[type[Method]]):
     if not isinstance(method, tuple(supported_methods)):
         supported_str = ', '.join(f'`{x.__name__}`' for x in supported_methods)
         raise TypeError(
@@ -187,7 +194,7 @@ def cartesian_vmap(
     return f
 
 
-def attach_batch_indices(x: Array, ndim_suffix: int = 2) -> tuple[Array, Array]:
+def attach_batch_indices(x: T, ndim_suffix: int = 2) -> tuple[T, Array]:
     """Bundle an array with an array of batch indices for key-folding through vmap.
 
     Returns ``(x, indices)`` where *indices* is an arange over the batch
--- a/dynamiqs/integrators/_utils.py
+++ b/dynamiqs/integrators/_utils.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import functools
 import math
+import operator
 from collections.abc import Callable, Iterable, Sequence
 from functools import wraps
 from typing import Any, Protocol, TypeVar
@@ -13,7 +15,7 @@ from jaxtyping import PRNGKeyArray, PyTree
 
 from .._utils import obj_type_str
 from ..method import Method, _DEAdaptiveStep
-from ..qarrays.qarray import QArrayLike
+from ..qarrays.qarray import QArray, QArrayLike
 from ..qarrays.utils import asqarray
 from ..time_qarray import (
     ConstantTimeQArray,
@@ -27,7 +29,11 @@ class HasShape(Protocol):
     shape: tuple[int, ...]
 
 
-HasShapeType = TypeVar('HasShapeType', bound=HasShape)
+HasShapeT = TypeVar('HasShapeT', bound=HasShape)
+
+
+def sum_qarrays(qarrays: list[QArray]) -> QArray:
+    return functools.reduce(operator.add, qarrays)
 
 
 def astimeqarray(x: QArrayLike | TimeQArray) -> TimeQArray:
@@ -194,9 +200,7 @@ def cartesian_vmap(
     return f
 
 
-def attach_batch_indices(
-    x: HasShapeType, ndim_suffix: int = 2
-) -> tuple[HasShapeType, Array]:
+def attach_batch_indices(x: HasShapeT, ndim_suffix: int = 2) -> tuple[HasShapeT, Array]:
     """Bundle an array with an array of batch indices for key-folding through vmap.
 
     Returns ``(x, indices)`` where *indices* is an arange over the batch
--- a/dynamiqs/integrators/apis/dsmesolve.py
+++ b/dynamiqs/integrators/apis/dsmesolve.py
@@ -44,7 +44,7 @@ def dsmesolve(
     gradient: Gradient | None = None,
     save_states: bool = True,
     cartesian_batching: bool = True,
-    save_extra: Callable[[Array], PyTree] | None = None,
+    save_extra: Callable[[QArray], PyTree] | None = None,
 ) -> DSMESolveResult:
     r"""Solve the diffusive stochastic master equation (SME).
 
@@ -398,7 +398,7 @@ def _dsmesolve_many_trajectories(
 
     # vectorize input over keys
     in_axes = (None, None, None, None, None, None, 0, None, None, None, None)
-    out_axes = DSMESolveResult(None, None, None, None, 0, 0, 0)
+    out_axes = DSMESolveResult(None, None, None, None, 0, 0, 0)  # ty: ignore
     f = jax.vmap(_dsmesolve_single_trajectory, in_axes, out_axes)
     return f(H, Lcs, Lms, etas, rho0, tsave, keys, exp_ops, method, gradient, options)
 
--- a/dynamiqs/integrators/apis/dsmesolve.py
+++ b/dynamiqs/integrators/apis/dsmesolve.py
@@ -277,9 +277,9 @@ def dsmesolve(
     rho0 = asqarray(rho0)
     keys = jnp.asarray(keys)
 
-    exp_ops_ = None
+    _exp_ops = None
     if exp_ops is not None:
-        exp_ops_ = [asqarray(E) for E in exp_ops] if len(exp_ops) > 0 else None
+        _exp_ops = [asqarray(E) for E in exp_ops] if len(exp_ops) > 0 else None
 
     # === build options
     options = Options(
@@ -289,18 +289,18 @@ def dsmesolve(
     )
 
     # === check arguments
-    _check_dsmesolve_args(H, Ls, etas, rho0, exp_ops_)
+    _check_dsmesolve_args(H, Ls, etas, rho0, _exp_ops)
     check_options(options, 'dsmesolve')
     options = options.initialise()
 
     # todo: fix static tsave
     # this condition allows the user to pass a tuple for tsave to bypass this bit of
     # code (e.g., to JIT-compile this function)
-    tsave_ = tsave
+    _tsave = tsave
     if not isinstance(tsave, tuple):
-        tsave_ = jnp.asarray(tsave)
-        tsave_ = check_times(tsave_, 'tsave')
-        tsave_ = tuple(tsave_.tolist())
+        _tsave = jnp.asarray(tsave)
+        _tsave = check_times(_tsave, 'tsave')
+        _tsave = tuple(_tsave.tolist())
 
     if method is None:
         raise ValueError('Argument `method` must be specified.')
@@ -318,7 +318,7 @@ def dsmesolve(
     # we implement the jitted vectorization in another function to pre-convert QuTiP
     # objects (which are not JIT-compatible) to JAX arrays
     return _vectorized_dsmesolve(
-        H, Lcs, Lms, etas, rho0, tsave_, keys, exp_ops_, method, gradient, options
+        H, Lcs, Lms, etas, rho0, _tsave, keys, _exp_ops, method, gradient, options
     )
 
 
--- a/dynamiqs/integrators/apis/dsmesolve.py
+++ b/dynamiqs/integrators/apis/dsmesolve.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from functools import partial
+from typing import cast
 
 import jax
 import jax.numpy as jnp
@@ -275,8 +276,10 @@ def dsmesolve(
     etas = jnp.asarray(etas)
     rho0 = asqarray(rho0)
     keys = jnp.asarray(keys)
+
+    exp_ops_ = None
     if exp_ops is not None:
-        exp_ops = [asqarray(E) for E in exp_ops] if len(exp_ops) > 0 else None
+        exp_ops_ = [asqarray(E) for E in exp_ops] if len(exp_ops) > 0 else None
 
     # === build options
     options = Options(
@@ -286,18 +289,18 @@ def dsmesolve(
     )
 
     # === check arguments
-    _check_dsmesolve_args(H, Ls, etas, rho0, exp_ops)
-    tsave = check_times(tsave, 'tsave')
+    _check_dsmesolve_args(H, Ls, etas, rho0, exp_ops_)
     check_options(options, 'dsmesolve')
     options = options.initialise()
 
     # todo: fix static tsave
     # this condition allows the user to pass a tuple for tsave to bypass this bit of
     # code (e.g., to JIT-compile this function)
+    tsave_ = tsave
     if not isinstance(tsave, tuple):
-        tsave = jnp.asarray(tsave)
-        tsave = check_times(tsave, 'tsave')
-        tsave = tuple(tsave.tolist())
+        tsave_ = jnp.asarray(tsave)
+        tsave_ = check_times(tsave_, 'tsave')
+        tsave_ = tuple(tsave_.tolist())
 
     if method is None:
         raise ValueError('Argument `method` must be specified.')
@@ -315,7 +318,7 @@ def dsmesolve(
     # we implement the jitted vectorization in another function to pre-convert QuTiP
     # objects (which are not JIT-compatible) to JAX arrays
     return _vectorized_dsmesolve(
-        H, Lcs, Lms, etas, rho0, tsave, keys, exp_ops, method, gradient, options
+        H, Lcs, Lms, etas, rho0, tsave_, keys, exp_ops_, method, gradient, options
     )
 
 
@@ -422,7 +425,7 @@ def _dsmesolve_single_trajectory(
         Rouchon1: dsmesolve_rouchon1_integrator_constructor,
     }
     assert_method_supported(method, integrator_constructors.keys())
-    integrator_constructor = integrator_constructors[type(method)]
+    integrator_constructor = integrator_constructors[type(method)]  # ty: ignore
 
     # === check gradient is supported
     method.assert_supports_gradient(gradient)
@@ -444,10 +447,7 @@ def _dsmesolve_single_trajectory(
     )
 
     # === run solver
-    result = integrator.run()
-
-    # === return result
-    return result  # noqa: RET504
+    return cast(DSMESolveResult, integrator.run())
 
 
 def _check_dsmesolve_args(
--- a/dynamiqs/integrators/core/abstract_integrator.py
+++ b/dynamiqs/integrators/core/abstract_integrator.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Generic, TypeVar
+from typing import Generic
 
 import equinox as eqx
 import jax.numpy as jnp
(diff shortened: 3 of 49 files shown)


Nicolas-Lepage:
Hi @Mayank447, we really like your draft PR and are keen to review it for the Hackathon. 
Are you still actively working on it, and do you have a timeline for submitting?

One small thing to clear when you get a chance: CI is failing on two small ruff lint errors (not typing)

Mayank447:
> Hi @Mayank447, we really like your draft PR and are keen to review it for the Hackathon. Are you still actively working on it, and do you have a timeline for submitting?

Hi @Nicolas-Lepage, I am still working on the PR. I was travelling past few days so didn't get time to finish the typehinting the remaining `integrator` module. I just came back today, so will complete the typehinting in 1-2 days.

Mayank447:
@Nicolas-Lepage This PR is ready for review

dynamiqs/integrators/apis/dsmesolve.py
```diff
@@ -275,8 +276,10 @@ def dsmesolve(
     etas = jnp.asarray(etas)
     rho0 = asqarray(rho0)
     keys = jnp.asarray(keys)
+
+    exp_ops_ = None
```
gautierronan: Instead of trailing underscores, could you use leading underscores? i.e. `_exp_ops` here, and similarly throughout the PR?

dynamiqs/integrators/apis/dsmesolve.py
```diff
@@ -286,18 +289,18 @@ def dsmesolve(
     )
 
     # === check arguments
-    _check_dsmesolve_args(H, Ls, etas, rho0, exp_ops)
-    tsave = check_times(tsave, 'tsave')
```
gautierronan, line 290: Why do you remove this check?

dynamiqs/integrators/apis/dsmesolve.py
```diff
@@ -398,7 +401,7 @@ def _dsmesolve_many_trajectories(
 
     # vectorize input over keys
     in_axes = (None, None, None, None, None, None, 0, None, None, None, None)
-    out_axes = DSMESolveResult(None, None, None, None, 0, 0, 0)
+    out_axes = DSMESolveResult(None, None, None, None, 0, 0, 0)  # ty: ignore
```
gautierronan, line 404: Why do we need a new `ty: ignore` statement here? Can we fix it instead?

dynamiqs/integrators/core/abstract_integrator.py
```diff
@@ -9,9 +10,11 @@
 
 from ...gradient import Gradient
 from ...method import Method
-from ...result import Result, Saved
+from ...result import Result, Saved, SolveSaved, StochasticSolveResult
 from .interfaces import OptionsInterface
 
+_SavedT = TypeVar('_SavedT', bound=Saved)
```
gautierronan: This looks like dark magic to me. Could you add a small comment stating why it's useful and why does `BaseIntegrator` needs to inherit from `Generic[_SavedT]` ?

dynamiqs/integrators/core/diffrax_integrator.py
```diff
@@ -60,24 +73,30 @@ def stepsize_controller(self) -> dx.AbstractStepSizeController:
...
         else:
-            jump_ts = None if len(self.discontinuity_ts) == 0 else self.discontinuity_ts
+            method = cast(_DEAdaptiveStep, self.method)
             return dx.PIDController(
-                rtol=self.method.rtol,
-                atol=self.method.atol,
-                safety=self.method.safety_factor,
-                factormin=self.method.min_factor,
-                factormax=self.method.max_factor,
-                jump_ts=jump_ts,
```
gautierronan, line 70: We should probably not remove the argument `jump_ts` here

dynamiqs/integrators/core/diffrax_integrator.py
```diff
@@ -60,24 +73,30 @@ def stepsize_controller(self) -> dx.AbstractStepSizeController:
...
 
     @property
     def max_steps(self) -> int:
-        # TODO: fix hard-coded max_steps for fixed methods
-        return 100_000 if self.fixed_step else self.method.max_steps
+        if self.fixed_step:
+            # TODO: fix hard-coded max_steps for fixed methods
+            return 100_000
+        else:
+            assert isinstance(self.method, _DEAdaptiveStep)
```
gautierronan: Why use an `assert` here whereas you used a `cast` previously? Can we unify those in one way or another? If possible, I prefer `assert` statements since they would return a more explicit error in case of failure.

Same comment throughout the rest of the PR

dynamiqs/integrators/core/event_integrator.py
```diff
@@ -60,10 +63,14 @@ class JSSESolveEventIntegrator(
...
     """Integrator computing the time evolution of the Jump SSE using Diffrax events."""
 
+    method: Event
+
     def run(self) -> Result:
         if self.method.smart_sampling:
             # sample a no-click trajectory and compute its probability
             solution = self._solve_noclick(self.ts, self.y0)
+
+            assert solution.ys is not None  # Only for the typechecker
```
gautierronan:
```suggestion
            assert solution.ys is not None
```

dynamiqs/integrators/core/event_integrator.py
```diff
@@ -74,17 +81,17 @@ def run(self) -> Result:
             infos = None
 
         # vectorize over keys
-        out_axes = JumpSolveSaved(0, 0, 0, 0)
+        out_axes = JumpSolveSaved(0, 0, 0, 0)  # ty: ignore[invalid-argument-type]
```
gautierronan, line 84: Can we remove this ignore statement?

dynamiqs/integrators/core/event_integrator.py
```diff
@@ -106,23 +113,25 @@ def _solve_noclick(
...
+    def _solve_until_click(self, y: JumpState, rand: Array) -> tuple[JumpState, bool]:
         # === solve until the next click event
         cond_fn = lambda t, y, *a, **kw: y.norm() ** 2 - rand  # noqa: ARG005
         event = dx.Event(cond_fn, self.method.root_finder)
         ts = jnp.where(self.ts >= y.t, self.ts, y.t)  # clip times to start at y.t
         solution = self._solve_noclick(ts, y.psi, event=event, save=self.save)
 
         # === collect solve result
+        assert solution.ys is not None  # Only for the typechecker
+        assert solution.ts is not None  # Only for the typechecker
```
gautierronan: Same comment here and later on

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
```diff
@@ -90,19 +96,20 @@ def __str__(self) -> str:
...
     def dt(self) -> float:
-        return self.method.dt
+        method = cast(_DEFixedStep, self.method)
+        return method.dt
 
     @property
     def total_nsteps(self) -> int:
         # total number of steps of length dt
-        return round((self.t1 - self.t0) / self.dt)
+        return round((self.t1 - self.t0) / self.dt)  # ty: ignore
```
gautierronan, line 107: Same comment here about the `ignore` statement

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
```diff
@@ -188,10 +195,9 @@ def outer_step(carry, key):  # noqa: ANN001, ANN202
         # === collect and format results
         # insert the initial saved result
         saved = jax.tree.map(lambda x, y: jnp.insert(x, 0, y, axis=0), saved, saved0)
-        # postprocess the saved results
-        saved = self.postprocess_saved(saved, ylast)
```
gautierronan, line 192: This seems surprising that we remove this call to `postprocess_saved` here

dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -214,46 +226,40 @@ def vector_field(t, m, _):  # noqa: ANN001, ANN202
...
 
-        saved = self.postprocess_saved(*solution.ys)
+        ys = cast(tuple, solution.ys)
+        saved = self.postprocess_saved(*ys)
         return self.result(saved, infos=self.infos(solution.stats))
 
-    def save(self, m: PyTree) -> SolveSaved:
-        m = normalize_m(m)
+    def save(self, y: PyTree) -> SolveSaved:
+        y = normalize_m(y)
```
gautierronan: Why do you rename the variable here? In this case, `m` is a better variable name

dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -214,46 +226,40 @@ def vector_field(t, m, _):  # noqa: ANN001, ANN202
...
         if self.Es_jax is not None:
-            Esave = jnp.stack([expval_from_m(m, E) for E in self.Es_jax])
+            Esave = jnp.stack([expval_from_m(y, E) for E in self.Es_jax])
         else:
             Esave = None
 
         return SolveSaved(ysave=msave, extra=extra, Esave=Esave)
 
-    def postprocess_saved(self, saved: Saved, mlast: PyTree) -> Saved:
+    def postprocess_saved(self, saved: SolveSaved, ylast: PyTree) -> SolveSaved:
```
gautierronan, line 254: Same comment for `mlast`

dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -214,46 +226,40 @@ def vector_field(t, m, _):  # noqa: ANN001, ANN202
...
     def infos(self, stats: dict[str, Array]) -> PyTree:
-        fixed_step = {
-            Euler: True,
-            Dopri5: False,
-            Dopri8: False,
-            Tsit5: False,
-            Kvaerno3: False,
-            Kvaerno5: False,
-        }[type(self.method.ode_method)]
+        fixed_step = isinstance(self.method.ode_method, Euler)
```
gautierronan: Can you revert this change and fix the typing in another way? We intentionally keep the dictionary like so in case we later want to introduce other fixed step methods

dynamiqs/integrators/core/rouchon_integrator.py
```diff
@@ -60,11 +63,11 @@ def step(
...
         y0: Y,
         args: Args,
-        solver_state: None,
-        made_jump: bool,
+        solver_state: PyTree | None,
+        made_jump: BoolScalarLike,
     ) -> tuple:
         del solver_state, made_jump, args
-        y1, error = terms.term.rouchon_step(t0, t1, y0)
+        y1, error = terms.term.rouchon_step(t0, t1, y0)  # ty: ignore
```
gautierronan, line 71: Can you remove the ignore statement here?

dynamiqs/integrators/core/rouchon_integrator.py
```diff
@@ -155,7 +158,10 @@ def propagator_ratio(self, ci: float, cj: float) -> QArray:
 
     def dissipator(self, c: float, rho: QArray) -> QArray:
         r"""Jump map $D_c(\rho) = \sum_k L_k \rho L_k^\dagger$ at $t + c\Delta t$."""
-        return sum(M_rho_Mdag(L, rho) for L in self.L(self.t + c * self.dt))
+        return sum(
+            (M_rho_Mdag(L, rho) for L in self.L(self.t + c * self.dt)),
+            start=zeros_like(rho),
+        )
```
gautierronan: I don't like this solution of using `sum(..., start=...)` because we are going to instantiate a bunch of zero-valued matrices, which is going to hinder performance.

Can you find an alternative solution to the typing that doesn't require creating a new array for nothing?

dynamiqs/integrators/_utils.py
```diff
@@ -23,6 +23,13 @@
 )
 
 
+class HasShape(Protocol):
+    shape: tuple[int, ...]
+
+
+T = TypeVar('T', bound=HasShape)
```
gautierronan: I don't really understand this either. Can you add a comment to explain what it's doing?

dynamiqs/method.py
```diff
@@ -289,6 +289,7 @@ def __init__(
         normalize: bool = True,
     ):
         _DEFixedStep.__init__(self, dt)  # ty: ignore[invalid-argument-type]
```
gautierronan, line 289: Is it possible to fix these ignore statements as well?

dynamiqs/time_qarray.py
```diff
@@ -477,7 +480,10 @@ def __radd__(self, y: QArrayLike | TimeQArray) -> TimeQArray:
         return self + y
 
     def __sub__(self, y: QArrayLike | TimeQArray) -> TimeQArray:
-        return self + (-y)
+        if isinstance(y, TimeQArray):
+            return self + (-y)
+        else:
+            return NotImplemented
```
gautierronan, line 488: I don't understand this change. Isn't the subtraction between a TimeQArray and a QArrayLike implemented?

gautierronan (changes requested):
Thank you for your contribution @Mayank447. It's very nice to improve our typing coverage, which was quite lacking for now.

It's a fairly long PR and tedious to review, but I tried to be thorough. I made quite a few comments below, some of which apply throughout the PR. Thanks again for taking the time to work on this!

dynamiqs/integrators/apis/dsmesolve.py
Mayank447: This check was already being done on line 302. I only found dsmesolve to a check both inside and outside `if not isinstance(tsave, tuple)`, every other integrator was doing this check inside the if statement.

dynamiqs/integrators/apis/dsmesolve.py
Mayank447: Essentially it will throw an error cause None is not of Array type and similarly 0 is a literal which is again not Array type. I couldn't find a better way to solve it other than casting 5 arguments, so that's why left a `# ty: ignore`

dynamiqs/integrators/core/low_rank_integrator.py
Mayank447: It was earlier throwing an error due to mismatch of argument name for `save` compared to its base classes. I can rename it to m within the function implementation

dynamiqs/integrators/core/low_rank_integrator.py
Mayank447: Same reason as above

dynamiqs/integrators/_utils.py
Mayank447: So this essentially declares `T` represents any type which has a shape attribute

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
Mayank447: Thanks, for the catch. Restored it, I don't know how but I for reason somehow mistakenly deleted it

dynamiqs/integrators/core/event_integrator.py
Mayank447: So plain 0 is a literal in python, not of type Array.

dynamiqs/integrators/core/diffrax_integrator.py
Mayank447: I had be happy to make the change. I had just one point, `assert` would have some overhead (though not a large one), `cast` is without a run time overhead and in future as `ty` improves some of those redundant cast statements won't be required.

dynamiqs/integrators/core/diffrax_integrator.py
Mayank447: diffrax `PIDController` doesn't have a `jump_ts` argument in its constructor. https://docs.kidger.site/diffrax/api/stepsize_controller/#diffrax.PIDController

dynamiqs/integrators/apis/dsmesolve.py
gautierronan: Hmm, ok. It's true that this jax syntax is a bit strange with respect to typing. Let's keep the `# ty: ignore` then!

dynamiqs/integrators/core/abstract_integrator.py
Mayank447: Will do

dynamiqs/integrators/apis/dsmesolve.py
Mayank447: Will do

dynamiqs/integrators/core/diffrax_integrator.py
gautierronan: Ok, good point. Let's keep `cast` then instead of `assert`!

dynamiqs/integrators/core/event_integrator.py
gautierronan: Yes I guess it's the same as my previous comment with `out_axes`. Let's keep the ignore statement

dynamiqs/integrators/core/low_rank_integrator.py
gautierronan: Ah, I didn't know typing would raise such errors. Yes, if possible, let's rename it `m = y` in the first line of the function then.

dynamiqs/integrators/core/rouchon_integrator.py
Mayank447: So essentially `AbstractRouchonTerm` does not define `term`, I was not sure what type `term` should be so left a ignore instead.

dynamiqs/integrators/_utils.py
gautierronan: Ok, it's again a bit of dark magic to me haha. Could you rename `T` to `HasShapeType` or something along those lines ?

dynamiqs/method.py
Mayank447: I don't think we can.

Fixed step integrator must have a time step, but `dt` can be `None` as per the function signature. Also `Rouchon2` test does call it with default `None`.

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
Mayank447: For some reason ty was not able to infer `round()` returns an `int`. But if I use `int(round(...))`, ruff throws an error

dynamiqs/integrators/core/low_rank_integrator.py
Mayank447: I have replaced `Euler` with `_DEFixedStep`, this should allow to later introduce more fixed step methods

dynamiqs/time_qarray.py
Mayank447: Good catch, I have made the change

Mayank447:
@gautierronan I have made the changes as per the review.

dynamiqs/time_qarray.py
```diff
@@ -477,7 +480,12 @@ def __radd__(self, y: QArrayLike | TimeQArray) -> TimeQArray:
         return self + y
 
     def __sub__(self, y: QArrayLike | TimeQArray) -> TimeQArray:
-        return self + (-y)
+        if isqarraylike(y):
+            y = ConstantTimeQArray(asqarray(y))
+        if isinstance(y, TimeQArray):
+            return self + (-y)
```
gautierronan, line 486: Why doesn't this work? This would avoid instanciating a new timeqarray
```suggestion
        if isqarraylike(y) or isinstance(y, TimeQArray):
            return self + (-y)
```

dynamiqs/integrators/core/rouchon_integrator.py
```diff
@@ -155,7 +160,9 @@ def propagator_ratio(self, ci: float, cj: float) -> QArray:
 
     def dissipator(self, c: float, rho: QArray) -> QArray:
         r"""Jump map $D_c(\rho) = \sum_k L_k \rho L_k^\dagger$ at $t + c\Delta t$."""
-        return sum(M_rho_Mdag(L, rho) for L in self.L(self.t + c * self.dt))
+        return functools.reduce(
+            operator.add, (M_rho_Mdag(L, rho) for L in self.L(self.t + c * self.dt))
+        )
```
gautierronan: For readability/maintainability, could you introduce (at the top of the file, or inside some utils file):
```python
def _sum_qarrays(qarrays: list[QArray]) -> QArray:
    return functools.reduce(operator.add, qarrays)
```

dynamiqs/integrators/core/abstract_integrator.py
gautierronan: Could you move this inside `save_mixing.py` and make it a not private type, i.e. `SavedT` instead of `_SavedT` ? In principle, it could be used by other modules

Actually, I realized it's duplicated already inside `save_mixing.py`.

dynamiqs/integrators/_utils.py
```diff
@@ -23,6 +23,13 @@
 )
 
 
+class HasShape(Protocol):
+    shape: tuple[int, ...]
+
+
+HasShapeType = TypeVar('HasShapeType', bound=HasShape)
```
gautierronan: I actually read a bit more about type variables, and PEP 8 recommands the T suffix. So, if you don't mind, could you revert back to the previous name `HasShapeT`? Sorry about this, I should have informed myself sooner

dynamiqs/integrators/core/diffrax_integrator.py
gautierronan: Oh, then we have a typing issue. Could it be a `diffrax.ClipStepSizeController` that we're creating here? It's quite surprising that it doesn't raise an error if we try to pass a kwarg that doesn't exist.

dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
gautierronan: Ok, I guess that's a `ty` mistake then. Can you add an inline comment above saying why you require this `ty: ignore` here? So we can fix it once `ty` has fixed it.

gautierronan (changes requested):
Thanks for the changes @Mayank447. I have a few more comments following your answers and your fixes. Could you take these into account as well? Then, I think we're good to merge!

dynamiqs/time_qarray.py
Mayank447: Essentially negation if not supported for all `QArrayLike` datatypes. This is the error

Unary operator `-` is not supported for object of type `Array | ndarray[tuple[Any, ...], dtype[Any]] | numpy.bool[builtins.bool] | ... omitted 8 union elements`
   --> dynamiqs/time_qarray.py:484:28
    |
484 |             return self + (-y)
    |                            ^^
    |
info: `numpy.bool[builtins.bool]` does not implement `__neg__`
info: `Sequence[Divergent]` does not implement `__neg__`

gautierronan: Ok, but the problem right now is that we create a `ConstantTimeQArray` just to fix typing. It might be fast, but I would prefer it if we could find a more direct solution. Would you have an idea?

Mayank447:
@gautierronan Hi, could you review the latest commit, I merged the main branch and there were some typehinting errors for `MaterializedQArray`, I have fixed those. 

Regarding `jump_ts` I will open a separate issue cause that's not actually related to type hinting. Even I was surprised it earlier didn't raise an error even though `diffrax. PIDController` doesn't have that argument nor `**kwargs`.

dynamiqs/time_qarray.py
Mayank447: Instead of `ConstantTimeQArray ` we can use `asqarray`?

dynamiqs/qarrays/utils.py
```diff
@@ -188,18 +189,25 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
-        data = jnp.stack(
-            [cast(DenseDataArray, q.data).data for q in qarrays], axis=axis
-        )
-        return QArray(dims, False, DenseDataArray(data))
-    elif all(isinstance(q.data, SparseDIADataArray) for q in qarrays):
+    if all(
+        isinstance(q, MaterializedQArray) and isinstance(q.data, DenseDataArray)
+        for q in qarrays
+    ):
+        mats = cast(list[MaterializedQArray], qarrays)
```
gautierronan:
```suggestion
        _qarrays = cast(list[MaterializedQArray], qarrays)
```

dynamiqs/time_qarray.py
gautierronan: Yes using `asqarray` is better if it works without typing issues!

gautierronan (changes requested):
Ok, we're getting there! I have two small comments left (about the timeqarray `__neg__` and about a variable name, and we're good I think! The CI also fails for a small typing issue, but it looks easy to fix

gautierronan (dismissed):
Everything looks good to me. That's a great PR and a great first contribution @Mayank447. Thank you for your quick iterations.

@Nicolas-Lepage Can you take a quick second look before merging?

dynamiqs/plot/utils.py
```diff
@@ -378,6 +385,7 @@ def wrapper(
             plt.close()
             plot_function(items[idx], *args, **kwargs)  # plot frame
             canvas = plt.gcf().canvas
```
Nicolas-Lepage: With the cast just after isn't this a duplicate ?

Nicolas-Lepage (dismissed):
Hi @Mayank447, this is a substantial PR, really impressive work!

I've taken a quick look, as I trust @gautierronan's review.

I only left one very small comment: I think there may be a duplicate line in the plotting code (`plot/utils.py`)

Otherwise, we're ready to merge. Thanks again for tackling this tedious piece of work!

dynamiqs/plot/utils.py
Mayank447: Nice catch, made the change

Nicolas-Lepage (approved):

PR #1101, opened 2026-06-05
split QArray into an abstract base + MaterializedQArray [CompositeQArray prerequisite]

Nicolas-Lepage:
**Prerequisite for the composite QArray feature. This PR does *not* implement `CompositeQArray`.**

linked to https://github.com/dynamiqs/dynamiqs/pull/1091 and https://github.com/dynamiqs/dynamiqs/pull/1091#pullrequestreview-4407719420

It turns the old concrete `QArray` into an abstract base class and moves the existing
implementation, unchanged in behavior, into a new concrete `MaterializedQArray`. The
target structure (PR implemnets only the first branch) :

```
QArray (abstract)
│
├── MaterializedQArray (concrete)
│   └── data : DataArray (abstract)
│       ├── DenseDataArray (concrete)
│       └── SparseDIADataArray (concrete)
│
└── CompositeQArray (concrete)
    └── terms : list[CompositeTerm]
        └── operators : list[MaterializedQArray]
            └── data : DataArray (abstract)
                ├── DenseDataArray (concrete)
                └── SparseDIADataArray (concrete)
```

The point of review is the placement decision made for every member of the old `QArray`.
Each was classified as **materialized-specific** (stays out of the base),
**abstract** (shared signature, per-subclass impl), or **concrete** (defined once in the
base, inherited). Lists below.

## Attributes (fields / class attrs)

- `dims` (static `eqx.field`) — **concrete in base**
- `__array_priority__ = 200` — **concrete in base**
- `data: DataArray` — **materialized-specific**
- `vectorized: bool` — **materialized-specific**

## Properties

- `dtype`, `layout`, `shape`, `mT`, `ndim` — **abstract**
- `ndiags` — **materialized-specific** (sparse-dia only)

## Methods

Abstract (signature in base, implemented per subclass):

- structure/shape: `reshape`, `_reshape_unchecked`, `broadcast_to`, `squeeze`, `sum`,
  `__getitem__`
- linalg: `conj`, `powm`, `expm`, `norm`, `trace`, `_eig`, `_eigh`, `_eigvals`,
  `_eigvalsh`, `isherm`
- element-wise: `addscalar`, `elmul`, `elpow`
- arithmetic primitives: `__mul__`, `__add__`, `__matmul__`, `__rmatmul__`, `__and__`
- quantum: `ptrace`
- conversion: `to_qutip`, `to_jax`, `to_numpy`, `__array__`, `asdense`, `assparsedia`
- infra: `__check_init__`, `__repr__`, `devices`, `block_until_ready`

Concrete in base (written purely on top of the abstract primitives — subclasses get them
for free):

- `dag` (← `mT` + `conj`)
- `mean` (← `sum`, `shape`, `ndim`) — promoted from the concrete override this PR
- `unit` (← `norm`)
- derived arithmetic: `__neg__`, `__rmul__`, `__truediv__`, `__rtruediv__`, `__radd__`,
  `__sub__`, `__rsub__`, `__pow__` (handles the diffrax `_Metaω` case)
- iteration: `__len__`, `__iter__`
- utils shortcuts: `cosm`, `sinm`, `signm`, `isket`, `isbra`, `isdm`, `isop`, `toket`,
  `tobra`, `todm`, `proj`

## Notes for reviewers

- Choice of most uncertainty is the utils shortcut just above i am not sure if we can keep them concrete.

## Other changed files

`mepropagator.py`, `qarrays/__init__.py`, `qarrays/utils.py`, `utils/operators.py`,
`test_sparse_dia_qarray.py` are stale `QArray` constructions/imports updated to
`MaterializedQArray`.

### DIFF
--- a/dynamiqs/qarrays/materialized_qarray.py
+++ b/dynamiqs/qarrays/materialized_qarray.py
@@ -23,6 +23,9 @@ class MaterializedQArray(QArray):
     data: DataArray
 
     def __check_init__(self):
+
+        super().__check_init__()
+
         # === ensure dims is compatible with the shape
         # for vectorized superoperators, we allow that the shape is the square
         # of the product of all dims
--- a/dynamiqs/qarrays/materialized_qarray.py
+++ b/dynamiqs/qarrays/materialized_qarray.py
@@ -23,14 +23,6 @@ class MaterializedQArray(QArray):
     data: DataArray
 
     def __check_init__(self):
-        # === ensure dims is a tuple of ints
-        if not isinstance(self.dims, tuple) or not all(
-            isinstance(d, int) for d in self.dims
-        ):
-            raise TypeError(
-                f'Argument `dims` must be a tuple of ints, but is {self.dims}.'
-            )
-
         # === ensure dims is compatible with the shape
         # for vectorized superoperators, we allow that the shape is the square
         # of the product of all dims
--- a/dynamiqs/qarrays/materialized_qarray.py
+++ b/dynamiqs/qarrays/materialized_qarray.py
@@ -1,0 +1,366 @@
+from __future__ import annotations
+
+from dataclasses import replace
+from math import prod
+
+import equinox as eqx
+import jax.numpy as jnp
+import numpy as np
+from jax import Array, Device
+from jaxtyping import ArrayLike
+from qutip import Qobj
+
+from .._utils import is_batched_scalar
+from .dataarray import DataArray, IndexType
+from .layout import Layout
+from .qarray import QArray, QArrayLike, check_compatible_dims, isqarraylike, to_jax
+
+__all__ = []
+
+
+class MaterializedQArray(QArray):
+    vectorized: bool = eqx.field(static=True)
+    data: DataArray
+
+    def __check_init__(self):
+        # === ensure dims is a tuple of ints
+        if not isinstance(self.dims, tuple) or not all(
+            isinstance(d, int) for d in self.dims
+        ):
+            raise TypeError(
+                f'Argument `dims` must be a tuple of ints, but is {self.dims}.'
+            )
+
+        # === ensure dims is compatible with the shape
+        # for vectorized superoperators, we allow that the shape is the square
+        # of the product of all dims
+        shape = self.data.shape
+        allowed_shapes = (prod(self.dims), prod(self.dims) ** 2)
+        if not (shape[-1] in allowed_shapes or shape[-2] in allowed_shapes):
+            raise ValueError(
+                'Argument `dims` must be compatible with the shape of the qarray, but '
+                f'got dims {self.dims} and shape {shape}.'
+            )
+
+    @property
+    def dtype(self) -> jnp.dtype:
+        return self.data.dtype
+
+    @property
+    def layout(self) -> Layout:
+        return self.data.layout
+
+    @property
+    def shape(self) -> tuple[int, ...]:
+        return self.data.shape
+
+    @property
+    def mT(self) -> QArray:
+        return replace(self, data=self.data.mT)
+
+    @property
+    def ndim(self) -> int:
+        return self.data.ndim
+
+    @property
+    def ndiags(self) -> int:
+        """Number of stored diagonals (only for sparse diagonal layout)."""
+        if not hasattr(self.data, 'ndiags'):
+            raise AttributeError(
+                f"Attribute 'ndiags' is only defined for sparse diagonal layouts; "
+                f'got layout {self.layout!r}.'
+            )
+        return self.data.ndiags
+
+    # === Array methods delegated to DataArray ===
+
+    def conj(self) -> QArray:
+        """Returns the element-wise complex conjugate of the qarray.
+
+        Returns:
+            New qarray with element-wise complex conjuguated values.
+        """
+        return replace(self, data=self.data.conj())
+
+    def reshape(self, *shape: int) -> QArray:
+        """Returns a reshaped copy of a qarray.
+
+        Args:
+            *shape: New shape, which must match the original size.
+
+        Returns:
+            New qarray with the given shape.
+        """
+        return replace(self, data=self.data.reshape(*shape))
+
+    def _reshape_unchecked(self, *shape: int) -> QArray:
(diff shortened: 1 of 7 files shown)


Nicolas-Lepage:
If this approach is suitable, I'd suggest #1091 keep only the new `composite_qarray.py`, then `hard-reset` the branch onto this one after review and add `composite_qarray.py` back on top. From there it's a matter of filling in the base-class design contract on the `CompositeQArray` side, deciding its specific attributes, properties, and methods,  which should be fairly contained.

dynamiqs/qarrays/qarray.py
```diff
@@ -212,79 +211,57 @@ class QArray(eqx.Module):
...
 
+    @abstractmethod
     def __check_init__(self):
-        # === ensure dims is a tuple of ints
-        if not isinstance(self.dims, tuple) or not all(
-            isinstance(d, int) for d in self.dims
-        ):
-            raise TypeError(
-                f'Argument `dims` must be a tuple of ints, but is {self.dims}.'
-            )
```
gautierronan, line 230: This should probably still be inside `QArray` since `dims` is a property of `QArray` (they will all be concerned by this check). Also, the method probably doesn't need to be astract here (the children classes do not need to implement `__check_init__` to be valid).

dynamiqs/qarrays/qarray.py
```diff
@@ -212,79 +211,57 @@ class QArray(eqx.Module):
     """  # noqa: E501
 
     dims: tuple[int, ...] = eqx.field(static=True)
-    vectorized: bool = eqx.field(static=True)
```
gautierronan, line 215: Wouldn't `vectorized` be compatible with `CompositeQArray`? I didn't think about it too much, but my first guess is that it should. It might be a bit tricky to set up however, we should probably discuss this over the board.

For now I'm ok with keeping it only inside `MaterializedQArray`.

gautierronan (changes requested):
The overall direction sounds good to me. I'm completely aligned with the class structure that you propose.

I have one minor comment below about `__check_init__` and one less minor about the `vectorized` attribute, but it's probably simpler to discuss it live (unless you have clear ideas already). We can merge without this second comment to not block progress.

dynamiqs/qarrays/qarray.py
Nicolas-Lepage: I wasn't sure about this point. 

It seems quite niche to want to vectorize and use a composite Qarray at the same time, but the math still seems doable. 
in addition, I think tracking positions and shapes can get quite complicated when combining composite + vectorized. 

I think I can note the comment and, for now, proceed with the two options being mutually exclusive, then look into how to make them compatible in a later PR, once the base implementation of the composite Qarray is finished.

dynamiqs/qarrays/qarray.py
Nicolas-Lepage: Thanks for the review :) 
I wasn't sure whether there are any composite Qarray specific checks we'd want to do. The only thing I can think of is that we'd want to create a compositeQarray object only if the dims indicate that we have several modes (len(dims) > 1? I forget exactly how dims and shape work).

dynamiqs/qarrays/qarray.py
gautierronan: I agree. Can you just add a `TODO` inside the code for later?

dynamiqs/qarrays/qarray.py
gautierronan: You can still extend `__check_init__` inside `CompositeQArray` if you want to, by calling the parent method and then adding your checks. The only thing is that we should avoid duplicate code, so checks that concern all qarrays should be inside `QArray`

Nicolas-Lepage:
@gautierronan there's another subject I forgot to mention. 

I wasn't exactly sure what should be made public and where the documentation should go. 

The main idea, for me, is that the change shouldn't be noticeable to the user, so they don't need to know that Qarray has two concrete child classes. They just need to know that Qarray exists. This points to the docs staying as they are. 

On the other hand, making a base class public seems strange to me. 

For the moment, the docs are unchanged and only the Qarray base class is public — but there were still some tests that needed to reference the child class.

dynamiqs/qarrays/qarray.py
Nicolas-Lepage: i have moved the first part of `__check_init__` to base class

dynamiqs/qarrays/qarray.py
Nicolas-Lepage: i have added the todo comment

gautierronan:
Indeed, here I think we're cheating a bit in some sense because we're documenting QArray but it's not really supposed to be called by users. I think we have two choices that make sense:

1. We only document QArray: it's an abstract class but users need to know about the different methods and attributes it has. There, we can talk about the dense/sparsedia data arrays, and possibly also about the materialized/composite.
2. We document everything: QArray, its children classes MaterializedQArray and CompositeQArray, but also DataArray, DenseDataArray, SparseDIADataArray. It's the most clean and clear, but also exposes quite a bunch of stuff that we don't necessarily want users to know about. Maybe this should come together with a small tutorial about QArrays in general.

In principle, the `Python API` section of the documentation should replicate the internal structure and document everything that is in the public API. That being said, my personal philosophy is that the documentation is made for humans, so whatever works best to make it clear how to use the library is what should be prioritized.

dynamiqs/qarrays/materialized_qarray.py
```diff
@@ -23,14 +23,6 @@ class MaterializedQArray(QArray):
     data: DataArray
 
     def __check_init__(self):
```
gautierronan, line 25: Here, you need to call the parent method, otherwise the method is just overriden. You can do it through `super().__check_init__()`.

gautierronan (changes requested):

dynamiqs/qarrays/materialized_qarray.py
Nicolas-Lepage: thanks, i am not super good with python classes . I have made the changes

gautierronan (approved):
LGTM!

PR #1103, opened 2026-06-05
Add a benchmark suite for Dynamiqs solvers

### LINKED ISSUE #1083: unitaryHACK - Build a benchmark suite for Dynamiqs solvers

slopesque:
# State of the feature

The feature is still in WIP. However, as the size of the feature is of considerable size, a draft pull request has been opened in advance to leave the current architecture and implementation up to discussions with package maintainers. 

# Summary

This pull request implements a benchmark workflow for quantum solvers, as requested by #1083 .

On top of the requested features, it defines a first yet structured, scalable architecture that may allow : 

- Easier extension of both benchmarked problems (beyond `sesolve`/`mesolve`-based) and solver methods.
- Convenient addition of new benchmark steps as the module continues to satisfy potential future needs. 

As of now, the benchmark currently returns :

- The name of the problem that was solved
- The name of the method that was used
- The parameters of the method (so far: `atol`, `rtol`, `factor_min`, `factor_max`, `factor_safety`. Metrics for non-adaptative methods are still a WIP)
- The number of solver steps
- The solver runtime duration
- The processing backend (`cpu`, `gpu` or `tpu`)
- The package version for `diffrax`, `dynamiqs`, `jax`

# Usage

One can run the benchmark engine by simply calling the folder as a module :

```sh
python -m tests.benchmark
```

The full interface is yet undefined and can be brought to discussions with collaborators to elaborate on a more extensive defintion.

# Changes

The current benchmark architecture is composed of 3 main submodules :

- `tests/benchmark/problems/base` : Define common bases for `Problem`s (ex : solving a problem based on an implementation for `build`)
- `tests/benchmark/problems/concrete`: Children classes defining concrete `Problem`s that will be used for the benchmark
-  `tests/benchmark/utils`: Utilitary functions for benchmark functions
- `tests/benchmark/workflow`:  Implementation of main benchmark steps (solve a problem with a method, retrieve the metrics, return a benchmark record)
- `tests/benchmark/[__init__,__main__,main].py`: Entrypoints for the benchmark

# Yet to be done

As contribution guidelines request pull requests to be as modular as possible, future tasks may be implemented in sequential pull requests should the first structure be considered convenient for the need raised by #1083 . They may also be implemented along with the other changes broughts by the pull request :

- Integrate the problems 2,3,4,5,6 requested by #1083 
- Integrate all the solvers that should be benchmarked (for now, the benchmark only runs one sample solver with the default method)
- Implement computation of reference metric
- Implement computation accuracy vs reference
- Implement status pass/fail flag
- Implement optional benchmark metrics (runtime memory usage, gradiant accuracy)
- Implement benchmark record conversion to CSV
- Implement deterministic random seed during benchmark
- Refine leaderboard output format elegance (for now, the benchmark only prints a pretty-print version of the record dictionnary)
- Refine documentation

Closes #1083


Nicolas-Lepage:
Hi @mccreemainwoody, is everything going ok with the PR? I took the liberty of checking the direction you are going in even if you aren't finished, and I think I already have a few pointers / comments.

Architecture:
I think creating an abstract contract through a base class for each problem is a good idea. I would say each problem needs to:
1. build the problem.
2. declare the methods that apply to it and their parameters. Not all solvers allow the same set of methods, and I think the problem itself should declare what it supports rather than the runner inferring it from tags. For all adaptive step methods I recommend sticking to the defaults. For fixed step methods, run several times with different dt in a good range.
3. define its own reference, based on my issue comment. If applicable prioritise the analytic solution, then second best is Expm but applicable only for small / medium size problems, and finally if that does not work define a reference that is ideally double precision with a very low tolerance high order method. Note this is the one place tolerances should differ from the defaults above.

Outside of this I think the structure is heavier than it needs to be.

One thing to watch on precision: dq.set_precision("double") affects everything and should be set first, so if we want the benchmark to choose between single and double precision this needs to be handled carefully.

For timing, be sure to use block_until_ready() because jax is asynchronous. Otherwise you just time how long it takes to launch the simulation but not the time of the simulation itself.

For error I would suggest using 1 minus dq.fidelity and L2 error on states at several times during the simulation for now.

For batched problems I suggest recording both of these, and for nsteps and errors keeping three values: min time / min error, max time_steps / max error, and average time_steps / average error.

slopesque:
Hey @Nicolas-Lepage , thank you so much for the feedback !! I'll be sure to update the changes with what you requested.

Studies kept me busier than expected recently, so I could not go as fast as I originally wanted. That being said, I still could continue on the PR locally _(implementing the references, computing the errors and adding the other problems)_. The point of the draft was to validate the architecture before going forward with the other changes, to follow the contribution guidelines. :)

Do you have parts of the architecture in mind that could be lightened ? The goal of the approach was mostly to make the architecture more scalable, so that one may easily add extra components in the future _(add new problems, benchmark other solver methods, add other benchmark steps...)_. However, perhaps this aspect is just not needed yet ahah.

gautierronan:
Thank you for your contribution. Unfortunately, we will be going for an in-house implementation (see https://github.com/dynamiqs/dynamiqs/pull/1146) as none of the open contributions were meeting our needs. Thanks again for taking the time to work on this.

PR #1104, opened 2026-06-06
add property tests for stochastic solvers (#1079)

### LINKED ISSUE #1079: unitaryHACK - Add tests for stochastic solvers
### Context

Stochastic Schrödinger equation (SSE) and stochastic master equation (SME) solvers are used to simulate continuously measured open quantum systems through individual quantum trajectories, not only ensemble-averaged evolution.

Beyond their physical interpretation, SSE solvers are also an important computational tool for open quantum systems. Since each trajectory evolves independently and propagates a state vector rather than a density matrix, trajectory-based methods can provide more favorable scaling and are naturally highly parallelizable compared to deterministic master-equation evolution.

Different stochastic unravelings (different stochastic trajectory descriptions of the same Lindblad master equation corresponding to different measurement processes) correspond to different physical measurement schemes:
- jump solvers model discrete detection events (e.g. photon counting);
- diffusive solvers model continuous weak measurements (e.g. homodyne detection).

A crucial physical requirement is that averaging many stochastic trajectories must recover the deterministic Lindblad evolution, while individual trajectories still exhibit the correct stochastic back-action and fluctuations.

Because of this, stochastic solvers are significantly harder to validate than deterministic solvers. Correctness must be checked both at the trajectory level and at the statistical ensemble level. The tests proposed in this issue are designed to verify these two aspects simultaneously.

This issue aims to add a small but meaningful test suite for stochastic solvers, covering all four stochastic APIs:

- [jump stochastic Schrödinger equation solver](https://www.dynamiqs.org/stable/python_api/integrators/jssesolve.html)
- [diffusive stochastic Schrödinger equation solver](https://www.dynamiqs.org/stable/python_api/integrators/dssesolve.html)
- [jump stochastic master equation solver](https://www.dynamiqs.org/stable/python_api/integrators/jsmesolve.html)
- [diffusive stochastic master equation solver](https://www.dynamiqs.org/stable/python_api/integrators/dsmesolve.html)

The goal is to test physically meaningful properties of the stochastic dynamics, not only that the solvers run without errors.

[existing stochastic solver test](https://github.com/dynamiqs/dynamiqs/blob/main/tests/jssesolve/test_jssesolve.py)
The current stochastic solver tests mainly verify first-order statistical convergence: averaging many stochastic trajectories should reproduce the corresponding deterministic `mesolve` result.

What is still missing are:
- analytical trajectory-level verification, checking that individual stochastic trajectories behave correctly in situations where the exact trajectory is known;
- analytical statistical tests validating both first- and second-order statistics directly against known theoretical results, rather than only against `mesolve`.

### What we want

Add tests that validate different aspects of stochastic solver correctness.

This includes extending the existing ensemble-convergence tests against `mesolve` to all four stochastic solvers.

Ideally, each test should cover all four solvers using the same physical problem. When the stochastic unravelings differ, the checked quantity may differ between solvers, but the underlying test problem should remain the same.

The tests should be non-trivial but lightweight enough to run reliably in CI. They should use a meaningful number of trajectories and saved time points, while keeping runtime reasonable.

### Test 1: No-back-action trajectory test

Add a test for a protected subspace where the stochastic measurement term should not introduce any back-action.

In this case, if the stochastic solvers are implemented correctly, all trajectories should collapse to the same deterministic analytical trajectory. This test checks that the stochastic part cancels exactly when it physically should.

This test should be implemented for all four stochastic solvers.

**A concrete proposed setup for this test is given below** 

### Test 2: Statistical ensemble test

Add a statistical test where the stochastic process has known analytical first- and second-order statistics.

The test should verify that, over many trajectories:

- the ensemble average follows the expected analytical mean;
- the ensemble variance follows the expected analytical variance;
- the accuracy is consistent with the number of trajectories.

This test checks that the solver produces the correct stochastic ensemble statistics, not only the correct deterministic average evolution.

The preferred setup is a single physical problem that can be tested across all four solvers. If necessary, the exact analytical formula may differ between jump and diffusive unravelings, but the tested physical situation should remain common.

### Acceptance criteria

- Existing ensemble-convergence tests against `mesolve` are extended to all four stochastic solvers.
- Two stochastic solver tests are added:
  - one no-back-action trajectory test;
  - one first- and second-order statistical ensemble test.
- Both tests cover all four stochastic solvers.
- Tests use fixed random seeds.
- Tests are meaningful but CI-friendly.
- Tests use enough trajectories and saved points to check non-trivial stochastic behavior.
- Tests would fail for incorrect stochastic back-action, incorrect trajectory statistics, or incorrect ensemble statistics.

### Focus 

Mixed between Physics understanding and coding :  Understanding solver behavior + testing practices

### Difficulty 

Hard

---

### Proposed setup for Test 1

**System :**

Use a two-qubit system with observation operator:

$$
L = -\sigma_z \otimes \sigma_z
$$

and Hamiltonian:

$$
H =
\frac{\omega}{2}
\left(
\sigma_x \otimes \sigma_x
+
\sigma_y \otimes \sigma_y
\right)
$$

Initial state:

$$
|\psi(0)\rangle = |01\rangle
$$

**Expected behavior :**

The dynamics remain in the odd-parity subspace spanned by:

$$
|01\rangle,\ |10\rangle
$$

In this subspace, the observation operator acts as the identity, so it should not introduce any stochastic back-action.

The expected trajectory is the deterministic evolution:

$$
|\psi(t)\rangle = \cos(\omega t)|01\rangle - i\sin(\omega t)|10\rangle
$$

**Checks :**

For each stochastic trajectory and time step, verify:

- agreement with the analytical state:

$$
1 -
\left|
\langle
\psi_{\mathrm{exact}}(t)
|
\psi_{\mathrm{stoch}}(t)
\rangle
\right|^2
\approx 0
$$

---

### Possible ideas for Test 2

One simple starting point is qubit de-excitation:

$$
H = 0,
\qquad
L = \sqrt{\gamma}\sigma_-,
\qquad
|\psi(0)\rangle = |e\rangle .
$$

For the jump stochastic solvers, this gives a clean Bernoulli process: each trajectory is either still excited or has emitted a jump. Therefore the excited-state population satisfies

$$
P_e^{(r)}(t) \in \{0, 1\},
\qquad
P_e^{(r)}(t) \sim \mathrm{Bernoulli}(e^{-\gamma t}).
$$

So the ensemble mean and variance are known:

$$
\mathbb{E}[P_e(t)] = e^{-\gamma t},
\qquad
\mathrm{Var}[P_e(t)] =
e^{-\gamma t}\left(1-e^{-\gamma t}\right).
$$

This is a good candidate for `jssesolve` and `jsmesolve`, using either the excited-state population or the jump information from `result.clicktimes`.

For the diffusive solvers, the same Bernoulli statistics do not generally apply because the unraveling is different. A similar test should be found where the first- and second-order statistics are known for the diffusive measurement record.

Ronit-Raj9:
Closes #1079.

Adds a small property-test suite for the four stochastic solvers (`jssesolve`, `dssesolve`, `jsmesolve`, `dsmesolve`). The tests are parametrized over all four solvers through a shared runner, so the same physical problem is checked for each unraveling instead of duplicating four near-identical files.

Files (all under `tests/stochastic/`):
- `utils.py` — solver runners (common signature) and physical systems
- `test_properties.py` — the tests

### Convergence to `mesolve` (`test_convergence_to_mesolve`)

Averaging trajectories recovers the Lindblad evolution, checked for both expectation values and states. This extends the existing jsse-only check to all four solvers.

### No back-action (`test_no_backaction`)

Uses the setup proposed in the issue: `H = (ω/2)(σx⊗σx + σy⊗σy)`, `L = -σz⊗σz`, `ψ0 = |01⟩`. On the odd-parity subspace `{|01⟩, |10⟩}`, the measurement acts as the identity, so every trajectory should follow the deterministic curve `cos(ωt)|01⟩ - i·sin(ωt)|10⟩`. Checks `1 - |⟨ψ_exact|ψ⟩|²` for each trajectory at each saved time.

### First/second-order statistics (`test_statistics`)

One common problem for all four solvers: QND measurement of `σz` on a `σz` eigenstate (`H = 0`, `L = √γ σz`, `ψ0 = |e⟩`). The state is a fixed point, so the only randomness is in the measurement record, with known moments:

- **Jump solvers:** click count over `[0, T)` is `Poisson(γT)`, so `mean = var = γT` (from `result.nclicks`)
- **Diffusive solvers:** time-averaged record is Gaussian with `mean = 2√γ` and `var = 1/Δt` (from `result.measurements`)

Also includes the issue's Bernoulli de-excitation example as an extra jump-only check (`test_jump_bernoulli_statistics`): `H = 0`, `L = √γ σ-`, `ψ0 = |e⟩`, where the excited-state population is `Bernoulli(e^{-γt})`.

## Negative controls

To ensure the assertions catch wrong physics and don't pass trivially:

- **`test_backaction_is_detected`** — same Hamiltonian but `L = σz⊗I`, which is not the identity on the subspace. Back-action shows up (max infidelity ≈ 1), so the no-back-action assertion correctly fails here.
- **`test_statistics_reject_wrong_rate`** — measured click rate rejects a wrong rate (`2γ` instead of `γ`).

## Notes

- The checked quantity in `test_statistics` differs by unraveling: Poisson click count for jump solvers, Gaussian record statistics for diffusive solvers. Bernoulli population statistics only apply to the jump unraveling, which is why the shared test uses the QND measurement record. For diffusive solvers the QND state is a fixed point, so record mean/variance is checked, not state population.
- Statistics tolerances are **5σ Monte Carlo confidence bounds**, scaling as `1/√N` with the number of samples (addresses the "accuracy consistent with the number of trajectories" requirement from the issue).
- Convergence tolerance is slightly looser than existing jsse tests because diffusive solvers use weak-order-1 `EulerMaruyama` with `dt=1e-3`; its ensemble-average bias (~1e-2) dominates here.
- Fixed seeds throughout. Tests are marked `TEST_LONG` and use 2000–4000 trajectories; the suite runs in well under a minute locally.

## Test plan

- [x] `uv run pytest tests/stochastic/ -v` — 20/20 passed locally (~41s)
- [x] `ruff check` and `ruff format --check` clean
- [x] CI green on this PR


tests/stochastic/utils.py
```diff
@@ -0,0 +1,125 @@
...
+def run_dsme(H, jump_ops, psi0, tsave, keys, exp_ops):
+    etas = jnp.ones(len(jump_ops))
+    method = dq.method.EulerMaruyama(dt=DT)
+    return dq.dsmesolve(
+        H, jump_ops, etas, psi0, tsave, keys=keys, exp_ops=exp_ops, method=method
+    )
+
+
+SOLVERS = {'jsse': run_jsse, 'dsse': run_dsse, 'jsme': run_jsme, 'dsme': run_dsme}
+JUMP_SOLVERS = ['jsse', 'jsme']
```
Nicolas-Lepage: These helpers are useful :) 

Can you please add 4 new types of "runs" to be even more complete 

- `jsse_event_no_smart` and `jsse_event_smart`: `Event` method with `smart_sampling`=`False` and `True` respectively, using the same `root_finder` setup as in the : [jssesolve test ](https://github.com/dynamiqs/dynamiqs/blob/main/tests/jssesolve/test_jssesolve.py)
- for dsse add the `Rouchon1()` with same DT as euler
- for dsme add the `Rouchon1()` with same DT as euler

Note: during review I observed that `test_statistics` fails for jsse with `Event(smart_sampling=True)` — the click count mean is off by 7σ. This looks like a solver bug rather than a test bug and should be opened as a separate issue. The test is correct to catch it.

tests/stochastic/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    bra = jnp.conj(jnp.swapaxes(exact, -1, -2))  # (ntsave, 1, n)
+    if states.shape[-1] == 1:  # ket (SSE)
+        overlap = (bra @ states)[..., 0, 0]  # (ntrajs, ntsave)
+        bra_psi = jnp.conj(jnp.swapaxes(states, -1, -2))
+        norm2 = (bra_psi @ states)[..., 0, 0].real
+        return 1 - jnp.abs(overlap) ** 2 / norm2
+    # density matrix (SME)
+    num = (bra @ states @ exact)[..., 0, 0].real  # (ntrajs, ntsave)
+    tr = jnp.trace(states, axis1=-2, axis2=-1).real
+    return 1 - num / tr
```
Nicolas-Lepage: The normalized infidelity is a good directional check but it hides normalization errors by construction. Please keep it, but add a complementary check that verifies norm preservation explicitly:
```
# norm should be preserved at every trajectory and time step
assert jnp.allclose(norm2, 1.0, atol=atol_norm)
```
This decomposes solver correctness into two orthogonal checks:
- direction: is the state pointing the right way? (existing normalized infidelity)
- scale: is probability conserved? (new norm check)

tests/stochastic/test_properties.py
```diff
@@ -0,0 +1,172 @@
...
+    omega = 1.0
+    H, jump_ops, psi0 = backaction_system(omega)
+    tsave = jnp.linspace(0.0, 1.0, 11)
+
+    keys = jax.random.split(jax.random.key(6), num=50)
+    result = SOLVERS[solver](H, jump_ops, psi0, tsave, keys, None)
+
+    exact = dq.stack([protected_subspace_state(t.item(), omega) for t in tsave])
+    infidelity = infidelity_with_state(result, exact)
+    assert jnp.max(infidelity) > atol
```
Nicolas-Lepage: This test is correctly checking fidelity to the analytical solution. To also test whether there is back-action, we could add an assert statement checking that the infidelity between a trajectory and all others is close to machine precision.

If there is truly no back-action then all errors should come from the integration method and this should be the same for all trajectories.

Nicolas-Lepage (changes requested):
This is a great PR, it implements everything the issue asks for and finds a good common physical problem for the first and second order statistics test.

In addition to the 3 comments I have left, please address one last general point:
I have run your tests with verbose printing of all asserted quantities and find that for most tests  (especially convergence to mesolve, no back-action, and first order statistics ) the tolerances can be tightened significantly. In some cases the actual errors are orders of magnitude below the current threshold, in the no back-action test as low as near machine precision.

Please tighten the tolerances to reflect the actual errors observed, on top of addressing the other comments. 

Thanks for the work !

Ronit-Raj9:
Thanks for the detailed review! Addressed all of it in the latest commit.

tests/stochastic/utils.py
Ronit-Raj9: implemented all four:
  - `jsse_event_no_smart` / `jsse_event_smart` — `Event(root_finder=optx.Newton(1e-4, 1e-4,  rms_norm), smart_sampling=False/True)`, same root finder as `test_jssesolve.py`
  - `dsse` and `dsme` each get a `Rouchon1(dt=DT)` runner

  On the `smart_sampling=True` failure - confirmed it's a solver bug, opened #1113. It biases the raw per-trajectory click ensemble, so it actually fails both `test_statistics` (Poisson count, ~20σ at N=4000) and `test_jump_bernoulli_statistics` (population). `mean_expects()`/`mean_states()` are corrected for it, so convergence and no-back-action still pass. Both affected parametrizations are marked `xfail(strict=True)` pointing at #1113, so a fix will show up as an xpass.

tests/stochastic/utils.py
Ronit-Raj9: Added. Kept the normalized infidelity for the direction check, and added a `trajectory_norms(result)` helper (returns `<psi|psi>` for kets, `Tr[rho]` for density matrices) with an explicit scale check in `test_no_backaction`:

  ```python
  assert jnp.allclose(trajectory_norms(result), 1.0, atol=atol_norm)  # atol_norm=2e-2
  ```

  Good call - it surfaces a real difference between the methods: the fixed-step Euler solvers `EulerJump`/`EulerMaruyama`) only conserve the norm to O(dt) (~8e-3 here), while the norm-preserving `Rouchon`/`Event` methods hold it to ~1e-6. The 2e-2 tolerance sits just above the Euler drift, so it still catches gross probability non-conservation.

tests/stochastic/test_properties.py
Ronit-Raj9: Added - this is the cross-trajectory check you described. New `cross_trajectory_infidelity(result)` helper computes the infidelity of every trajectory against the first one at each saved time (kets via normalized overlap, density matrices via `Tr[rho_0 rho_i]`).

  With no back-action every trajectory follows the same deterministic evolution, so they only differ at the
  integrator's floating-point level - in `test_no_backaction`:

  ```python
  assert jnp.all(cross_trajectory_infidelity(result) < atol)  #
  atol=1e-4, measured ~0–2e-5
  ```

  I also used it in the negative control (`test_backaction_is_detected`): with real back-action the trajectories diverge from each other, so there I assert it's large (`> 0.1`, measured ~0.9). That gives the two ends of the  check - identical trajectories when the measurement is harmless, divergent ones when it isn't.

tests/stochastic/utils.py
Nicolas-Lepage: Thanks for the modification, this is great work !

Nicolas-Lepage (dismissed):
This is great work ! thanks for the tests and helping catch a solver issue !

Ronit-Raj9:
Since you finally approved that issue is correct... should i go on raising PR for that... i had solved it locally.

Nicolas-Lepage:
@Ronit-Raj9 i am just waiting for a colleague to give his approval as well. If he is ok i will merge and you will receive the bounty. And i am not sure if i understood your last message but you can open a PR for the issue raised by the tests, we welcome the help gladly, but just as a disclaimer : it will not be in the hackathon.

Ronit-Raj9:
Yeah, I understood it won't be part of hackathon... its just i had fixed it locally so thought to raise the PR.

Nicolas-Lepage:
Sorry, I had not understood your message earlier. We welcome your fix! You can go ahead and open a new PR!

tests/stochastic/test_properties.py
```diff
@@ -0,0 +1,212 @@
...
+    jump_ops = [a, 0.3 * a.dag()]
+    psi0 = dq.coherent(n, 0.5)
+    exp_ops = [a.dag() @ a, a + a.dag()]
+    tsave = jnp.linspace(0.0, 1.0, 6)
+
+    keys = jax.random.split(jax.random.key(1), num=2000)
+    sresult = SOLVERS[solver](H, jump_ops, psi0, tsave, keys, exp_ops)
+    meresult = dq.mesolve(
+        H, jump_ops, psi0, tsave, exp_ops=exp_ops, progress_meter=None
+    )
```
gautierronan: Ideally, we should test against an analytical result (as we do for non-stochastic solvers) instead of `mesolve`. It would be faster, and it would be more robust to a possible bug that would affect both `mesolve` and stochastic solvers at the same time.

gautierronan (changes requested):
Hi @Ronit-Raj9, thanks a lot for your contribution to the library. I like the overall PR content and the variety of tests: it's indeed great that it caught a non-trivial bug on the smart_sampling method.

However, the current implementation deviates significantly from the current structure of the `tests` folder. Could you thus refactor your PR to align with this structure before merging? Roughly speaking, you could:
- Introduce a `StochasticTester` class inside `tests/stochastic_tester.py`, similarly to the existing `IntegratorTester`. This would host the various test functions that you later use for each stochastic solver.
- Introduce a file `tests/systems/stochastic_system.py` where you can cleanly instanciate your various "physical systems"
- Have at least one test file per solver, for instance `tests/jsmesolve/test_jsmesolve.py` where you call the various tests functions on the various systems that you have defined above.

Note that a `test_jssesolve.py` file already exists, but feel free to completely override it if it becomes redundant with your new tests.

tests/stochastic/test_properties.py
Ronit-Raj9: Done - dropped `mesolve` entirely. Convergence now checks the trajectory average against the analytical Lindblad result of a driven-damped oscillator: `<a>(t) = α_ss(1 − e^{−st})`, `s = κ/2 + iω`, and the full state against the analytical  coherent state `|α(t)⟩⟨α(t)|` (the oscillator stays coherent under amplitude damping) - mirroring how the deterministic solvers check both `states` and `expects`. Faster (no solver call for the reference) and independent of any solver, as you suggested.

Ronit-Raj9:
Thanks @gautierronan! Refactored to match the repo structure:
  
  - `tests/stochastic_tester.py` - `StochasticTester` class hosting the test functions (same shape as `IntegratorTester`, `_test_*(self, system, method)`)
  - `tests/systems/stochastic_system.py` - the physical systems as `System` subclasses (`ProtectedSubspace`, `DampedOscillator`, `DecayQubit`, `QNDQubit`, `BackactionQubit`), instantiated at the bottom
  - one test file per solver (`tests/{jssesolve,dssesolve,jsmesolve,dsmesolve}/test_*.py`), each a `StochasticTester` subclass that wires the systems + that solver's methods; the old `test_jssesolve.py` is overridden
  
  All references are now analytical (no `mesolve`) - see the reply on your inline comment. Rebased on latest main and re-ran locally: green (the 2 `xfail`s are the smart_sampling bug #1113).

tests/stochastic_tester.py
```diff
@@ -0,0 +1,203 @@
...
+    bra = jnp.conj(jnp.swapaxes(exact, -1, -2))  # (ntsave, 1, n)
+    if states.shape[-1] == 1:  # ket (SSE)
+        overlap = (bra @ states)[..., 0, 0]  # (ntrajs, ntsave)
+        bra_psi = jnp.conj(jnp.swapaxes(states, -1, -2))
+        norm2 = (bra_psi @ states)[..., 0, 0].real
+        return 1 - jnp.abs(overlap) ** 2 / norm2
+    # density matrix (SME)
+    num = (bra @ states @ exact)[..., 0, 0].real  # (ntrajs, ntsave)
+    tr = jnp.trace(states, axis1=-2, axis2=-1).real
+    return 1 - num / tr
```
gautierronan: I think you can essentially replace this helper with a single call to `dq.overlap`, and avoid conversion to jax arrays.

tests/stochastic_tester.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def trajectory_norms(result) -> Array:
+    # state norm per trajectory and time: <psi|psi> for kets, Tr[rho] for density
+    # matrices. Probability conservation requires this to stay 1.
+    states = result.states.to_jax()
+    if states.shape[-1] == 1:
+        bra = jnp.conj(jnp.swapaxes(states, -1, -2))
+        return (bra @ states)[..., 0, 0].real
+    return jnp.trace(states, axis1=-2, axis2=-1).real
```
gautierronan: Same, I think you can replace this with `dq.norm` or `dq.trace`.

tests/stochastic_tester.py
```diff
@@ -0,0 +1,203 @@
...
+    if states.shape[-1] == 1:  # ket
+        states = states / jnp.linalg.norm(states, axis=-2, keepdims=True)
+        ref = states[0:1]
+        overlap = jnp.abs((jnp.conj(jnp.swapaxes(ref, -1, -2)) @ states)[..., 0, 0])
+        return 1 - overlap**2
+    # density matrix: 1 - Tr[rho_0 rho_i] with unit-trace normalisation
+    tr = jnp.trace(states, axis1=-2, axis2=-1).real[..., None, None]
+    states = states / tr
+    ref = states[0:1]
+    return 1 - jnp.trace(ref @ states, axis1=-2, axis2=-1).real
```
gautierronan: Same, internal implementation should use dynamiqs utilities (probably `dq.overlap` or `dq.fidelity` here).

tests/stochastic_tester.py
```diff
@@ -0,0 +1,203 @@
...
+        seed: int = 5,
+    ):
+        # click count over [0, T) is Poisson(gamma T): mean = var = gamma T. The
+        # tolerances are 5-sigma Monte Carlo confidence bounds (scale as 1/sqrt(N)).
+        result = system.run(self.SOLVER, method, _keys(seed, ntrajs))
+        nclicks = result.nclicks[..., 0]
+        n = nclicks.shape[0]
+        lam = system.poisson_lambda()
+        sem = jnp.sqrt(lam / n)
+        sev = jnp.sqrt((lam + 2 * lam**2) / n)
```
gautierronan: nit: can you avoid unnecessary abbreviations here? It makes it easier to read and thus maintain the code. e.g. `lam` -> `poisson_lambda`, `sem` -> `expected_mean` (might not be the best name, but something along those lines), `sev` -> `expected_variance` (same).

tests/stochastic_tester.py
```diff
@@ -0,0 +1,203 @@
...
+        # time-averaged record I = <L+Ld> + dW/dt is Gaussian, with analytically
+        # known mean and variance; 5-sigma Monte Carlo confidence bounds
+        result = system.run(self.SOLVER, method, _keys(seed, ntrajs))
+        samples = result.measurements[..., 0, :].reshape(-1)
+        n = samples.shape[0]
+        mean, var = system.record_mean(), system.record_variance()
+        sem = jnp.sqrt(var / n)
+        sev = var * jnp.sqrt(2 / n)
+        assert jnp.abs(samples.mean() - mean) < 5 * sem
+        assert jnp.abs(samples.var() - var) < 5 * sev
```
gautierronan: Same comment

gautierronan (changes requested):
Thanks a lot for the changes @Ronit-Raj9. I really like the PR as is. Great work!

I have added a few minor comments below, and then I'm happy to merge!

tests/systems/stochastic_system.py
```diff
@@ -0,0 +1,242 @@
...
+    solver name, so the same physical system can be checked for every unraveling.
+    The SME solvers use perfect efficiency (eta=1) and no dark counts (theta=0),
+    for which their trajectory-averaged dynamics matches the corresponding SSE.
+    """
+
+    @abstractmethod
+    def Ls(self, params: PyTree) -> list[QArray | TimeQArray]:
+        """Compute the jump operators."""
+
+    def run(self, solver: str, method: Method, keys: Array, **kwargs) -> Result:
```
Nicolas-Lepage: The `**kwargs` is not needed and should be removed to be consistant with the rest of the tests.

Nicolas-Lepage (changes requested):
Thanks again for the changes. this seems much closer to our other tests. I have added one small comment on the arguments of the run function.

tests/systems/stochastic_system.py
Ronit-Raj9: Yupp.. removed **kwargs.

tests/stochastic_tester.py
Ronit-Raj9: Done - renamed to poisson_lambda / mean_standard_error / variance_standard_error.

tests/stochastic_tester.py
Ronit-Raj9: Done - same here: expected_mean / expected_variance / mean_standard_error / variance_standard_error.

tests/stochastic_tester.py
Ronit-Raj9: Done - replaced with 1 - dq.overlap(exact, dq.unit(result.states, psd=True)), no jax conversion. Used psd=True so density matrices normalise by trace - the default .unit() uses the trace-norm and slightly over-normalises a numerically-near-PSD ρ.

tests/stochastic_tester.py
Ronit-Raj9: Done - now just dq.norm(result.states, psd=True) (gives √⟨ψ|ψ⟩ for kets, Tr[ρ] for density matrices).

tests/stochastic_tester.py
Ronit-Raj9: Done - dq.unit(result.states, psd=True) then 1 - dq.overlap(states[0:1], states).

gautierronan (approved):
LGTM!

Nicolas-Lepage (approved):
Thanks @Ronit-Raj9 for this great PR and for being responsive during the review phase!

Ronit-Raj9:
Thank you both! Really appreciate how responsive you were during the review - the quick, specific feedback made the changes fast and kept the context fresh. Learned a lot about the codebase along the way. Glad it caught the smart_sampling bug too.

PR #1105, opened 2026-06-06
Add HigherOrder gradient mode for Hessians

### LINKED ISSUE #1082: unitaryHACK - Enable Hessian computation

spital:
## Summary

Closes #1082 

Adds `dq.gradient.HigherOrder`, a documented higher-order autodiff / Hessian-compatible gradient mode for the existing solver `gradient=` API.

- Routes `HigherOrder` through Diffrax [`DirectAdjoint`](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.DirectAdjoint), which already configures the solver for higher-order differentiation; an earlier explicit `scan_kind="bounded"` override was tested and removed as redundant during review.
- Supports `Euler`, `Dopri5`, `Dopri8`, and `Tsit5`.
- Deliberately does not advertise `HigherOrder` for implicit Kvaerno methods, LowRank, or non-Diffrax methods until those paths are robustly validated (tracked by a TODO in `dynamiqs/method.py`).
- Covers both `dq.sesolve()` and `dq.mesolve()` through the public save/output path (`result.expects`).

## Example

```python
import jax
import dynamiqs as dq


def loss(omega):
    result = dq.sesolve(
        0.5 * omega * dq.sigmax(),
        dq.fock(2, 0),
        [0.0, 0.3],
        exp_ops=[dq.sigmaz()],
        gradient=dq.gradient.HigherOrder(),                                                                                                                                                     )
    return result.expects[0, -1].real
                                                                                                                                                                                            hessian = jax.hessian(loss)(1.3)
```

## Tests

Hessian coverage follows the existing test-suite structure:

- Analytic Hessian references (`hessian_expect`) live on the systems in `tests/systems/` (`Cavity`, `OCavity`, `TDQubit`, `OTDQubit`).
- `IntegratorTester._test_hessian()` compares `jax.hessian` of a `result.expects`-based loss against the analytic Hessian, with default method tolerances.
- `tests/sesolve/test_hessian.py` sweeps `dense_cavity`, `dia_cavity`, and `tdqubit` across `Dopri5`, `Dopri8`, and `Tsit5`.
- `tests/mesolve/test_hessian.py` sweeps `dense_ocavity`, `dia_ocavity`, and `otdqubit` across the same methods.
- `tests/core/test_gradient_support.py` guards the support gate without running solvers: `Euler`/`Dopri5`/`Dopri8`/`Tsit5` accept `HigherOrder`; `Kvaerno3`/`Kvaerno5`/`LowRank`/`Rouchon1` reject it cleanly; the default first-order path still rejects second-order differentiation. (`Euler` is gate-tested only, its fixed-step Hessian is too inaccurate for a meaningful analytic comparison.)

Local verification:

```text
uv run --python 3.11 --with pytest --with-editable . pytest \
  tests/core/test_gradient_support.py \
  tests/sesolve/test_hessian.py \
  tests/mesolve/test_hessian.py -q
# 29 passed

uv run --python 3.11 --with ruff ruff check \
  dynamiqs/gradient.py \
  dynamiqs/method.py \
  dynamiqs/integrators/core/diffrax_integrator.py \
  tests/core/test_gradient_support.py \
  tests/integrator_tester.py \
  tests/sesolve/test_hessian.py \
  tests/mesolve/test_hessian.py \
  tests/systems/_system.py \
  tests/systems/closed_system.py \
  tests/systems/open_system.py
# All checks passed!
```

## Notes for reviewers

For supported explicit Diffrax ODE solvers, `HigherOrder` uses the same `DirectAdjoint` differentiation path that `Direct` uses. The purpose of the new mode is to make the higher-order-AD contract explicit and discoverable through the public Dynamiqs API, as requested in #1082.

The explicit `scan_kind` handling present in earlier revisions was removed after testing: Diffrax `DirectAdjoint` already applies the required bounded scan internally for RK/SRK solvers, so a Dynamiqs-side override would duplicate Diffrax internals and add future API risk.

The support matrix is intentionally conservative: this PR only advertises explicit solvers covered by analytical tests. Implicit Kvaerno and LowRank Hessian support can be revisited in a follow-up once those paths have robust validation.

## AI disclosure

I used Codex, Claude Code, and Hermes Agent as AI assistants to help prototype, review, compare competing PRs, and draft this PR text. I manually reviewed the resulting diff, checked the GitHub issue/PR state, and locally ran the focused Hessian tests plus lint/format checks listed above. The implementation and PR are submitted with human oversight; any remaining mistakes are mine.


dynamiqs/integrators/core/diffrax_integrator.py
```diff
@@ -49,6 +49,23 @@ def __str__(self) -> str:
...
+        isinstance(gradient, HigherOrder)
+        and hasattr(solver, 'scan_kind')
+        and solver.scan_kind is None
+    ):
+        # DirectAdjoint currently applies the same bounded scan internally; setting
+        # it here keeps HigherOrder's Hessian contract explicit and satisfies the
+        # issue's requested solver configuration.
+        return eqx.tree_at(
+            lambda s: s.scan_kind, solver, 'bounded', is_leaf=lambda x: x is None
+        )
```
Nicolas-Lepage: Have you tested removing the explicit scan_kind handling  ? 
if it is not needed then i think we should remove it from the code as this would likely lead to problems in the future if Diffrax API changes.

Nicolas-Lepage (changes requested):
Thanks for the great PR! The implementation seems very solid.

Could you please try removing the explicit `scan_kind` handling and test out your comment about it being redundant? If the adaptive-solver Hessian tests still pass without it, we can drop the whole branch, that simplifies the implementation and removes a point of future Diffrax incompatibility.
If you do drop it, please also revisit the note added to the `Direct` docstring.

Concerning the tests: the logic and coverage are great, I just have a comment on the structure. Right now `tests/test_hessian.py` is a flat top-level file that sidesteps how the rest of the suite is laid out, we don't organise tests this way. Could you make yours fit our structure?

Roughly: the integration-test systems should be defined in `tests/systems/` and checked via something similar to the existing `_test_gradient()` helper, then called from `tests/sesolve/` and `tests/mesolve/`. The pure unit tests should move to the right folders and shouldn't depend on systems.

dynamiqs/integrators/core/diffrax_integrator.py
spital: I'll look into it

gautierronan (changes requested):
Thanks for your contribution @spital! I reviewed the PR and I agree with the comments of @Nicolas-Lepage. I don't have anything else to add :)

dynamiqs/integrators/core/diffrax_integrator.py
spital: Tested it and removed. DirectAdjoint.loop already sets scan_kind="bounded" for RK/SRK solvers when it's None (diffrax _adjoint.py), so hessian_compatible_solver just duplicated that.
Dropping it and re-running the adaptive-solver Hessian tests (Dopri5/Dopri8/Tsit5)
All still pass. Good call on the future-proofing: diffrax has a TODO to deprecate kind="bounded", so this removes that dependency too.

spital:
Done. Analytic Hessian systems now live in `tests/systems/` (hessian_expect on Cavity/OCavity), checked via a new IntegratorTester._test_hessian() alongside `_test_gradient()`, called from `tests/sesolve/` and `tests/mesolve/`. The unsupported-gradient guards moved to `tests/core/test_gradient_support.py` as pure unit tests on the gradient-support gate (no system dependency).
The top-level tests/test_hessian.py is gone.

tests/sesolve/test_hessian.py
```diff
@@ -0,0 +1,27 @@
...
+
+@pytest.mark.run(order=TEST_LONG)
+class TestSESolveHessian(IntegratorTester):
+    @pytest.mark.parametrize('system', [dense_cavity, dia_cavity])
+    @pytest.mark.parametrize(
+        'method',
+        [
+            Dopri5(rtol=1e-8, atol=1e-8),
+            Dopri8(rtol=1e-8, atol=1e-8),
+            Tsit5(rtol=1e-8, atol=1e-8),
```
gautierronan: Is it required that we lower the tolerances here? If so, it would be best to keep the default tolerance of 1e-6, but to increase `rtol` and `atol` inside `_test_hessian` up until the point where tests pass. In general, it's not a good habit to have tolerances that are lower than the standard single precision (roughly 1e-7 - 1e-8).

tests/mesolve/test_hessian.py
```diff
@@ -0,0 +1,19 @@
...
+
+# we only test Tsit5 to keep the unit test suite fast
+
+
+@pytest.mark.run(order=TEST_LONG)
+class TestMESolveHessian(IntegratorTester):
+    @pytest.mark.parametrize('system', [dense_ocavity, dia_ocavity])
+    def test_hessian(self, system):
+        self._test_hessian(
+            system, Tsit5(rtol=1e-8, atol=1e-8), HigherOrder(), rtol=1e-3, atol=1e-4
```
gautierronan: Why do you test only `Tsit5` here, whereas you test `Dopri5` and `Dopri8` as well for `sesolve`? We should chose one or the other: if we're confident that testing only `Tsit5` is sufficient to test all methods, then it's ok to keep only one method. Otherwise, we should test them all.

dynamiqs/gradient.py
```diff
@@ -108,3 +115,24 @@ class Forward(Gradient):
...
+        [`dq.gradient.Direct`][dynamiqs.gradient.Direct], it uses
+        [`diffrax.DirectAdjoint`](https://docs.kidger.site/diffrax/api/adjoints/#diffrax.DirectAdjoint),
+        which already configures the solver for higher-order differentiation. Use this
+        mode to make the higher-order differentiation requirement explicit and to
+        restrict to the methods on which it is supported.
+    """
+
+    # dummy init to have the signature in the documentation
+    def __init__(self):
+        pass
```
gautierronan, line 138: Now that I see this class doesn't actually do anything internally compared to `dq.gradient.Direct`, I wonder whether it's actually useful to introduce it. It may be sufficient to add a proper set of tests and to document `dq.gradient.Direct` as is done in the PR.

WDYT @Nicolas-Lepage @abocquet ?

tests/systems/open_system.py
```diff
@@ -143,6 +143,26 @@ def grads_expect(self, t: float) -> PyTree:
             [grad_x_kappa, grad_p_kappa],
         )
 
+    def hessian_expect(self, t: float) -> PyTree:
```
gautierronan, line 146: Can we also add tests for the `OTDQubit` method? In other words, is it possible to analytically compute `hessian_expect` for this other system? It's quite useful to have both systems working because they are complementary: one tests high-dimensional systems, the other tests time-dependent ones.

gautierronan (changes requested):
Thanks for the changes @spital. I have read through the PR again, and I actually added a few more comments below now that you refactored to follow our internal test structure

dynamiqs/gradient.py
Nicolas-Lepage: This is a good point. 
Currently the implicit solvers and LowRank aren't tested for Hessians, and I think testing them would bloat this PR (this is also debatable). So HigherOrder acts as a guard against running higher-order AD on those untested (though normally compatible) methods.
That leaves two choices: keep this guard, or drop it and add a note to the gradient section of the untested methods saying higher-order AD can work but is untested and therefore experimental.

gautierronan: Hmm, ok that's a good point that I had missed. So indeed, this `HigherOrder` class is essentially here to tell users "you're free to compute the Hessian with any method using `Direct` gradient, but this is at your own risk". Ok then it makes sense to keep this class indeed.

My only concern is that if we end up confirming that higher-order gradients are compatible with all methods that also support `Direct`, then this argument doesn't hold anymore and we might want to remove `HigherOrder` from the public API. This would be a (small-ish) breaking change, which is fine in itself.

Nicolas-Lepage: Agreed, let's keep it. Would a TODO comment in `HigherOrder` on testing implicit/LowRank help ?

tests/sesolve/test_hessian.py
spital: Good point, I reverted the method tolerances to the defaults. The only leaf that didn't clear the old comparison `atol` was Dopri5's analytically-zero `∂²/∂α₀²` (≈1.03e-4), so I loosened the `_test_hessian` comparison to `atol=2e-4` (Dopri8/Tsit5 are ≤7e-5). No more `1e-8` solver tolerances.

tests/mesolve/test_hessian.py
spital: Made them consistent: both sesolve and mesolve now sweep `Dopri5/Dopri8/Tsit5`. With default tolerances all open-system cases match the analytic Hessian to <1e-6, so the extra two methods are essentially free here.

dynamiqs/gradient.py
spital: Ok. keeping `HigherOrder` as the explicit guard for the methods whose higher-order AD path is tested. Added a TODO next to `_DIFFRAX_EXPLICIT_ODE_GRADIENTS` to validate implicit solvers and `LowRank` before broadening it (and to fold it back into `Direct` if they all turn out compatible).

tests/systems/open_system.py
spital: Added `OTDQubit.hessian_expect` (analytic) and included `otdqubit` in the mesolve Hessian sweep. I derived the closed form and checked it against autodiff of the analytic expectation (≈1e-17) and against the solver Hessian (≤7e-6), so it's a genuine analytic reference, not a self-check.

tests/systems/closed_system.py
```diff
@@ -112,6 +112,21 @@ def grads_expect(self, t: float) -> PyTree:
...
+        d2_delta_alpha0 = jnp.array([-t * sdt, -t * cdt])
+        d2_alpha02 = jnp.array([0.0, 0.0])
+
+        return self.Params(
+            self.Params(d2_delta2, d2_delta_alpha0),
+            self.Params(d2_delta_alpha0, d2_alpha02),
+        )
+
 
 class TDQubit(ClosedSystem):
```
gautierronan, line 131: Can you also generalize to `TDQubit` for the same reasons ? Once you have `OTDQubit`, it should be fairly straightforward

Nicolas-Lepage (dismissed):
Thanks @spital for the PR, it's great and I really appreciated the quality of your work! Everything seems good to me.

gautierronan (changes requested):

tests/systems/closed_system.py
spital: OK, I generalized hessian_expect to TDQubit (mirroring OTDQubit) and added tdqubit to the sesolve Hessian sweep

spital: Is there anything else I should update or improve here?

I’m asking because my open PR slots are currently limited by the hackathon rules, so I would appreciate any guidance when you have a chance.

gautierronan: No it's ok, I just didn't have time to review yet ;)

gautierronan (approved):
Thanks for the changes @spital. Everything looks good to me!

spital:
Thank you, just edited PR description, it shall better reflect what was updated

gautierronan:
I'll let @Nicolas-Lepage handle the merge because I don't know the exact rules/requirements for unitaryHACK

PR #1112, opened 2026-06-08
Add vacuum and cat state helpers

gautierronan:
## Summary

Adds four new state helpers to `dq.utils.states`:

- **`vacuum(dim)` / `vacuum_dm(dim)`** — ket and density matrix of the vacuum state (the zero-photon Fock state).
- **`cat(dim, alpha, theta=0.0)` / `cat_dm(dim, alpha, theta=0.0)`** — ket and density matrix of a Schrödinger cat state, the normalized superposition $\ket{\alpha} + e^{i\theta}\ket{-\alpha}$. The relative phase `theta` defaults to `0` (even cat); `theta=pi` gives the odd cat.

The `cat`/`cat_dm` helpers support **batching over both `alpha` and `theta`**, which are broadcast together following NumPy rules:

```python
dq.cat(8, alpha=[1.0, 2.0, 3.0])              # -> (3, 8, 1)
dq.cat(8, alpha[None, :], theta[:, None])     # -> (n_theta, 3, 8, 1)
```

The new helpers are registered in the API reference and covered by tests (parity with `fock`/`coherent`, even/odd-cat Fock support, normalization, alpha-batch equivalence to a manual `stack`, broadcast shapes, and jit-trace checks).

## Commits

1. `vacuum` / `vacuum_dm` helpers.
2. `cat` / `cat_dm` helpers with batching support.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dynamiqs/utils/states.py
+++ b/dynamiqs/utils/states.py
@@ -554,6 +554,12 @@ def cat(dim: int, alpha: ArrayLike, theta: ArrayLike = 0.0) -> QArray:
         Arguments `alpha` and `theta` are broadcast together following NumPy
         broadcasting rules, allowing batching over either or both.
 
+    Note:
+        In the vanishing-amplitude limit $\alpha\to 0$ the two coherent states
+        collapse onto the vacuum. The returned state is then the vacuum
+        $\ket{0}$ for the even cat, and the single-photon Fock state $\ket{1}$
+        for the odd cat ($\theta=\pi$), where the vacuum contributions cancel.
+
     Returns:
         (qarray of shape (..., dim, 1)): Ket of the cat state.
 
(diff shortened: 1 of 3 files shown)


abocquet (dismissed):

dynamiqs/utils/states.py
```diff
@@ -481,3 +485,152 @@ def _single_thermal_dm(dim: int, nth: Array) -> QArray:
...
+    theta = jnp.broadcast_to(theta, bshape)
+
+    # compute the two coherent components, each of shape (..., dim, 1)
+    plus_alpha = coherent(dim, alpha)
+    minus_alpha = coherent(dim, -alpha)
+
+    # reshape the phase as a batched scalar of shape (..., 1, 1)
+    phase = jnp.exp(1j * theta)[..., None, None]
+
+    return unit(plus_alpha + minus_alpha * phase)
```
Nicolas-Lepage: I think there is one edge case that will trigger a problem:
if we do `cat(dim, alpha=0, theta=jnp.pi)` then we have` (∣0⟩−∣0⟩)/norm(∣0⟩−∣0⟩)` which will divide by 0 and return NaNs.
We can test this with:
```
result = dq.cat(dim, 0.0, jnp.pi)
assert jnp.any(jnp.isnan(result.to_jax()))
```
If this fails, we should raise an error when `alpha == 0`.

Nicolas-Lepage (changes requested):
Thanks for the helper Additions ! everything seems great except for one question i have left :)

dynamiqs/utils/states.py
gautierronan: Indeed, thanks for raising. Made a fix in https://github.com/dynamiqs/dynamiqs/pull/1112/changes/d2438e52d777d7ed0e76e38b8eb713c822345db5. It makes sure that `dq.cat(n, 0, pi)` does return the proper state, although to be thorough, it also introduces a discontinuity in the limit of alpha -> 0 and theta -> pi. However, that should be niche enough. I was not sure how to handle this better.

Nicolas-Lepage: Thanks for the modification, this seems good to me !

Nicolas-Lepage (approved):

PR #1115, opened 2026-06-10
Add QArray manipulation utilities

### LINKED ISSUE #1080: unitaryHACK - Expand QArray manipulation utilities
### Context

The current `QArray` API lacks standard array manipulation utilities commonly available in NumPy/JAX, such as `swapaxes`, `moveaxis`, `expand_dims`, and `where`.

As a result:

- users often need to convert `QArray` objects to `jax.Array` for basic operations;
- this introduces unnecessary overhead and breaks abstraction consistency;
- the API is less ergonomic and harder to adopt.

This issue aims to close that gap by extending `QArray` with commonly used manipulation primitives.

### Goal

Add support for the following `QArray` manipulation utilities.

**Core additions**

- `swapaxes`
- `moveaxis`
- `expand_dims`
- `where`
- `concatenate`

**Optional additions**
- `take`
- `unstack`

The core additions are required for completion. Optional additions are welcome but not required.

### Expected behavior

The new utilities should follow NumPy/JAX semantics as closely as possible.

They should:

- work directly on `QArray` inputs;
- preserve the `QArray` abstraction in their outputs when appropriate;
- support batching dimensions consistently with the existing `QArray` API;
- avoid requiring users to manually convert to raw `jax.Array`.

### Tests

Add tests comparing the new behavior against the equivalent NumPy/JAX operations.

Tests should cover:

- output shape;
- output values;
- dtype preservation where relevant;
- behavior on batched `QArray` objects;
- compatibility with the existing `QArray` conventions.

### Acceptance criteria

- All core additions are implemented.
- Each core addition has tests.
- Behavior matches NumPy/JAX semantics where applicable.
- Outputs preserve `QArray` behavior where appropriate.
- Existing QArray tests continue to pass.
- Optional additions may be included but are not required.

### Focus

Mostly engineering. Focus on API design, usability, consistency with NumPy/JAX semantics, and integration with the existing `QArray` API.

### Difficulty

Easy

jackson5689:
Fixes #1080.

## Summary

- add `swapaxes`, `moveaxis`, `expand_dims`, `where`, and `concatenate` qarray utilities
- expose matching `QArray` convenience methods
- preserve qarray metadata and sparse DIA layout for supported batch-axis operations
- protect the final two quantum-object dimensions from accidental batch-axis manipulation
- add dense, sparse, scalar, shape, and error-path tests against equivalent JAX behavior

## Tests

- `.venv/bin/ruff check dynamiqs/qarrays/qarray.py dynamiqs/qarrays/utils.py tests/qarrays/test_qarray_manipulation.py`
- `.venv/bin/python -m pytest tests/qarrays/test_qarray_manipulation.py`
- `.venv/bin/python -m pytest tests/qarrays`

The full qarray test run passed with 82 tests and 3 existing sparse scalar-add warnings.

## AI assistance disclosure

I used AI assistance to inspect the existing qarray structure, draft parts of this patch, and prepare tests. I reviewed the implementation, adjusted it to maintainer feedback, and ran the validation commands listed above.

### DIFF
--- a/dynamiqs/qarrays/dataarray.py
+++ b/dynamiqs/qarrays/dataarray.py
@@ -126,15 +126,13 @@ class DataArray(eqx.Module):
         self, source: int | Sequence[int], destination: int | Sequence[int]
     ) -> DataArray:
         """Moves axes to new positions."""
-        source_axes = tuple(
-            _normalize_axis(axis, self.ndim)
-            for axis in ((source,) if isinstance(source, int) else tuple(source))
+        source_axes = (source,) if isinstance(source, int) else tuple(source)
+        source_axes = tuple(_normalize_axis(axis, self.ndim) for axis in source_axes)
+        destination_axes = (
+            (destination,) if isinstance(destination, int) else tuple(destination)
         )
         destination_axes = tuple(
-            _normalize_axis(axis, self.ndim)
-            for axis in (
-                (destination,) if isinstance(destination, int) else tuple(destination)
-            )
+            _normalize_axis(axis, self.ndim) for axis in destination_axes
         )
         _check_axis_tuples(source_axes, destination_axes)
 
@@ -145,9 +143,18 @@ class DataArray(eqx.Module):
             return self
         if axis_order == final_swap_order:
             return self.mT
-        if in_last_two_dims((*source_axes, *destination_axes), self.ndim):
+        if set(axis_order[:-2]) != set(range(self.ndim - 2)):
             _raise_matrix_axis_error('moveaxis')
-        return self._moveaxis_unchecked(source_axes, destination_axes)
+
+        batch_source = tuple(range(self.ndim - 2))
+        batch_destination = tuple(axis_order[:-2].index(axis) for axis in batch_source)
+        if axis_order[-2:] == identity_order[-2:]:
+            return self._moveaxis_unchecked(batch_source, batch_destination)
+        elif axis_order[-2:] == final_swap_order[-2:]:
+            return self.mT._moveaxis_unchecked(batch_source, batch_destination)
+        else:
+            _raise_matrix_axis_error('moveaxis')
+        raise AssertionError('unreachable')
 
     @abstractmethod
     def _moveaxis_unchecked(
@@ -333,9 +340,9 @@ def _normalize_axis(axis: int, ndim: int) -> int:
 def _normalize_insert_axes(axis: int | Sequence[int], ndim: int) -> tuple[int, ...]:
     axes = (axis,) if isinstance(axis, int) else tuple(axis)
     out_ndim = ndim + len(axes)
-    normalized_axes = tuple(_normalize_axis(a, out_ndim) for a in axes)
+    normalized_axes = tuple(_normalize_axis(ax, out_ndim) for ax in axes)
     if len(set(normalized_axes)) != len(normalized_axes):
-        raise ValueError('repeated axis')
+        raise ValueError(f'Repeated axis found in axes to insert. Got `axis={axis}`.')
     return normalized_axes
 
 
@@ -345,9 +352,11 @@ def _check_axis_tuples(source: tuple[int, ...], destination: tuple[int, ...]) ->
             '`source` and `destination` arguments must have the same number of axes.'
         )
     if len(set(source)) != len(source):
-        raise ValueError('repeated axis in `source` argument')
+        raise ValueError(f'Repeated axis found in `source` argument. Got {source}.')
     if len(set(destination)) != len(destination):
-        raise ValueError('repeated axis in `destination` argument')
+        raise ValueError(
+            f'Repeated axis found in `destination` argument. Got {destination}.'
+        )
 
 
 def _moveaxis_order(
--- a/dynamiqs/qarrays/dataarray.py
+++ b/dynamiqs/qarrays/dataarray.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from collections.abc import Sequence
 from math import prod
 from types import EllipsisType
 from typing import TYPE_CHECKING, ClassVar, TypeAlias
@@ -100,19 +101,65 @@ class DataArray(eqx.Module):
     def broadcast_to(self, *shape: int) -> DataArray:
         """Broadcasts to a new shape."""
 
-    @abstractmethod
     def swapaxes(self, axis1: int, axis2: int) -> DataArray:
         """Interchanges two axes."""
+        axis1 = _normalize_axis(axis1, self.ndim)
+        axis2 = _normalize_axis(axis2, self.ndim)
+        if axis1 == axis2:
+            return self
+        if {axis1, axis2} == {self.ndim - 2, self.ndim - 1}:
+            return self.mT
+        if in_last_two_dims((axis1, axis2), self.ndim):
+            _raise_matrix_axis_error('swapaxes')
+        return self._swapaxes_unchecked(axis1, axis2)
 
     @abstractmethod
+    def _swapaxes_unchecked(self, axis1: int, axis2: int) -> DataArray:
+        """Does the heavy-lifting for `swapaxes` but skips all checks."""
+
     def moveaxis(
-        self, source: int | tuple[int, ...], destination: int | tuple[int, ...]
+        self, source: int | Sequence[int], destination: int | Sequence[int]
     ) -> DataArray:
         """Moves axes to new positions."""
+        source_axes = tuple(
+            _normalize_axis(axis, self.ndim)
+            for axis in ((source,) if isinstance(source, int) else tuple(source))
+        )
+        destination_axes = tuple(
+            _normalize_axis(axis, self.ndim)
+            for axis in (
+                (destination,) if isinstance(destination, int) else tuple(destination)
+            )
+        )
+        _check_axis_tuples(source_axes, destination_axes)
+
+        axis_order = _moveaxis_order(source_axes, destination_axes, self.ndim)
+        identity_order = tuple(range(self.ndim))
+        final_swap_order = (*range(self.ndim - 2), self.ndim - 1, self.ndim - 2)
+        if axis_order == identity_order:
+            return self
+        if axis_order == final_swap_order:
+            return self.mT
+        if in_last_two_dims((*source_axes, *destination_axes), self.ndim):
+            _raise_matrix_axis_error('moveaxis')
+        return self._moveaxis_unchecked(source_axes, destination_axes)
 
     @abstractmethod
-    def expand_dims(self, axis: int | tuple[int, ...]) -> DataArray:
+    def _moveaxis_unchecked(
+        self, source: tuple[int, ...], destination: tuple[int, ...]
+    ) -> DataArray:
+        """Does the heavy-lifting for `moveaxis` but skips all checks."""
+
+    def expand_dims(self, axis: int | Sequence[int]) -> DataArray:
         """Expands the shape by inserting new axes."""
+        axes = _normalize_insert_axes(axis, self.ndim)
+        if in_last_two_dims(axes, self.ndim + len(axes)):
+            _raise_matrix_axis_error('expand_dims')
+        return self._expand_dims_unchecked(axes)
+
+    @abstractmethod
+    def _expand_dims_unchecked(self, axis: tuple[int, ...]) -> DataArray:
+        """Does the heavy-lifting for `expand_dims` but skips all checks."""
 
     @abstractmethod
     def powm(self, n: int) -> DataArray:
@@ -267,6 +314,49 @@ def include_last_two_dims(axis: int | tuple[int, ...] | None, ndim: int) -> bool
     axis = (axis,) if isinstance(axis, int) else axis
     return axis is None or (
         ndim - 1 in [a % ndim for a in axis] and ndim - 2 in [a % ndim for a in axis]
+    )
+
+
+def _normalize_axis(axis: int, ndim: int) -> int:
+    if not -ndim <= axis < ndim:
+        raise ValueError(f'axis {axis} is out of bounds for array of dimension {ndim}')
+    return axis % ndim
+
+
+def _normalize_insert_axes(axis: int | Sequence[int], ndim: int) -> tuple[int, ...]:
+    axes = (axis,) if isinstance(axis, int) else tuple(axis)
+    out_ndim = ndim + len(axes)
+    normalized_axes = tuple(_normalize_axis(a, out_ndim) for a in axes)
+    if len(set(normalized_axes)) != len(normalized_axes):
+        raise ValueError('repeated axis')
+    return normalized_axes
+
+
+def _check_axis_tuples(source: tuple[int, ...], destination: tuple[int, ...]) -> None:
+    if len(source) != len(destination):
+        raise ValueError(
+            '`source` and `destination` arguments must have the same number of axes.'
+        )
+    if len(set(source)) != len(source):
+        raise ValueError('repeated axis in `source` argument')
+    if len(set(destination)) != len(destination):
+        raise ValueError('repeated axis in `destination` argument')
+
+
+def _moveaxis_order(
+    source: tuple[int, ...], destination: tuple[int, ...], ndim: int
+) -> tuple[int, ...]:
+    order = [axis for axis in range(ndim) if axis not in source]
+    for destination_axis, source_axis in sorted(zip(destination, source, strict=True)):
+        order.insert(destination_axis, source_axis)
+    return tuple(order)
+
+
+def _raise_matrix_axis_error(operation: str) -> None:
+    raise ValueError(
+        f'`{operation}` can only manipulate batching dimensions of a data array; the '
+        'final two dimensions are matrix dimensions. Use `swapaxes(-1, -2)` or '
+        '`.mT` to transpose those final dimensions.'
     )
 
 
--- a/dynamiqs/qarrays/dataarray.py
+++ b/dynamiqs/qarrays/dataarray.py
@@ -101,6 +101,20 @@ class DataArray(eqx.Module):
         """Broadcasts to a new shape."""
 
     @abstractmethod
+    def swapaxes(self, axis1: int, axis2: int) -> DataArray:
+        """Interchanges two axes."""
+
+    @abstractmethod
+    def moveaxis(
+        self, source: int | tuple[int, ...], destination: int | tuple[int, ...]
+    ) -> DataArray:
+        """Moves axes to new positions."""
+
+    @abstractmethod
+    def expand_dims(self, axis: int | tuple[int, ...]) -> DataArray:
+        """Expands the shape by inserting new axes."""
+
+    @abstractmethod
     def powm(self, n: int) -> DataArray:
         pass
 
--- a/dynamiqs/qarrays/dense_dataarray.py
+++ b/dynamiqs/qarrays/dense_dataarray.py
@@ -58,17 +58,17 @@ class DenseDataArray(DataArray):
 
     def _swapaxes_unchecked(self, axis1: int, axis2: int) -> DataArray:
         data = jnp.swapaxes(self.data, axis1, axis2)
-        return replace(self, data=data)  # ty: ignore[invalid-argument-type]
+        return replace(self, data=data)
 
     def _moveaxis_unchecked(
         self, source: tuple[int, ...], destination: tuple[int, ...]
     ) -> DataArray:
         data = jnp.moveaxis(self.data, source, destination)
-        return replace(self, data=data)  # ty: ignore[invalid-argument-type]
+        return replace(self, data=data)
 
     def _expand_dims_unchecked(self, axis: tuple[int, ...]) -> DataArray:
         data = jnp.expand_dims(self.data, axis=axis)
-        return replace(self, data=data)  # ty: ignore[invalid-argument-type]
+        return replace(self, data=data)
 
     def powm(self, n: int) -> DataArray:
         data = jnp.linalg.matrix_power(self.data, n)
--- a/dynamiqs/qarrays/dense_dataarray.py
+++ b/dynamiqs/qarrays/dense_dataarray.py
@@ -56,17 +56,17 @@ class DenseDataArray(DataArray):
         data = jnp.broadcast_to(self.data, shape)
         return replace(self, data=data)  # ty: ignore[invalid-argument-type]
 
-    def swapaxes(self, axis1: int, axis2: int) -> DataArray:
+    def _swapaxes_unchecked(self, axis1: int, axis2: int) -> DataArray:
         data = jnp.swapaxes(self.data, axis1, axis2)
         return replace(self, data=data)  # ty: ignore[invalid-argument-type]
 
-    def moveaxis(
-        self, source: int | tuple[int, ...], destination: int | tuple[int, ...]
+    def _moveaxis_unchecked(
+        self, source: tuple[int, ...], destination: tuple[int, ...]
     ) -> DataArray:
         data = jnp.moveaxis(self.data, source, destination)
         return replace(self, data=data)  # ty: ignore[invalid-argument-type]
 
-    def expand_dims(self, axis: int | tuple[int, ...]) -> DataArray:
+    def _expand_dims_unchecked(self, axis: tuple[int, ...]) -> DataArray:
         data = jnp.expand_dims(self.data, axis=axis)
         return replace(self, data=data)  # ty: ignore[invalid-argument-type]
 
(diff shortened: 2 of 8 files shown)


dynamiqs/qarrays/qarray.py
```diff
@@ -296,6 +301,33 @@ def broadcast_to(self, *shape: int) -> QArray:
...
+        from .utils import expand_dims  # noqa: PLC0415
+
+        return expand_dims(self, axis)
+
+    def where(self, condition: ArrayLike, y: QArrayLike) -> QArray:
+        from .utils import where  # noqa: PLC0415
+
+        return where(condition, self, y)
+
+    def concatenate(self, qarrays: Sequence[QArrayLike], axis: int = 0) -> QArray:
```
Nicolas-Lepage: Concatenate is a N-ary function so it should not be a method but purely in utils ( can call helpers in sparse data arrays )

dynamiqs/qarrays/utils.py
```diff
@@ -204,6 +209,191 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+    if isinstance(x.data, DenseDataArray):
+        data = DenseDataArray(jnp.swapaxes(x.data.data, axis1, axis2))
+    elif _axes_are_batch_axes((axis1, axis2), x.ndim):
+        data = SparseDIADataArray(
+            x.data.offsets, jnp.swapaxes(x.data.diags, axis1, axis2)
+        )
+    else:
+        data = x.data.mT
+
+    return MaterializedQArray(x.dims, x.vectorized, data)
```
Nicolas-Lepage: this function should be a Qarray method

dynamiqs/qarrays/utils.py
```diff
@@ -204,6 +209,191 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+
+    if isinstance(x.data, DenseDataArray):
+        data = DenseDataArray(jnp.moveaxis(x.data.data, source_axes, destination_axes))
+    else:
+        sparse_data = (
+            x.data.mT if axis_order[-2:] == (x.ndim - 1, x.ndim - 2) else x.data
+        )
+        batch_order = axis_order[:-2]
+        diags = jnp.transpose(sparse_data.diags, (*batch_order, x.ndim - 2, x.ndim - 1))
+        data = SparseDIADataArray(sparse_data.offsets, diags)
```
Nicolas-Lepage: this function should be a Qarray method

dynamiqs/qarrays/utils.py
```diff
@@ -204,6 +209,191 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+        _raise_quantum_axis_error('expand_dims')
+
+    if isinstance(x.data, DenseDataArray):
+        data = DenseDataArray(jnp.expand_dims(x.data.data, axis=axes))
+    else:
+        data = SparseDIADataArray(
+            x.data.offsets, jnp.expand_dims(x.data.diags, axis=axes)
+        )
+
+    return MaterializedQArray(x.dims, x.vectorized, data)
```
Nicolas-Lepage: this function should be a Qarray method

dynamiqs/qarrays/utils.py
```diff
@@ -204,6 +209,191 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+    else:
+        _check_compatible_operand_dims(y, x)
+        dims = y.dims
+        vectorized = y.vectorized
+
+    if _contains_sparse_qarray(x, y):
+        _warn_sparse_to_dense('where')
+
+    data = jnp.where(condition, to_jax(x), to_jax(y))
+    return MaterializedQArray(dims, vectorized, DenseDataArray(data))
```
Nicolas-Lepage, line 322: The where implementation is a bit off. Mainly the if/elif/else is more complex than the problem needs, and there's a latent crash.

The four-way branch over `(x_is_qarray, y_is_qarray)` collapses to a single decision: take `dims`/`vectorized` from whichever operand is a qarray. It doesn't need four cases or the two `_check_compatible_*` helpers. 

this also creates a bug : the neither-qarray branch runs : `asqarray` on its operands, and `asqarray(1.0)` is 0-d, so it dies on `shape[-2]`. `dq.where(cond, 1.0, 2.0)` currently raises an opaque `IndexError`.

what i suggest : 

- If neither `x` nor `y` is a qarray, raise and point the user to `jnp.where `(no qarray metadata to preserve).
- Otherwise take metadata from the qarray operand and check the other is compatible via `check_compatible_dims` — but don't coerce it to a qarray. Scalar fills like dq.where(cond, rho, 0.0) are normal; the non-qarray operand should go through to_jax and let jnp.where broadcast it.

Keep the dense output and the densify warning — those are correct.

Nicolas-Lepage (changes requested):
Thank you for the PR! The functions have the intended end effect. But they diverge a fair bit from the library's structure. I'd advise following the existing structure much more closely. It should make the endpoint functions a lot simpler.

Looking at the utilities:
`swapaxes` / `moveaxis` / `expand_dims` are unary. One qarray in and one qarray out. So their place is a `QArray` method. To implement them follow the structure and the chain of method or function calls used by `reshape` or `broadcast_to`.

One point of caution for that restructure. The `axis1 axis2 == -1 -2` case shouldn't be an inline `x.data.mT`. The pure `{ndim-2 ndim-1} `swap is the matrix transpose. The hierarchy already has `mT` for it and sparse goes through `transpose_sparsedia`. So the base guard should be "batch axes only or the pure last two swap routes to self.mT. Anything mixing batch and quantum axes raises." This reuses `mT` rather than reaching into the data.

`concatenate` is N-ary. Multiple qarrays in and one out. So it should live only in utils following the `stack` structure. It can call helpers on the sparse data arrays.

`where` is its own case. It selects between two operands and always returns dense. So it stays a free function in utils and doesn't go through the hierarchy. See the inline comment for details.

h2110w0r1d:
Thanks for working on this PR! Since this is part of the hackathon and multiple PRs for the same issue are allowed, I’m also thinking of trying an alternative implementation for #1080 that follows the requested QArray method structure more closely.

My understanding from the review is that the unary operations (`swapaxes`, `moveaxis`, `expand_dims`) should be implemented through the existing QArray method hierarchy, similar to `reshape` / `broadcast_to`, while `concatenate` should remain a utility function like `stack`, and `where` should stay as a free function with at least one QArray operand.

I’ll try to keep my version focused on that structure. Thanks again — the discussion here is very helpful for understanding the intended design.

jackson5689:
Thanks for the review. I pushed an update addressing the requested API structure changes:

- moved `swapaxes`, `moveaxis`, and `expand_dims` into the `QArray` hierarchy, with `MaterializedQArray` delegating to `DataArray` implementations
- route pure final-axis transposes through `.mT`
- kept `concatenate` as a utility function only, following `stack`
- simplified `where` so at least one operand must be a `QArray`, scalar operands are left to `jnp.where`, and qarray metadata compatibility is checked where applicable

Validation after rebasing on current `main`:

- `.venv/bin/ruff check dynamiqs/qarrays/qarray.py dynamiqs/qarrays/materialized_qarray.py dynamiqs/qarrays/dataarray.py dynamiqs/qarrays/dense_dataarray.py dynamiqs/qarrays/sparsedia_dataarray.py dynamiqs/qarrays/utils.py tests/qarrays/test_qarray_manipulation.py`
- `.venv/bin/python -m pytest tests/qarrays/test_qarray_manipulation.py -q` -> `18 passed`
- `.venv/bin/python -m pytest tests/qarrays -q` -> `83 passed, 3 existing sparse scalar-add warnings`

Nicolas-Lepage (changes requested):
Thanks for the modification. this is much closer to our usual structure. I still have some more comments : 

1. Move the guard from `MaterializedQArray` into `DataArray`, mirroring `reshape`.  then call an abstract `_…_unchecked` that the dense/sparse children implement with the raw `jnp` op. `MaterializedQArray` becomes pure delegation (`replace(self, data=self.data.swapaxes(...))`). Reword the guard's error to be about the matrix (last-two) dims, not "quantum object," to fit `DataArray`'s neutral scope. 

 2. Remove three helpers:- `_axes_are_batch_axes` -> use the existing `in_last_two_dims` (negated).- `_normalize_axes` -> inline at its one `moveaxis` site.- `_check_qarraylike_dims` -> inline its single guarded `check_compatible_dims`.

3. Move to `dataarray.py` (next to `in_last_two_dims` / `include_last_two_dims`, since the `DataArray` guard now consumes them): `_normalize_axis`, `_normalize_insert_axes`, `_check_axis_tuples`, `_moveaxis_order`, `_raise_quantum_axis_error`.

4. `concatenate`: do its single-axis batch check inline (stack-style), so it doesn't
import the axis helpers. Keep the guard.

5. Leave in `utils.py` the rest of the helpers useful for `where`.

6. tests : 
- Rework the error-message matches to fit the new check placement. Once the
  guard moves to `DataArray` with a matrix-centric message, the current
  'final two dimensions represent the quantum object' match breaks -- match a
  stable substring instead.
- Add a multi-axis `moveaxis` case. Every current `moveaxis` test passes single
  ints, so `_moveaxis_order` and `_check_axis_tuples` are unexercised on their
  real path -- and that's the logic being relocated. e.g. moveaxis((0,1),(2,0))
  against jnp.

Nicolas-Lepage (changes requested):
Thanks, this revision is in good shape and the earlier structural points are all addressed. A few remaining comments, mostly about sparse `concatenate` following the `stack` precedent.

1. Sparse `concatenate` should delegate to a primitive instead of inlining the diag handling. Right now it reaches into `q.data.offsets` and `q.data.diags` and calls `jnp.concatenate` directly, which breaks from the `stack` structure and silently degrades to dense whenever inputs don't share identical offsets. Please add a `concatenate_sparsedia` primitive in `sparsedia_primitives.py` mirroring `stack_sparsedia`: union the offsets, scatter each input's diags into the merged layout, then `jnp.concatenate` along the batch axis. Then have `concatenate` call it. That keeps the sparse path sparse and drops the all-offsets-identical restriction.

2. Once the primitive exists, select the sparse branch purely on all inputs being `SparseDIADataArray` and drop the identical-offsets condition. Offset reconciliation is the primitive's job, same as `stack`.

3. Make `concatenate` input handling follow `stack` rather than `asqarray`-coercing. Align the signature to `Sequence[QArray]` with the same strict `isinstance(q, QArray)` check, and add a batch-`ndim` consistency check so a mismatch raises a friendly error. Keep the relaxed `shape[-2:]` comparison though. That one's correct, since the concat axis sizes are expected to differ.

The rest looks good to me.

jackson5689:
Thanks for the follow-up review. I pushed `b232fbe` addressing the remaining sparse `concatenate` points:

- added `concatenate_sparsedia`, mirroring the `stack_sparsedia` offset-union/scatter structure
- sparse `concatenate` now stays in DIA layout for all-sparse inputs, including different offset sets
- `concatenate` now follows `stack` input handling with strict `QArray` inputs and a batch-ndim consistency check
- added regression coverage for different sparse offsets, non-`QArray` inputs, and mismatched batch ndim

Validation:

```bash
.venv/bin/ruff check dynamiqs/qarrays/sparsedia_primitives.py dynamiqs/qarrays/utils.py tests/qarrays/test_qarray_manipulation.py
.venv/bin/ruff format --check dynamiqs/qarrays/sparsedia_primitives.py dynamiqs/qarrays/utils.py tests/qarrays/test_qarray_manipulation.py
.venv/bin/python -m pytest tests/qarrays/test_qarray_manipulation.py -q
.venv/bin/python -m pytest tests/qarrays -q
```

Results: `22 passed` for `test_qarray_manipulation.py`; `87 passed, 3 warnings` for `tests/qarrays` (the warnings are the existing sparse scalar-add dense-conversion warnings).

jackson5689:
Synced the branch with current `origin/main` to clear the dirty merge state. Current head is `e5214c7`.

I reran the relevant validation after resolving the sparse primitive conflict:

```bash
.venv/bin/ruff check dynamiqs/qarrays/sparsedia_primitives.py dynamiqs/qarrays/utils.py tests/qarrays/test_qarray_manipulation.py
.venv/bin/ruff format --check dynamiqs/qarrays/sparsedia_primitives.py dynamiqs/qarrays/utils.py tests/qarrays/test_qarray_manipulation.py
.venv/bin/python -m pytest tests/qarrays/test_qarray_manipulation.py -q
.venv/bin/python -m pytest tests/qarrays -q
```

Results: `22 passed` for `test_qarray_manipulation.py`; `87 passed, 3 warnings` for `tests/qarrays` (same existing sparse scalar-add dense-conversion warnings).

Nicolas-Lepage (changes requested):
Hi @jackson5689 thanks for replying! The implementation looks good to me! Thanks for sticking through what we thought was a small issue and turned out to be quite large. 

@gautierronan Do you have time to take a look, especially to see if the structure of the code matches our library ?

I will just ask you for one final modification concerning tests. I have taken a final look and think that some behaviors should be tested:

**Vectorized preservation**

* Add a parametrized test running `swapaxes`, `moveaxis`, and `expand_dims` on a `vectorized=True` qarray and asserting the result is still vectorized (plus shape/value against JAX).
* Add a `concatenate` case on `vectorized=True` inputs asserting the result stays vectorized.
* Add a `where` case with two `vectorized=True` operands asserting the result is vectorized.
* Add the error cases: `where` with mismatched `vectorized` operands should raise (hits the branch in `_check_compatible_qarray_metadata`), and `concatenate` with mismatched `vectorized` should raise (the branch in `_check_concatenate_input`). Match a stable substring.

**Sparse paths**

* `swapaxes(-1, -2)` under `dq.dia`. `moveaxis(-1, -2)` sparse is tested but its `swapaxes` twin isn't, even though both hit the `self.mT` → `transpose_sparsedia` routing — that's the asymmetry I'd most like closed.
* Multi-axis `moveaxis` under sparse, e.g. `moveaxis((0,1),(2,0))` on a `dq.dia` qarray against JAX. Currently the multi-axis case is dense-only.
* Tuple-axis `expand_dims` under sparse, e.g. `axis=(0,2)`. The sparse test only does a single positive axis.

jackson5689:
Thanks for the final test review. I pushed `ca37a99` adding the requested coverage:

- vectorized preservation for `swapaxes`, `moveaxis`, `expand_dims`, `concatenate`, and `where`
- mismatched `vectorized` error cases for `where` and `concatenate`
- sparse `swapaxes(-1, -2)`, sparse multi-axis `moveaxis`, and sparse tuple-axis `expand_dims`

Validation:

```bash
.venv/bin/ruff check tests/qarrays/test_qarray_manipulation.py
.venv/bin/ruff format --check tests/qarrays/test_qarray_manipulation.py
.venv/bin/python -m pytest tests/qarrays/test_qarray_manipulation.py -q
.venv/bin/python -m pytest tests/qarrays -q
```

Results: `32 passed` for `test_qarray_manipulation.py`; `97 passed, 3 warnings` for `tests/qarrays` (same existing sparse scalar-add dense-conversion warnings).

jackson5689:
Hi @Nicolas-Lepage and @gautierronan, just following up on the latest revision when you have a chance.

I pushed `ca37a99` to cover the remaining test cases from the last review: vectorized preservation/error paths plus the additional sparse `swapaxes`, multi-axis `moveaxis`, and tuple-axis `expand_dims` cases. The focused qarray test file and the full `tests/qarrays` suite passed locally.

Please let me know if there is anything else you would like adjusted before this can move forward. Thanks again for the detailed reviews.

dynamiqs/qarrays/dataarray.py
```diff
@@ -105,6 +106,66 @@ def _reshape_unchecked(self, *shape: int) -> DataArray:
...
+        axis_order = _moveaxis_order(source_axes, destination_axes, self.ndim)
+        identity_order = tuple(range(self.ndim))
+        final_swap_order = (*range(self.ndim - 2), self.ndim - 1, self.ndim - 2)
+        if axis_order == identity_order:
+            return self
+        if axis_order == final_swap_order:
+            return self.mT
+        if in_last_two_dims((*source_axes, *destination_axes), self.ndim):
+            _raise_matrix_axis_error('moveaxis')
+        return self._moveaxis_unchecked(source_axes, destination_axes)
```
gautierronan: I'm nit-picking here, but in principle we could also handle the edge case where we want to swap the last two axes and move other axes around. For instance,
```python
qarray = dq.eye(4).broadcast_to(1, 1, 4, 4)
qarray.moveaxis((0, -1, -2), (1, -2, -1))
```
will raise an error. If it's too complicated to fix, I'm ok with skipping this edge case.

dynamiqs/qarrays/dataarray.py
```diff
@@ -105,6 +106,66 @@ def _reshape_unchecked(self, *shape: int) -> DataArray:
...
+        """Does the heavy-lifting for `swapaxes` but skips all checks."""
+
+    def moveaxis(
+        self, source: int | Sequence[int], destination: int | Sequence[int]
+    ) -> DataArray:
+        """Moves axes to new positions."""
+        source_axes = tuple(
+            _normalize_axis(axis, self.ndim)
+            for axis in ((source,) if isinstance(source, int) else tuple(source))
+        )
```
gautierronan:
```suggestion
        source_axes = (source,) if isinstance(source, int) else tuple(source)
        source_axes = tuple(_normalize_axis(axis, self.ndim) for axis in source_axes)
```
may be more readable, if the type-checker allows it (same for `destimation_axes`)

dynamiqs/qarrays/dataarray.py
```diff
@@ -263,5 +324,48 @@ def include_last_two_dims(axis: int | tuple[int, ...] | None, ndim: int) -> bool
...
+def _normalize_axis(axis: int, ndim: int) -> int:
+    if not -ndim <= axis < ndim:
+        raise ValueError(f'axis {axis} is out of bounds for array of dimension {ndim}')
+    return axis % ndim
+
+
+def _normalize_insert_axes(axis: int | Sequence[int], ndim: int) -> tuple[int, ...]:
+    axes = (axis,) if isinstance(axis, int) else tuple(axis)
+    out_ndim = ndim + len(axes)
+    normalized_axes = tuple(_normalize_axis(a, out_ndim) for a in axes)
```
gautierronan:
```suggestion
    normalized_axes = tuple(_normalize_axis(ax, out_ndim) for ax in axes)
```

dynamiqs/qarrays/dataarray.py
```diff
@@ -263,5 +324,48 @@ def include_last_two_dims(axis: int | tuple[int, ...] | None, ndim: int) -> bool
...
+    if len(set(normalized_axes)) != len(normalized_axes):
+        raise ValueError('repeated axis')
```
gautierronan: Can you improve the error message here and just below, to help users debug? For instance,
```python
        raise ValueError(f'Repeated axis found in axes to insert. Got `axis` = {axis}.')
```

dynamiqs/qarrays/dense_dataarray.py
```diff
@@ -56,6 +56,20 @@ def broadcast_to(self, *shape: int) -> DataArray:
         data = jnp.broadcast_to(self.data, shape)
         return replace(self, data=data)
 
+    def _swapaxes_unchecked(self, axis1: int, axis2: int) -> DataArray:
+        data = jnp.swapaxes(self.data, axis1, axis2)
+        return replace(self, data=data)  # ty: ignore[invalid-argument-type]
```
gautierronan: Why do you need the typing ignore here? Can we remove it?

dynamiqs/qarrays/qarray.py
```diff
@@ -302,6 +306,48 @@ def broadcast_to(self, *shape: int) -> QArray:
...
+        """Expand the shape of a qarray by inserting new axes.
+
+        Args:
+            axis: Axis or axes where new dimensions are inserted.
+
+        Returns:
+            Qarray with additional dimensions.
+        """
+
+    def where(self, condition: ArrayLike, y: QArrayLike) -> QArray:
```
gautierronan, line 346: Doesn't this method require a docstring for the public documentation?

dynamiqs/qarrays/utils.py
```diff
@@ -216,6 +230,166 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+        vectorized = x.vectorized
+    else:
+        x_dims = get_dims(x)
+        if x_dims is not None:
+            check_compatible_dims(y.dims, x_dims)
+        dims = y.dims
+        vectorized = y.vectorized
+
+    if _contains_sparse_qarray(x, y):
+        _warn_sparse_to_dense('where')
```
gautierronan, line 319: Can you add a `TODO: Generalize implementation of `dq.where` to sparse dia layout.` ?

dynamiqs/qarrays/utils.py
```diff
@@ -216,6 +230,166 @@ def stack(qarrays: Sequence[QArray], axis: int = 0) -> QArray:
...
+
+    data = jnp.where(condition, to_jax(x), to_jax(y))
+    return MaterializedQArray(dims, vectorized, DenseDataArray(data))
+
+
+def _check_concatenate_input(
```
gautierronan, line 355: Can you move this function below `concatenate` since it's a utility?

gautierronan (changes requested):
Thanks a lot for the great PR @jackson5689, and sorry for the delay in reviewing.

Overall it's in a great shape. I have a few nit-picking comments below, but nothing major. Could you fix them and we can then merge? It also seems like the CI is not passing due to typing, if you don't mind fixing this. Thanks again

jackson5689:
Thanks for the review, @gautierronan. I pushed `cad6376` addressing the nit comments and the typing failure:

- cleaned up `moveaxis` axis normalization and added support for moving batch axes while swapping the final matrix axes
- improved repeated-axis error messages
- removed the unused `ty: ignore` comments
- added the `QArray.where` docstring
- added the sparse DIA `where` TODO
- moved `_check_concatenate_input` below `concatenate`
- tightened the materialized-qarray type narrowing in `where`/`concatenate` and replaced the `np.union1d` reduce path to satisfy `ty`

Validation:

```bash
.venv/bin/ruff check dynamiqs/qarrays/dataarray.py dynamiqs/qarrays/dense_dataarray.py dynamiqs/qarrays/sparsedia_dataarray.py dynamiqs/qarrays/qarray.py dynamiqs/qarrays/utils.py dynamiqs/qarrays/sparsedia_primitives.py tests/qarrays/test_qarray_manipulation.py
.venv/bin/ruff format --check dynamiqs/qarrays/dataarray.py dynamiqs/qarrays/dense_dataarray.py dynamiqs/qarrays/sparsedia_dataarray.py dynamiqs/qarrays/qarray.py dynamiqs/qarrays/utils.py dynamiqs/qarrays/sparsedia_primitives.py tests/qarrays/test_qarray_manipulation.py
.venv/bin/ty check dynamiqs
.venv/bin/python -m pytest tests/qarrays/test_qarray_manipulation.py -q
.venv/bin/python -m pytest tests/qarrays -q
```

Results: `ty` passed, `test_qarray_manipulation.py` passed with `33 passed`, and `tests/qarrays` passed with `98 passed, 3 warnings` (same existing sparse scalar-add dense-conversion warnings).

jackson5689:
I also synced the PR branch with the latest `main` in `6cd23fc` and reran the focused checks locally:

- `ruff check` on the touched qarray files
- `ruff format --check` on the touched qarray files
- `ty check dynamiqs`
- `pytest tests/qarrays/test_qarray_manipulation.py -q` (`33 passed`)
- `pytest tests/qarrays -q` (`98 passed`, with the existing sparse dense-conversion warnings)

The branch should now be up to date for re-review.

gautierronan (approved):
LGTM! Thanks a lot @jackson5689

jackson5689:
I synced the branch with the latest `main` in `c996061` to clear the stale/behind state. The merge only brought in upstream `c06d46b` (`CompositeQArray` skeleton), with no conflicts.

I reran the focused checks locally:

- `ruff check` on the touched qarray files plus `composite_qarray.py`
- `ruff format --check` on the same files
- `ty check dynamiqs`
- `pytest tests/qarrays/test_qarray_manipulation.py -q` (`33 passed`)
- `pytest tests/qarrays -q` (`98 passed`, existing sparse dense-conversion warnings)

The new GitHub Actions run is awaiting maintainer approval for the fork workflow. @Nicolas-Lepage, when you have a chance, could you please confirm whether your older requested-changes review is still blocking anything after Gautier’s approval? Thanks.

Nicolas-Lepage (approved):
@jackson5689 Thanks for the great work! I'm sorry it couldn't be finished within the hackathon timeline. This turned out to be a much larger PR than initially estimated, and you handled it really well. Congrats!

jackson5689:
Thanks again for merging this. I noticed #1080 was closed after the merge, but I do not see `jackson5689` assigned on the issue / unitaryHACK leaderboard yet. If this PR is intended to count for the unitaryHACK bounty, could you please assign `jackson5689` to #1080 or let me know if there is another step I should take?

PR #1116, opened 2026-06-10
Add trajectory-level and statistical ensemble tests for stochastic solvers

### LINKED ISSUE #1079: unitaryHACK - Add tests for stochastic solvers

Oishi1029:
Closes #1079.

## What this adds

**1. Ensemble-convergence vs `mesolve` extended to all four stochastic APIs**
(acceptance criterion 1). New `tests/dssesolve/test_dssesolve.py`,
`tests/dsmesolve/test_dsmesolve.py`, `tests/jsmesolve/test_jsmesolve.py`, mirroring
the existing `tests/jssesolve/test_jssesolve.py` (same driven-qubit system). The SME
tests run with partial efficiency (`etas=0.7`/`0.6`) and dark counts (`thetas=0.2`),
since the trajectory average must recover the Lindblad evolution for any detector
imperfection. Both methods of the diffusive solvers (`EulerMaruyama`, `Rouchon1`)
are covered.

**2. No-back-action trajectory test** (`tests/stochastic/test_no_backaction.py`),
exactly the setup proposed in the issue: `H = ω/2(σx⊗σx + σy⊗σy)`, `L = −σz⊗σz`,
`ψ(0)=|01⟩`. The dynamics stay in the odd-parity subspace where `L` acts as the
identity, so the measurement back-action must cancel *exactly* (algebraically, for
any `dt`). Every trajectory is compared to the analytical state
`cos(ωt)|01⟩ − i sin(ωt)|10⟩` with per-trajectory, per-time infidelity `< 1e-4`.
All four solvers, both methods each, plus imperfect-detection SME cases
(`η=0.4`, `θ=0.5`). For the jump solvers the test also asserts that clicks *do*
occur — they simply must have no effect.

**3. Statistical ensemble test** (`tests/stochastic/test_ensemble_statistics.py`),
one common physical problem for all four solvers: free qubit de-excitation
(`H=0`, `L=√γσ₋`, `ψ(0)=|e⟩`).
- *Jump solvers*: the per-trajectory excited population is exactly
  `Bernoulli(e^{−γt})`. Checked: binarity of trajectories, ensemble mean `e^{−γt}`,
  ensemble variance `p(1−p)`, and consistency between `result.clicktimes` and the
  saved populations (`Pe(t)=1` iff no click before `t`).
- *Diffusive solvers*: the first- and second-order statistics of the measurement
  record are known in closed form for this problem. An Itô computation (full
  derivation in the test docstring, cross-checked against the homodyne-current
  autocorrelation from the quantum regression theorem) gives `E[I_n] = 0` and

  `Var[I_n] = 1/Δ + (8η/(γΔ²))(e^{−γt_n/2} − e^{−γt_{n+1}/2})²`

  for the time-averaged record over `[t_n, t_{n+1}]`. The second term is the
  measurement-back-action contribution (up to +78% over shot noise with the chosen
  parameters), so this check fails if the back-action is missing, has the wrong
  sign, or the wrong amplitude — second-order record statistics see exactly what
  ensemble means are blind to.

## Statistical design

All tolerances are 5 standard errors of the corresponding estimator (false-failure
probability ~1e-6 per assertion, derivations in comments/docstrings), the mean and
variance bounds are re-checked at `ntrajs/4` to verify the `1/√ntrajs` accuracy
scaling, and the diffusive test contains an explicit statistical-power guard: the
back-action term must exceed 7 standard errors of the variance estimator, so the
test provably resolves the effect it targets. All seeds fixed; the suite also passes
with two other seed sets (not seed-tuned).

## Falsification check

To verify the tests fail for incorrect physics (acceptance criterion), I locally
injected three bugs into `fixed_step_stochastic_integrator.py` and confirmed each is
caught:
- diffusive SSE back-action coefficient `0.5 → 0.45` → no-back-action test fails;
- diffusive SME state decoupled from the noise → record-variance test fails (both η);
- jump SSE click rate scaled by 1.3 → Bernoulli mean test fails.

## Tests

```
$ pytest tests/stochastic tests/dssesolve/test_dssesolve.py tests/dsmesolve/test_dsmesolve.py tests/jsmesolve/test_jsmesolve.py
19 passed in 83.55s

$ pytest tests/jssesolve tests/dssesolve tests/dsmesolve tests/jsmesolve tests/stochastic
123 passed, 8 skipped in 214.95s

$ ruff check tests/ && ruff format --check tests/
All checks passed!
65 files already formatted
```

(macOS arm64, Python 3.12.13, jax 0.10.1, CPU.)

Runtime of the new tests is ~85 s on a laptop CPU (16–2000 trajectories depending on
the test, `save_states=False` where states are not needed).

---
Disclosure: I used Claude (Anthropic) to help draft these tests, which I then
reviewed, manually verified, and ran locally. unitaryHACK 2026 submission.


Nicolas-Lepage:
Thanks for the PR! Unfortunately, we ended up merging another contribution that addressed the same issue, so I'll be closing this PR. Thanks again for taking the time to work on it.

PR #1119, opened 2026-06-12
solved small copy paste bug

### LINKED ISSUE #1118: Bug: _test_gradient always asserts ysave gradients, never Esave gradients
**File**: `tests/integrator_tester.py`, `IntegratorTester._test_gradient`

The final assertion in `_test_gradient` is a copy-paste error. it re-checks the `ysave` gradient pair instead of the `Esave` one:

```
python# === test gradients depending on final Esave
true_grads_Esave = system.grads_expect(system.tsave[-1])
grads_Esave = jax_jac(loss_Esave)(system.params_default)

assert_allclose(true_grads_ysave, grads_ysave)  # ← bug: should be (true_grads_Esave, grads_Esave)
```

Should be:
`python assert_allclose(true_grads_Esave, grads_Esave)`
`grads_Esave` is computed and logged but its values are never actually checked against the analytic truth. Since `_test_gradient` is the shared tester for all integrators, `Esave` gradient correctness is currently unverified across the entire test suite.

Nicolas-Lepage:
Closes #1118

Reviewing A PR for low rank method i took a look at intergator_tester.py and think there is a copy paste bug. 

the assert for gradeint accuracy of final state is present 2 times when the second time should be gradient accuracy of expectation with respect to observable.

### DIFF
--- a/tests/integrator_tester.py
+++ b/tests/integrator_tester.py
@@ -94,7 +94,7 @@ class IntegratorTester:
         logging.warning(f'true_grads_Esave = {true_grads_Esave}')
         logging.warning(f'grads_Esave      = {grads_Esave}')
 
-        assert_allclose(true_grads_ysave, grads_ysave)
+        assert_allclose(true_grads_Esave, grads_Esave)
 
     def _test_hessian(
         self,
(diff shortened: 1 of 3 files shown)


gautierronan (dismissed):
Good catch! FYI, for such small fixes, you don't need to open an issue beforehand.

gautierronan (approved):
Thanks for fixing!

PR #1124, opened 2026-06-30
Fix wrong factor in Rouchon1 dsmesolve

HHeda:
I noticed a wrong factor in Rouchon1 dsmesolve (a missing square root in the expression of non monitored Kraus operators). It now agrees with all the others (dssesolve and EulerMayurama).

### DIFF
--- a/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
+++ b/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
@@ -657,7 +657,7 @@ class DSMESolveRouchon1Integrator(
                 jnp.sqrt((1 - eta) * self.dt) * _Lm
                 for eta, _Lm in zip(self.etas, Lm, strict=True)
             ]
-            + [self.dt * _Lc for _Lc in Lc]
+            + [jnp.sqrt(self.dt) * _Lc for _Lc in Lc]
         )
 
         rho = sum([M @ rho @ M.dag() for M in Ms])
--- a/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
+++ b/dynamiqs/integrators/core/fixed_step_stochastic_integrator.py
@@ -336,6 +336,7 @@ class JSMESolveEulerJumpIntegrator(JSMESolveFixedStepIntegrator):
         Lm = stack(self.Lm(t))
 
         # === Ccal(rho)
+        L_rho_Ldag = L @ rho @ L.dag()  # (nL, n, n)
         Lm_rho_Lmdag = Lm @ rho @ Lm.dag()  # (nLm, n, n)
         thetas = self.thetas[:, None, None]  # (nLm, 1, 1)
         etas = self.etas[:, None, None]  # (nLm, 1, 1)
@@ -366,7 +367,7 @@ class JSMESolveEulerJumpIntegrator(JSMESolveFixedStepIntegrator):
         # compute Lcal(rho), see `MESolveDiffraxIntegrator` in
         # `integrators/core/diffrax_integrator.py`
         Hnh = -1j * H - 0.5 * sum(L.dag() @ L)
-        tmp = Hnh @ rho + 0.5 * sum(Lm_rho_Lmdag)
+        tmp = Hnh @ rho + 0.5 * sum(L_rho_Ldag)
         Lcal_rho = tmp + dag(tmp)
         rho_noclick = (
             rho
--- a/tests/systems/stochastic_system.py
+++ b/tests/systems/stochastic_system.py
@@ -29,6 +29,11 @@ class StochasticSystem(System):
     def Ls(self, params: PyTree) -> list[QArray | TimeQArray]:
         """Compute the jump operators."""
 
+    @property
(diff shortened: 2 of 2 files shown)


gautierronan (changes requested):
Good catch! That was a pretty big mistake...

It's quite strange that none of our newly merged stochastic tests caught this. Do you think you could add a regression test to catch further mistakes like this one?

HHeda:
The error wasn't caught because there is no system in the test that features an unmonitored loss channel (all the quantum efficiencies are set to 1, and there is always one channel anyway).

gautierronan:
Hmm indeed, so I guess the problem is in [this file](https://github.com/dynamiqs/dynamiqs/blob/main/tests/systems/stochastic_system.py).

This means we should attach the `eta` values directly to the children classes instead of the `StochasticSystem` parent class, and that we should at least introduce one system with an unmonitored jump operator. The simplest way is probably to change `DampedOscillator` and to split its single jump operator in two ([here](https://github.com/dynamiqs/dynamiqs/blob/523d7f331d3158f63b7a56d90ca8025d688c82b5/tests/systems/stochastic_system.py#L126)) according to:
```python
    def Ls(self, params: PyTree) -> list[QArray | TimeQArray]:  # noqa: ARG002
        return [
            jnp.sqrt(0.5 * self.kappa) * dq.destroy(self.n),
            jnp.sqrt(0.5 * self.kappa) * dq.destroy(self.n)
        ]

    def etas(self) -> Array:
        return jnp.array([1.0, 0.0])
```

Do you think you could make these changes? It would be best to do it inside this PR such that we check this test indeed passes with the fix you made, and failed before. Thanks!!

HHeda:
I'm having trouble finding a test that flags the error. I did what you suggested, but it looks like the error introduced is very small

gautierronan:
Actually your latest test does seem to catch an error in the CI:
```python
___________________ TestJSMESolve.test_convergence[method0] ____________________
[gw0] linux -- Python 3.11.15 /home/runner/work/dynamiqs/dynamiqs/.venv/bin/python3

self = <tests.jsmesolve.test_jsmesolve.TestJSMESolve object at 0x7f5e382fa290>
method = EulerJump(dt=0.001)

    @pytest.mark.parametrize('method', _METHODS)
    def test_convergence(self, method):
>       self._test_convergence(damped_oscillator, method)

tests/jsmesolve/test_jsmesolve.py:25: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tests.jsmesolve.test_jsmesolve.TestJSMESolve object at 0x7f5e382fa290>
system = <tests.systems.stochastic_system.DampedOscillator object at 0x7f5e549b0410>
method = EulerJump(dt=0.001), ntrajs = 2000, atol = 0.005, atol_state = 0.01
seed = 1

    def _test_convergence(
        self,
        system: StochasticSystem,
        method: Method,
        *,
        ntrajs: int = 2000,
        atol: float = 5e-3,
        atol_state: float = 1e-2,
        seed: int = 1,
    ):
        # averaging the trajectories recovers the analytical Lindblad evolution,
        # checked on both the expectation values and the full state (as in
        # IntegratorTester for the deterministic solvers)
        result = system.run(self.SOLVER, method, _keys(seed, ntrajs))
    
        true_expects = system.expects(system.tsave)
>       assert jnp.allclose(result.mean_expects(), true_expects, atol=atol)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError

tests/stochastic_tester.py:65: AssertionError
=============================== warnings summary ===============================
tests/qarrays/test_sparse_dia_qarray.py::TestSparseDIAQArray::test_scalaradd[simple]
tests/qarrays/test_sparse_dia_qarray.py::TestSparseDIAQArray::test_scalaradd[batch]
tests/qarrays/test_sparse_dia_qarray.py::TestSparseDIAQArray::test_scalaradd[batch_broadcast]
  /home/runner/work/dynamiqs/dynamiqs/dynamiqs/qarrays/materialized_qarray.py:336: UserWarning: A sparse data array has been converted to dense layout due to element-wise addition with a raw array.
    return replace(self, data=self.data + jnp.asarray(y))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/jsmesolve/test_jsmesolve.py::TestJSMESolve::test_convergence[method0] - AssertionError
```

So I guess it's working as intended, no?

HHeda:
Indeed, but it would not have caught the first one.
I don't know why the ci fails here.

gautierronan:
It's a typing error when running `ty check dynamiqs`, but it's just a new release of `ty` that broke the CI. I'll fix it in a separate PR and we can then merge yours

gautierronan:
Argh, I didn't mean to close the PR, and I can't seem to reopen it for some reason... I'll open a new one, sorry

PR #1125, opened 2026-07-01
Simplify gradient and hessian computation in tests

gautierronan:
Small refactor to simplify lengthy and intractable analytical formulas inside tests.


abocquet (approved):

Nicolas-Lepage (approved):
Thanks @gautierronan for the second pass on the hessian code !

PR #1126, opened 2026-07-01
Fix bug when substracting qarray with timeqarray

gautierronan:
Fixes:
```python
import dynamiqs as dq

a = dq.destroy(4)
f = lambda t: t
x = dq.modulated(f, a)

print(x + a) # ok
print(a + x) # ok
print(x - a) # ok
print(a - x) # TypeError: must be real number, not ModulatedTimeQArray
```

### DIFF
--- a/dynamiqs/qarrays/qarray.py
+++ b/dynamiqs/qarrays/qarray.py
@@ -512,6 +512,8 @@ class QArray(eqx.Module):
         return self.__add__(y)
 
     def __sub__(self, y: QArrayLike) -> QArray:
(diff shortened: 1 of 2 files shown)


abocquet (approved):

Nicolas-Lepage (approved):
good catch! i took a look at if there were any other such errors in QArray but couldn't confirm any. Just  `__mul__/__rmul__` raises instead of returning NotImplemented for non-scalars, but not sure if it's actually asymmetric like `__sub__` was.

PR #1128, opened 2026-07-01
Fix arithmetic methods fallback for QArrays

### LINKED ISSUE #1123: QArray.__mul__ raises "NotImplementedError" instead of returning "NotImplemented", breaking __rmul__ dispatch
`SparseDIAQArray.__mul__` (and likely `DenseQArray.__mul__`) raises `NotImplementedError` instead of returning `NotImplemented` when the operand is not a batched scalar. This breaks  Python's reflected operator protocol: `__rmul__` on the right-hand operand is never tried because a raised exception propagates immediately, whereas returning `NotImplemented` would trigger the fallback.

## Reproducer

```python
import dynamiqs as dq

class MyScalar:
    def __rmul__(self, other):
        return "__rmul__ called"

s = MyScalar()
print(dq.sigmax() * s)  # raises NotImplementedError, should print "__rmul__ called"
```

## Expected behaviour
`dq.sigmax() * s` falls back to `s.__rmul__(dq.sigmax())`.

## Actual behaviour
`NotImplementedError` is raised in `QArray.__mul__`.

## Fix
Return `NotImplemented` instead of raising in the `is_batched_scalar` guard:
```python
if not is_batched_scalar(y):
    return NotImplemented
```

## Version
dynamiqs 0.3.4

gautierronan:
- Fix `NotImplemented` protocol handling in QArray arithmetic (`__mul__`, `__truediv__`, `__sub__`, `__pow__`) — previously any unrecognized operand raised `NotImplementedError`, breaking fallback to the other operand's reflected method (e.g. `dq.sigmax() * custom_obj` couldn't use `custom_obj.__rmul__`). This fixes https://github.com/dynamiqs/dynamiqs/issues/1123. 
- Fix infinite recursion when a `str` is passed to `isqarraylike`, `to_jax`, `to_numpy`, `get_dims` (e.g. `dq.asqarray('abc')` used to raise `RecursionError`).
- Widen `__mul__`/`__rmul__`/`__truediv__` type hints from `ArrayLike` to `QArrayLike` for consistency with `__add__`/`__sub__`; make `is_batched_scalar` a `TypeGuard` to keep this type-safe.
- Deduplicate repeated operand-resolution logic in `MaterializedQArray` into a shared `_resolve_operand` helper.

### DIFF
--- a/dynamiqs/_utils.py
+++ b/dynamiqs/_utils.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, get_args
+from typing import Any, TypeGuard, get_args
 
 import equinox as eqx
 import jax.numpy as jnp
@@ -45,7 +45,7 @@ def concatenate_sort(*args: Array) -> Array:
     return jnp.sort(jnp.concatenate(args))
 
 
-def is_batched_scalar(y: Any) -> bool:
+def is_batched_scalar(y: Any) -> TypeGuard[ArrayLike]:
     # check if a qarray-like is a scalar or a set of scalars of shape (..., 1, 1)
     return isinstance(y, get_args(ScalarLike)) or (
         isinstance(y, get_args(ArrayLike))
--- a/dynamiqs/qarrays/composite_qarray.py
+++ b/dynamiqs/qarrays/composite_qarray.py
@@ -141,7 +141,7 @@ class CompositeTerm(eqx.Module):
 
     # === Arithmetic ===
 
-    def __mul__(self, y: ArrayLike) -> CompositeTerm:
+    def __mul__(self, y: QArrayLike) -> CompositeTerm:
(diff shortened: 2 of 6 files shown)


Nicolas-Lepage (dismissed):
Thanks for this! I'll admit I was a bit lost in the reflection logic at first. One thing I noticed: all the new tests check the `NotImplemented`/`raise` dispatch logic, but none check that the valid paths (e.g. dividing by an actual scalar, multiplying by a real `QArrayLike`) still compute the correct result after these changes. Do we have enough coverage elsewhere to be confident this didn't accidentally break or over-raise on legitimate inputs?

gautierronan:
Yes, the tests that check the validity of the output are stored in `tests/qarrays/test_dense_qarray.py` and `tests/qarrays/test_sparse_dia_qarray.py` respectively. There's a check for each arithmetic operation for the various input types it accepts.

PR #1131, opened 2026-07-09
Speedup CI by introducing test sharding

gautierronan:
This PR introduces test sharding using `pytest-split`, such that different runners run different parts of the tests in parallel. This speeds up the CI from ~25 minutes to ~5 minutes, and also makes it easier to see what failed. 

To split the tests, I introduce a very simple rule that sets tests marked as instant to 0.1 seconds, tests marked as short to 2 seconds, and tests marked as long to 15 seconds. Then, the split is made automatically such as to balance the expected runtime between runners.

Example of the new CI (failing here):
Screenshot 2026-07-09 at 18 25 50


abocquet (dismissed):

Nicolas-Lepage (dismissed):
Thanks @gautierronan for this helpful CI change. I took a look more out of curiosity than actual review since I don't know much on this, but one thing to flag: tests without an explicit priority default to the "instant" bucket, so if we're not careful, larger tests could end up miscategorized as instant and throw off the shard balance.

gautierronan:
> Thanks @gautierronan for this helpful CI change. I took a look more out of curiosity than actual review since I don't know much on this, but one thing to flag: tests without an explicit priority default to the "instant" bucket, so if we're not careful, larger tests could end up miscategorized as instant and throw off the shard balance.

Good point, I'll fix this in a separate PR.

PR #1134, opened 2026-07-16
Avoid densifying expectation operators in low-rank integrator

### LINKED ISSUE #1133: mesolve() with LowRank densifies the operators upon measuring exp_ops()
@p-pietro 

mesolve() with LowRank densifies the operators upon measuring exp_ops() which causes a memory bottleneck.

[Line:](https://github.com/dynamiqs/dynamiqs/blob/main/dynamiqs/integrators/core/low_rank_integrator.py#L189)
```
def expval_from_m(m: Array, op: Array) -> Array:
    return jnp.sum(jnp.conj(m) * (op @ m))
```
`m` and `op` are treated as an Array so the operator gets densified

p-pietro:
Fixes #1133 : when using `method = LowRank()` in `mesolve()`, we are not limited anymore by the size in memory of the expectation operators.

I think that the CI fails due to pre-existing problems on `main`.


dynamiqs/integrators/core/low_rank_integrator.py
```diff
@@ -186,8 +187,9 @@ def initialize_m0_from_dm(
     return normalize_m(cols)
 
 
-def expval_from_m(m: Array, op: Array) -> Array:
-    return jnp.sum(jnp.conj(m) * (op @ m))
+def expval_from_m(m: Array, op: QArray) -> Array:
+    op_m = cast(QArray, op @ m).to_jax()
```
gautierronan, line 191: Is it the `to_jax` call that fixes this? I don't understand why there's a difference with the previous version?

gautierronan (changes requested):
Thanks a lot for the fix @p-pietro! It looks good but I don't understand the fix haha!

Would you be ok adding a short regression test as well?

p-pietro:
Hello @gautierronan , the difference is that instead of calling `to_jax` on each of the expectation operators, we keep them as (possibly sparse) `QArrays`, and call `to_jax` only on the combined `op @ m` object. 

This avoid converting all the expectation operators from DIA sparse to dense jax arrays, which partly defeated the purpose of the low rank method. 

We should be able to get rid of all the `jax.Array` conversions in this method once `QArray`s become aware of what they contain (ket / density matrix / low rank d.m.).

gautierronan (approved):
Perfect, thanks for the fix

PR #1135, opened 2026-07-17
Low rank integrator: optionally make save_extra act on m(t)

### LINKED ISSUE #1132: mesolve with LowRank method evaluates a full dense density matrix in save_extra
@p-pietro @gautierronan 

Right now, when running mesolve with `method=LowRank()` and measuring a quantity via a custom funtion in save_extra, the custom function operates on a full dense density matrix (dim = n x n, where n is the total Hilbert space dimension). This means that the simulation is going to be limited by the memory required to store a full state, which is exactly what we wanted to avoid with introducing the low-rank method.

**Proposal:**
I propose that, instead, the custom functions operate on a low-rank density matrix, whose dimension is n x K, K<<n being the chosen rank.

**Arguments:** 
- Unless `save_states=True` (which again eats up a lot of memory, so we want to avoid it), I see no other way of measuring useful stuff like the Wigner function
- Measuring expectation values via exp_ops currently evaluates the full operators. This is a bug that should ideally be resolved (I will open another issue for this), but before it gets resolved we could measure expectation values via custom functions if neccessary
- Altogether, without this feature there is no much advantage in using LowRank with respect to the standard methods, and benchmarking results imply the same. Based on benchmarking, the gain in available Hilbert space sizes that we get if we avoid evaluating a full state could be huge

**Downsides of a proposal:**
- User has to be aware that the custom functions need to be defined for low-rank form and not the full state. But, if one wants to return to the full form, it's a one-line change in the custom function, we should just make sure to document it properly

**Existing code** ([dynamiqs/integrators/core/low_rank_integrator.py](https://github.com/dynamiqs/dynamiqs/blob/main/dynamiqs/integrators/core/low_rank_integrator.py) line `301-320`):
```
def save(self, y: PyTree) -> SolveSaved:
        m = y
        m = normalize_m(m)

        msave = None
        if self.options.save_states:
            msave = asqarray(m, dims=self.dims)

        extra = None
        if self.options.save_extra:
            rho = asqarray(rho_from_m(m), dims=self.dims)
            extra = self.options.save_extra(rho)

        if self.Es_jax is not None:
            Esave = jnp.stack([expval_from_m(m, E) for E in self.Es_jax])
        else:
            Esave = None

        return SolveSaved(ysave=msave, extra=extra, Esave=Esave)
```

**Change:**
instead of 
```
     if self.options.save_extra:
            rho = asqarray(rho_from_m(m), dims=self.dims)
            extra = self.options.save_extra(rho)
```
call a function on m
```
     if self.options.save_extra:
            extra = self.options.save_extra(m)
```

p-pietro:
Closes #1132 .

We add an additional `save_extra_low_rank` boolean flag to `dq.method.LowRank`, which if `True` makes the user defined `save_extra` function act on `m` instead of `rho`. 

Unfortunately this implementation requires a pretty ugly `cast` to make type checking pass: https://github.com/dynamiqs/dynamiqs/blob/ec95bb2c9a52495c4e3104f5450961df90f08d5e/dynamiqs/integrators/core/low_rank_integrator.py#L313 this is because `save_extra`'s signature assumes that it is acting on a `QArray`, not a raw `jax.Array`. https://github.com/dynamiqs/dynamiqs/blob/ef2dde23b2bf5009803a97afb5ae28c67af77ba1/dynamiqs/options.py#L23


dynamiqs/method.py
```diff
@@ -805,6 +808,7 @@ class LowRank(Method):
     key: PRNGKeyArray
     linear_solver: LinearSolver = eqx.field(static=True, default=LinearSolver.QR)
     perturbation_scale: float = eqx.field(static=True, default=1e-5)
+    save_extra_low_rank: bool = eqx.field(static=True, default=False)
```
gautierronan:
```suggestion
    is_save_extra_low_rank: bool = eqx.field(static=True, default=False)
```
I propose this name instead. The previous one suggested that this was a new `save_extra` function: here it's a bit more verbose but makes it clear that it's a boolean. If you can find a shorter name that would be great, but I couldn't find one myself.

gautierronan (changes requested):
Thanks for the feature. I don't think the `cast` is ugly: it's exactly how it should be precisely because we need to respect the typing of `save_extra`. I suggested a small renaming below, otherwise it LGTM.

p-pietro:
I find it a bit ugly in the sense that we are making `m` pretend to be something it is not: if a method of `QArray` is called on it it will just fail.

However this should not be an issue if the user (or Claude) reads the documentation.

gautierronan:
The issue here is rather that the LowRank method uses jax.Arrays instead of QArrays (unlike all other methods). Ideally, we should only make use of QArrays internally as this is where we have optimized operations. But for now it's not possible because of the lack of such a LowRankQArray.

p-pietro:
Yes exactly, but I think it is adequate for now.

gautierronan (approved):
LGTM. Thanks!

PR #1136, opened 2026-07-20
Fix wrong factor in Rouchon1 dsmesolve

gautierronan:
Reopens https://github.com/dynamiqs/dynamiqs/pull/1124 after mistake in closing. Written by @HHeda.


gautierronan:
Merging because previously approved in the original PR.

PR #1138, opened 2026-08-16
Fix sparse DIA initialization on TPU

### LINKED ISSUE #1137: Sparse DIA initialization fails on TPU due to non-scalar `eqx.error_if` predicate
Sparse DIA QArrays currently fail to initialize on TPU.

Minimal example:

```python
import dynamiqs as dq

dq.eye(25, layout=dq.dia)
```

On a TPU v5e with:

```text
dynamiqs 0.3.6
jax 0.11.0
jaxlib 0.11.0
equinox 0.13.8
```

this gives:

```text
JaxRuntimeError: INVALID_ARGUMENT: Unknown send channel 2 for host callbacks
```

The issue seems to come from `SparseDIADataArray.__check_init__`:

```python
zero_slice = slice(None, offset) if offset >= 0 else slice(offset, None)
check = self.diags[..., i, zero_slice] != 0
eqx.error_if(self.diags, check, error)
```

For the main diagonal (`offset == 0`), `check` has shape `(0,)`. This is enough to reproduce the TPU error without Dynamiqs:

```python
import jax
import jax.numpy as jnp
import equinox as eqx

@jax.jit
def f(diags):
    check = diags[..., 0, :0] != 0
    return eqx.error_if(diags, check, "error")

f(jnp.ones((1, 25))).block_until_ready()
```

A scalar predicate works correctly on the same TPU:

```python
@jax.jit
def f(x):
    return eqx.error_if(x, jnp.any(x < 0), "error")
```

Independently of the TPU error, Equinox documents `pred` as a scalar boolean, so passing `check` directly also appears to violate the `error_if` API contract.

A possible fix is:

```diff
- check = self.diags[..., i, zero_slice] != 0
+ check = jnp.any(self.diags[..., i, zero_slice] != 0)
  eqx.error_if(self.diags, check, error)
```

p-pietro:
Closes #1137.

The CI failure seems pre-existing.


gautierronan (approved):
LGTM! Thanks!

PR #1139, opened 2026-08-19
Give Rouchon methods a third-order dense output to fix forward-mode gradients

schlegeldavid:
## Motivation

`Rouchon1/2/3` declare `Forward` as a supported gradient, and it does work for parameters
entering `H` or the jump operators. But it silently returns a **wrong** Jacobian as soon as
the differentiated parameter enters the time grid, e.g. a time-of-flight or a delay shifting
`tsave`. On a driven damped qubit, `jax.jacfwd` of an expectation value with respect to a
scalar shift of `tsave` gives `-1.56e+02` with `Rouchon2()` where the true derivative is
`+3.38e-02`, and exactly `0.0` with `Rouchon1(dt=...)`. No exception is raised, so callers
get a plausible-looking Jacobian (the worst failure mode for gradient-based fitting).

The cause was the `step_ts` injection in `MESolveAdaptiveRouchonIntegrator`, flagged with a
`# todo: can we do better?`:

```python
# fix incorrect default linear interpolation by stepping exactly at all times
# in tsave, so interpolation is bypassed
return replace(stepsize_controller, step_ts=self.ts)
```

That hack existed because `AbstractRouchonTerm.interpolation_cls` was
`LocalLinearInterpolation`, which is only first order, so saving at `tsave` by interpolation
would have destroyed the order of the scheme. Forcing a step at every save time routes a
differentiable quantity (`ts`) into step-size control and makes every save-time
interpolation interval degenerate, so `∂/∂t_k` of the linear interpolant becomes
`(y1 - y0) / (t1 - t0) ≈ 0/0`.

## Changes

- `RouchonPropertiesMixin` gains a `lindbladian(t, rho)` method, the vector field
  `dρ/dt = G ρ + ρ G† + Σ_k L_k ρ L_k†`, built from the existing `G(t)` and `self.L(t)`.
- `AbstractRouchonTerm` carries it as a second field next to `rouchon_step`, and
  `RouchonDXSolver.step` emits `dense_info = dict(y0=y0, y1=y1, k0=k0, k1=k1)` with
  `k0 = dt * lindbladian(t0, y0)` and `k1 = dt * lindbladian(t1, y1)`
  (`ThirdOrderHermitePolynomialInterpolation` expects dt-scaled derivatives).
- `interpolation_cls` becomes `ThirdOrderHermitePolynomialInterpolation`, so saving at times
  that are not step times no longer degrades the order of the scheme.
- The `stepsize_controller` override is deleted; adaptive Rouchon now uses the inherited
  controller, with `step_ts` unset.
- Diffrax progress meters are wrapped in `_ForwardModeProgressMeter`, which applies
  `stop_gradient` to the progress. Diffrax reduces the progress across batch dimensions with
  `unvmap_max`, which has no JVP rule, so any progress meter raised
  `NotImplementedError: Differentiation rule for 'unvmap_max' not implemented` as soon as the
  time grid carried a tangent. This is method-independent (`Tsit5` failed identically); the
  progress is a display-only quantity, so dropping its tangent is safe.

With this, `∂ρ/∂t_k` evaluates to the Lindbladian at `t_k`, the physically correct
derivative, and fixed-step Rouchon gains a nonzero time-shift derivative for free. Cost is
two extra applications of the Lindbladian per step, small next to the Kraus map.

## Behavior change

Adaptive Rouchon no longer forces a step at every time in `tsave`. Previously this
implicitly capped the step size at the save-grid spacing, so a fine `tsave` with a loose
tolerance silently returned results far more accurate than requested. Now `atol`/`rtol` are
honored as they are for every other method. On a modulated-`H` damped oscillator (`n = 8`,
301 save points, max error against `Tsit5(atol=rtol=1e-12)`):

| `atol = rtol` | `Rouchon2()` before → after | `Rouchon3()` before → after |
|---|---|---|
| `1e-4` | 2.49e-04 → 2.89e-03 | 1.91e-05 → 3.45e-03 |
| `1e-6` (default) | 5.19e-05 → 4.82e-05 | 1.91e-05 → 3.16e-05 |
| `1e-8` | `max_steps` reached | 3.71e-04 → 2.99e-05 |

The dense output itself costs no accuracy: with 4 save points instead of 301 the errors after
the change are identical (2.891e-03 and 3.446e-03 at `atol = rtol = 1e-4`), i.e. the
remaining error is step truncation error, not interpolation error. Note also that the
previous behavior could exhaust `max_steps` at tight tolerances on a fine save grid, which no
longer happens. Users relying on the old implicit step cap should pass a tighter tolerance or
a `dtmax`.

## Tests

- New `tests/mesolve/test_rouchon_forward_ad.py`: `jacfwd` of an expectation value with
  respect to a shift of `tsave`, compared against a central difference computed with
  `Tsit5(rtol=atol=1e-8)`, for `Rouchon1/2/3` fixed step and `Rouchon2/3` adaptive. The
  Hamiltonian is time-dependent and `t0` is pinned explicitly, since a constant Hamiltonian
  is invariant under time translation and would pass even with a broken tangent path (there
  is a comment saying so, to keep the test from being simplified into a no-op).
- Same file: a `Forward()` gradient with a progress meter enabled, which used to raise
  `NotImplementedError: unvmap_max`.
- `tests/mesolve/test_adaptive_rouchon.py::test_gradient` is now parametrized with
  `Forward()` in addition to `Direct()` (the fixed-step suite already was).
- `test_pwc_hamiltonian` still passes: the `ClipStepSizeController` wrapping around PWC
  Hamiltonians is diffrax's own and independent of the deleted override. Its docstring is
  updated, since the `dataclasses.replace` `TypeError` it guarded no longer exists.

### DIFF
--- a/dynamiqs/integrators/core/diffrax_integrator.py
+++ b/dynamiqs/integrators/core/diffrax_integrator.py
@@ -26,7 +26,7 @@ from ...method import (
     _DEFixedStep,
 )
 from ...options import Options
-from ...progress_meter import AbstractProgressMeter
+from ...progress_meter import AbstractProgressMeter, _ForwardModeProgressMeter
 from ...result import MESolveResult, Result, SolveSaved
 from ...utils.vectorization import slindbladian, unvectorize, vectorize
 from .abstract_integrator import BaseIntegrator
@@ -144,6 +144,11 @@ class DiffraxIntegrator(BaseIntegrator, AbstractSaveMixin, AbstractTimeInterface
                     get_dtmax, stepsize_controller, dtmax, is_leaf=lambda x: x is None
                 )
 
+            # === prepare progress meter
+            progress_meter = _ForwardModeProgressMeter(
+                cast(AbstractProgressMeter, self.options.progress_meter).to_diffrax()
+            )
+
             # === solve differential equation with diffrax
             return dx.diffeqsolve(
                 self.terms,
@@ -157,9 +162,7 @@ class DiffraxIntegrator(BaseIntegrator, AbstractSaveMixin, AbstractTimeInterface
                 adjoint=self.adjoint,
                 event=event,
                 max_steps=self.max_steps,
-                progress_meter=cast(
-                    AbstractProgressMeter, self.options.progress_meter
-                ).to_diffrax(),
+                progress_meter=progress_meter,
             )
 
     def run(self) -> Result:
--- a/dynamiqs/integrators/core/rouchon_integrator.py
+++ b/dynamiqs/integrators/core/rouchon_integrator.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from abc import abstractmethod
 from collections.abc import Callable, Sequence
-from dataclasses import replace
 from typing import ClassVar
 
 import diffrax as dx
@@ -11,7 +10,7 @@ import jax
 import jax.numpy as jnp
 from diffrax import AbstractRungeKutta, Bosh3, Euler, Midpoint, ODETerm
 from diffrax._custom_types import VF, Args, BoolScalarLike, Control, RealScalarLike, Y
-from diffrax._local_interpolation import LocalLinearInterpolation
+from diffrax._local_interpolation import ThirdOrderHermitePolynomialInterpolation
 from jaxtyping import PyTree
 
 from ...gradient import Forward, Gradient
@@ -31,6 +30,8 @@ class AbstractRouchonTerm(dx.AbstractTerm):
 
     rouchon_step: Callable[[RealScalarLike, RealScalarLike, Y], tuple[Y, Y]]
     # should be defined as `rouchon_step(t0, t1, y0) -> y1, error`
+    lindbladian: Callable[[RealScalarLike, Y], Y]
+    # should be defined as `lindbladian(t, y) -> dy/dt`, only used for dense output
 
     def vf(self, t: RealScalarLike, y: Y, args: Args):
         del t, y, args
@@ -45,7 +46,7 @@ class AbstractRouchonTerm(dx.AbstractTerm):
 class RouchonDXSolver(dx.AbstractSolver):
     _order: int
     term_structure = AbstractRouchonTerm
-    interpolation_cls = LocalLinearInterpolation
+    interpolation_cls = ThirdOrderHermitePolynomialInterpolation
 
     def init(
         self,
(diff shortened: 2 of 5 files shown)


Nicolas-Lepage:
Hi David!

Thanks for the PR, I think this is a needed addition for Dynamiqs.

I spent a bit of time looking into it, not the code itself for now but the general approach.

My worry is that with Rouchon methods dense output is quite tricky, and that is actually why the previous code forced every tsave to be a real time step. When you pick Rouchon methods it most likely means you want your saved snapshots to respect the CPTP nature of the equation.

The interpolation scheme you propose does preserve the trace (as long as the renormalization at time steps is kept), but it can still violate the PSD nature of density matrices at the interior points.

So I think we need to decide with @abocquet and @gautierronan what the best approach is here.

It looks like it is possible to reuse the algebra already computed by the base Rouchon step and rescale the Kraus map to whatever tsave falls inside the step. This preserves CP exactly, and you can normalize cheaply for these dense outputs by dividing by the trace of rho instead of going through the full Cholesky step. That is fine because the state you start from (rho0 of the step) is already exactly trace one from the step level normalization, and the dense output is only ever used as an output, never fed back as the starting point of another step, so there is nothing for a small residual to accumulate into.

The drawback of this approach is that reconstructing the exact Kraus map for a given ratio between the dense time and the full step needs a linear solve for every dense output, and this can get slow if you have a lot of tsave points compared to time steps.

Last point, this does not seem to suffer from the same problem the original method had, and it still allows differentiating with respect to time offsets.

Nicolas-Lepage (changes requested):
Hi David!

Thanks for the PR, I think this is a needed addition for Dynamiqs.

I spent a bit of time looking into it, not the code itself for now but the general approach.

My worry is that with Rouchon methods dense output is quite tricky, and that is actually why the previous code forced every tsave to be a real time step. When you pick Rouchon methods it most likely means you want your saved snapshots to respect the CPTP nature of the equation.

The interpolation scheme you propose does preserve the trace (as long as the renormalization at time steps is kept), but it can still violate the PSD nature of density matrices at the interior points.

So I think we need to decide with @abocquet and @gautierronan what the best approach is here.

It looks like it is possible to reuse the algebra already computed by the base Rouchon step and rescale the Kraus map to whatever tsave falls inside the step. This preserves CP exactly, and you can normalize cheaply for these dense outputs by dividing by the trace of rho instead of going through the full Cholesky step. That is fine because the state you start from (rho0 of the step) is already exactly trace one from the step level normalization, and the dense output is only ever used as an output, never fed back as the starting point of another step, so there is nothing for a small residual to accumulate into.

The drawback of this approach is that reconstructing the exact Kraus map for a given ratio between the dense time and the full step needs a linear solve for every dense output, and this can get slow if you have a lot of tsave points compared to time steps.

Last point, this does not seem to suffer from the same problem the original method had, and it still allows differentiating with respect to time offsets.

gautierronan:
Agreed with @Nicolas-Lepage. The fact that there's a silent bug on the returned gradient is pretty bad and needs to be fixed. Nonetheless, not preserving the positivity of the output for a method that is explicitly built for this property is not a good solution.

I don't quite understand where the bug comes from actually. Is it a direct consequence of the fact that we can't save at arbitrary times?

If so, I agree that the obvious solution is to make smaller-sized Rouchon steps for each intermediate save time. However, it's not ideal in terms of speed/overhead. Do you have a better idea? Or could you prompt Claude to see if he can find one?

Nicolas-Lepage:
@gautierronan I asked Claude for better solutions but couldn't find a better one than this sub-Kraus step.

What seems to take time is the linear solves needed for the different Kraus map terms, so these could be cached, but the cache can start to get quite large. 

For the compute time : 
I've done some tests with a vibe-coded version of the idea. It seems that letting the solver choose its own steps, but having sub-Kraus steps for dense evaluation, sort of evens out when the number of dense evals is moderate. But when it's much larger than the number of time steps, it starts to bite pretty hard.

schlegeldavid:
I agree, that we have to be careful here about CPTP.
It seems that we have two options, when it comes to save times:

1. Use interpolation, violating CPTP. Just as any other dynamiqs method. Most flexible, no real overhead.
2. Since the method is CPTP by construction, enforce also CPTP at the save times by additionally evaluating the Kraus map at the requested points. Some overhead required

Could there be a modification of point 2, where instead of computing the density matrix at the requested times a posteriori, one directly tells the adaptive step solver to step at these points (maybe this could be directly done through `discontinuity_ts`)? This would avoid a lot of coding overhead.

PR #1142, opened 2026-08-27
Pass the current time to save_extra

rsnegrin:
### Summary

`save_extra` only received the state, so time-dependent callbacks were not possible. This changes the API from `f(y)` to `f(t, y)`.

`diffrax` already provides `t` to the save callback, but we were not using it:

`fn = lambda t, y, args: self.save(y)`

### Changes

- Give `t` as the first argument to `save()` across all integrator families.
- Change `Options.save_extra` to `Callable[[Scalar, QArray], PyTree]`.
- In `expm_integrator`, carry `times[1:]` alongside the propagators in `lax.scan`.
- In `montecarlo_integrator`, apply `save_extra` to each saved mean state via `vmap`, matching the other integrator families.

This is a **breaking API change**. Current callbacks such as `save_extra=lambda y: ...` must be changed to `save_extra=lambda t, y: ...`.

### Testing

Added `tests/test_save_extra.py` covering all solver families that support `save_extra`: `mesolve`, `sesolve`, propagators, `jssesolve`, `jsmesolve`, `ds/dsmesolve`, both Monte Carlo methods, and low-rank.

### Total changes
Core changes are around 100 lines of code. The rest is the added tests and updating [.test_durations](https://github.com/dynamiqs/dynamiqs/pull/1142/changes#diff-807f057362128031f70331c790526648b1a2eee6e447f55198e43680addc852f)


gautierronan:
Thanks for the PR @rsnegrin. I'm not against this change, but it's indeed breaking the API and, most importantly, making it a bit more complicated. Could you share your use case for this?

rsnegrin:
Hello @gautierronan !

My main use case is to compute different time-bin higher-order correlation functions,

$$
\mathbb{E}[I_t, I_{t+\tau}] = \text{Tr}\left[C_L e^{\tau \mathcal{L}} C_l \rho_t\right].
$$

For very large matrices, I can't save the full array of $\rho_t$ in memory and compute them a posteriori, so I have to do it with `save_extra`. 

I don't think I currently have a way of doing this, since I can't just use an observable with `exp_ops`, and to use `save_extra` I need access to the time $t$.

Also this will apply to any time dependent callback.

Maybe I'm missing a simpler way of doing this.

Let me know!

tests/test_save_extra.py
gautierronan, line 1: This file should live inside `tests/core/`

tests/test_save_extra.py
```diff
@@ -0,0 +1,214 @@
...
+        (Htd, Euler(dt=DT)),
+        (Htd, Rouchon1(dt=DT)),
+        (Htd, Rouchon2(dt=DT)),
+        (Htd, Rouchon3(dt=DT)),
+        (Hpwc, Expm()),
+    ],
+)
+def test_mesolve(H, method):
+    result = dq.mesolve(H, jump_ops, psi0, tsave, method=method, save_extra=save_extra)
+    assert_extra_correct(result)
```
gautierronan: All tests in this file are missing a time marker, e.g. `@pytest.mark.run(order=TEST_SHORT)`, which are needed to generate `.test_durations`

tests/test_save_extra.py
```diff
@@ -0,0 +1,214 @@
...
+# time-dependent Hamiltonian (a constant drive cannot distinguish a correct t from
+# a wrong one combined with a compensating state error)
+Htd = dq.modulated(lambda t: jnp.cos(2.0 * t), adag_a)
+# piecewise constant Hamiltonian for the Expm method, with discontinuity times
+# interleaving with tsave to catch a times[1:]/times[:-1] off-by-one
+Hpwc = dq.pwc(
+    jnp.asarray([0.0, 0.15, 0.55, 1.0]), jnp.asarray([1.0, -0.5, 2.0]), adag_a
+)
+
+jump_ops = [0.5 * a]
```
gautierronan: Could you reuse (and update if needed) the systems that are already inside `systems/` here? It's exactly what they are for.

tests/test_save_extra.py
gautierronan: Could you curate the list of tests in this file a bit more? For now, it looks like there are many tests for such a simple feature change.

Also, you may group tests together to improve readability, e.g.:
```python
CASES = [
    pytest.param(partial(dq.mesolve, Htd, jump_ops, psi0, tsave, method=Tsit5()), id='mesolve-tsit5'),
    pytest.param(partial(dq.mesolve, Hpwc, jump_ops, psi0, tsave, method=Expm()), id='mesolve-expm'),
    pytest.param(partial(dq.jssesolve, Htd, jump_ops, psi0, tsave, keys, method=EulerJump(dt=DT)), id='jsse-eulerjump'),
    ...
]

@pytest.mark.run(order=TEST_SHORT)
@pytest.mark.parametrize('run', CASES)
def test_save_extra_receives_current_time(run):
    result = run(save_extra=save_extra)
    expected = (1 + result.tsave) * dq.expect(adag_a, result.states)
    assert jnp.allclose(result.extra, expected, atol=1e-6)
```

tests/test_save_extra.py
```diff
@@ -0,0 +1,214 @@
...
+    method = _lowrank_method(is_save_extra_low_rank=True)
+    result = dq.mesolve(
+        Htd,
+        jump_ops,
+        psi0,
+        tsave,
+        method=method,
+        save_extra=lambda t, m: t,  # noqa: ARG005
+    )
+    assert jnp.allclose(result.extra, tsave, atol=1e-6)
```
gautierronan: There is already a test for `is_save_extra_low_rank` inside `mesolve/test_adapative_lowrank.py`. You can update the test there.

dynamiqs/integrators/core/event_integrator.py
```diff
@@ -197,7 +197,7 @@ def click(y: JumpState) -> JumpState:
         clicktimes = jnp.full((len(self.Ls), self.options.nmaxclick), jnp.nan)
         indices = jnp.zeros(len(self.Ls), dtype=int)
         y = stack([self.y0] * len(self.ts))
-        saved = self.reorder_Esave(self.save(y))
+        saved = jax.vmap(self.save)(jnp.asarray(self.ts), y)
```
gautierronan, line 200: Why is `reorder_Esave` not needed anymore here? That's surprising, no?

dynamiqs/integrators/core/expm_integrator.py
```diff
@@ -75,12 +75,14 @@ def run(self) -> Result:
         step_propagators = expm(delta_ts[:, None, None] * As)  # (ntimes-1, N, N)
 
         # === combine the propagators together
-        def step(carry: QArray, x: QArray) -> tuple[QArray, Saved]:
+        def step(carry: QArray, xs: tuple[QArray, Array]) -> tuple[QArray, Saved]:
             # note the ordering x @ carry: we accumulate propagators from the left
+            x, t = xs
```
gautierronan: nit: can you maintain the ordering (t, x) instead of (x, t) as you have throughout the rest of the PR?

gautierronan (changes requested):
Indeed I don't think there's currently a way to achieve your usecase. In principle, it's completely in the philosophy of this `save_extra` to have complete freedom over what you want to save, so it makes sense to add the time in. Let's do it!

Below are some review comments. Could you fix them before merging? The CI is also broken I believe

dynamiqs/integrators/core/event_integrator.py
rsnegrin: So my understanding is that the old code build `Esave` as `(nE, ntsave)`  because it was a batched call, so it used `reorder_Esave` to flip it to `(ntsave, nE) ` to match diffrax's layout.

Usuing `jax.vmap(self.save)` returns `(ntsave, nE)` directly, so no reorder is needed.

tests/core/test_save_extra.py
```diff
@@ -0,0 +1,125 @@
...
+class TestSaveExtra:
+    # Expm only supports constant or pwc Hamiltonians, hence the constant-H systems
+    @pytest.mark.parametrize(
+        ('system', 'method'),
+        [
+            (tdqubit, Tsit5()),  # diffrax adaptive step
+            (tdqubit, Euler(dt=DT)),  # diffrax fixed step
+            (dense_cavity, Expm()),  # expm scan
+        ],
+    )
```
gautierronan: We don't need to test so extensively such a simple feature, where essentially all integrators use the same pipeline. I propose to keep only a single (system, method) combination per set of solvers. So one for `test_sesolve`, one for `test_mesolve`, one for `test_stochastic`, one for `test_propagator` and you can keep `test_jit_and_gradient` as is.

gautierronan (changes requested):
Thanks for the changes. I have one last comment about tests, which are a bit too heavy for now. We already have a pretty long CI, we're trying to keep it contained.

tests/core/test_save_extra.py
rsnegrin: Yes makes sense. Sometimes Claude gets a little excessive with the tests. I'll fix it

gautierronan (approved):
Thank you for the changes @rsnegrin. Happy to merge this one. @Nicolas-Lepage Do you want to have a quick look?

Nicolas-Lepage (approved):
I took a quick look and everything seems good ! 
thanks @rsnegrin

PR #1143, opened 2026-08-28
Fix ruff and ty CI failures from ruff 0.16 and ty upgrades

Nicolas-Lepage:
This PR fixes 38 ruff errors and 3 ty errors that appeared after the CI environment picked up ruff 0.16 and a newer ty release.

The 2 RUF036 errors were fixable directly.

The 36 PLR0917 errors flagged functions and methods with more than five positional arguments, resolved by ignoring that rule in pyproject.toml since the project already tolerates many arguments through the existing PLR0913 ignore.

The 2 dataclass field order errors came from Rouchon2 and Rouchon3 declaring dt without a default while combining base classes that do have defaults, fixed by giving dt a proper default in both classes instead of ignoring the check.

The last error was isherm being annotated to return bool everywhere while actually returning a jax Array on purpose so it stays JIT compatible, fixed by correcting the return type to Array across the whole QArray hierarchy instead of ignoring the mismatch.


gautierronan (approved):
Thanks!

PR #849, opened 2024-12-16
Fix the Wigner transpose bug

gautierronan:
Closes DYN-234

PR #1144, opened 2026-08-28
Fix dq.wigner() docstring

Nicolas-Lepage:
The docstring claimed w has shape (..., nyvec, nxvec), but the actual returned shape is (..., nxvec, nyvec) (confirmed by tests and reading PR #849). This caused user confusion about how to index/plot the array.


gautierronan (approved):
Thanks for fixing :)

PR #1149, opened 2026-08-31
Add a Performance documentation page

gautierronan:


p-pietro:
I think it may be useful to document here a couple more environment flags for advanced usage:

- `EQX_ON_ERROR=off` removes the performance penalty due to error checking. Due to XLA compiler differences, this gives only a ~10% speed-up on CUDA backends but is much more significant on TPUs.
- `TF_FORCE_UNIFIED_MEMORY = 1`, `XLA_PYTHON_CLIENT_MEM_FRACTION = 4` (or any integer >1) allows XLA memory to spill out of the GPU into system memory (unified memory), thus allowing to simulate larger systems at the cost of performance. This is especially useful on systems such as Grace Hopper, but in principle works anywhere with nVidia GPUs.

Nicolas-Lepage (approved):

PR #1156, opened 2026-09-03
Fix batched DIA Kronecker product and `SparseDIADataArray.devices()'

Nicolas-Lepage:
While finishing `CompositeQArray` I stumbled on two pre-existing, unrelated
bugs in the sparse DIA layer. Splitting them out here rather than bundling
them with the larger change.

## Problem 1 — batched Kronecker product

`and_sparsedia_sparsedia` used a bare `jnp.kron(left_diags, right_diags)`,
which kron's the batch axes together instead of broadcasting them. Only
worked by coincidence when unbatched or when one side's batch size is `1`
(what the existing tests cover), crashes otherwise:

```python
sA = dq.stack([a, 10 * a])   # batch=2
sB = dq.stack([b, 100 * b])  # batch=2
sA & sB
# shape is (4, 6, 6),  when dense version gives (2, 6, 6)
```

## Problem 2 — `SparseDIADataArray.devices()`

Unconditionally `raise NotImplementedError`. Any `.devices()` call on a
sparse qarray fails.

## Fix

Both mirror how `DenseDataArray` already solves the same problem.

- **Kron**: replace the bare `jnp.kron` with `_bkron`, a
  `jnp.vectorize(signature='(a,b),(c,d)->(ac,bd)')`-wrapped kron over the last
  two axes only. `dense_dataarray.py` already has an identical `_bkron` for
  the dense case (predates this PR by ~a year).
- **`devices()`**: `return self.diags.devices()`. `offsets` is
  `eqx.field(static=True)`, so it's baked into the pytree structure, not
  traced, it holds no device-bound data and plays no part in `jit`/`vmap`.
  `diags` is the only field that can ever live on a device, so this is the
  same one-liner `DenseDataArray.devices()` uses (`self.data.devices()`),
  just for the one field that matters here.

## Tests added

- `test_kronecker_batch_batch` : kron of two operands with the same non-1
  batch size (the case `valid_operation_keys` never covers).
- `test_devices`,  `SparseDIADataArray.devices()` vs. dense vs. `jax.devices()`.

Both fail on unpatched code, pass with the fix


gautierronan:
Hmm, I didn't catch the bug you report. The following works fine on my end:
```python
import dynamiqs as dq

a = dq.destroy(6)
b = dq.destroy(6)

sA = dq.stack([a, 10 * a])   # batch=2
sB = dq.stack([b, 100 * b])  # batch=2
sA & sB # shape (4, 36, 36)
```

Nicolas-Lepage:
sorry my description was wrong but the error is still there. 

using dense as the base behavior the `&` in dense broadcasts for batches which i would say should be the default behaviour as broadcasting is how most array manipulations work in python. 

there i changed the description from :

```python
sA = dq.stack([a, 10 * a])   # batch=2
sB = dq.stack([b, 100 * b])  # batch=2
sA & sB
# TypeError: incompatible shapes for broadcasting: (4, 6, 6), (2, 6, 6)
```

to

```python
sA = dq.stack([a, 10 * a])   # batch=2
sB = dq.stack([b, 100 * b])  # batch=2
sA & sB
# shape is (4, 6, 6),  when dense version gives (2, 6, 6)
```

dynamiqs/qarrays/sparsedia_primitives.py
```diff
@@ -275,14 +275,21 @@ def and_sparsedia_sparsedia(
         np.ravel(left_offsets_np[:, None] * n + right_offsets_np)
     )
 
-    # compute new diagonals
-    out_diags = jnp.kron(left_diags, right_diags)
+    # compute new diagonals. The Kronecker product acts on the (ndiags, n) block
+    # only; batch axes are broadcast, not multiplied together.
```
gautierronan:
```suggestion
    # compute new diagonals with broadcasted batch axes
```

dynamiqs/qarrays/sparsedia_primitives.py
```diff
@@ -275,14 +275,21 @@ def and_sparsedia_sparsedia(
...
 
     # merge duplicate offsets and return
     out_offsets, out_diags = _compress_sparsedia(out_offsets, out_diags)
     return out_offsets, out_diags
 
 
+@partial(jnp.vectorize, signature='(a,b),(c,d)->(ac,bd)')
+def _bkron(a: Array, b: Array) -> Array:
+    # batched Kronecker product over the last two axes
+    return jnp.kron(a, b)
```
gautierronan, line 288: This exact function already exists inside `dense_dataarray.py`: we can move it one layer up inside `qarrays/utils.py` and import it in both files to avoid duplication

gautierronan (changes requested):
Thanks for the fixes. I added two minor comments below if you don't mind fixing!

dynamiqs/qarrays/sparsedia_primitives.py
Nicolas-Lepage: Agreed :) 
I tried moving `_bkron()` to`utils.py` first, but that creates a circular import: `sparsedia_primitives.py` is itself imported by `utils.py`, so `utils.py` can't be imported back into it. Same problem on the `dense_dataarray.py` side. (`from __future__ import annotations` doesn't help here because we have a runtime import / call)

So instead I kept the single definition in `sparsedia_primitives.py` and imported it into d`ense_dataarray.py`, since `dense_dataarray.py` already imports from `sparsedia_primitives.py` for dense-to-sparse conversion.

gautierronan: Hmm, yes these circular imports are a bit annoying. That's good. thanks!

gautierronan (approved):

PR #1157, opened 2026-09-03
filling composite Qarray contract Composite qarray construct inspect materialize

Nicolas-Lepage:
This is the first PR in breaking down the CompositeQArray class into pieces that can actually be reviewed. The full class landed as one giant diff, more then 1000 lines of new code, and about the same amount in tests. Going through all of that at once is not realistic, so the plan is to split it into 4 or 5 PRs, each one covering a coherent slice of behavior.

## Scope

This PR only covers three things for `CompositeTerm` and `CompositeQArray`:

* constructing them, with validation
* inspecting them: shape, dtype, layout, ndiags, devices, and repr
* materializing them: building the full matrix, and everything built on top of that, like `to_jax`, `to_numpy`, `to_qutip`, `asdense`, `assparsedia`

There is no arithmetic, no batch axis manipulation, and no spectral methods in this PR. Those are for the PRs that follow (things like swapaxes, moveaxis, expand_dims, the arithmetic dunders, and eig/eigh/expm/powm/trace/norm/ptrace).

## What construction guarantees

Both classes validate themselves on construction. Right now that guarantees:

* at least one operator (or one term)
* every operator is a materialized qarray, never a raw array or something else
* the coefficient is a scalar, or a properly batched scalar
* everything sits on the same device
* everything shares the same layout, for now

## Choices

* Printing a CompositeQArray never builds the full matrix. Everything shown in the repr (shape, dtype, layout, number of terms, per term coefficient and dims) can be computed from the pieces directly, without ever materializing.
* The single layout requirement. Right now, mixing a dense operator with a sparse diagonal one, in the same term or across terms, is rejected outright.
* The `_devices` helper. This one is only used inside the construction check, and it is separate from the public `devices()` method on purpose. `_devices` walks the whole object and simply skips any array that is a tracer, so the device check never blocks construction when it happens inside a jitted function. public device() should always give after correct innit only one device.
* there is an init guard that CompositeQarray must have multiple subsystems, this might be to restrictive

## Tests

15 tests in 5 functions.

* test_rejects_invalid_construction: 11 cases, one per invalid construction rule.
* test_qarray_matches_independent_oracle: shape, dtype, layout, mT and every conversion checked against a plain numpy computation.
* test_ndiags_is_lazy_and_guarded: diagonal count on a sparse composite, and a raise on a dense one.
* test_layout_conversion_stays_lazy: asdense and assparsedia stay lazy, explicit offsets fall back to materializing.
* test_repr_does_not_materialize: repr never touches the full matrix.


dynamiqs/qarrays/composite_qarray.py
```diff
@@ -113,17 +201,38 @@ def _eigvalsh(self) -> Array:
...
+    def assparsedia(
+        self, offsets: tuple[int, ...] | None = None
+    ) -> CompositeTerm | MaterializedQArray:
+        # `offsets` designates diagonals of the full matrix, which do not decompose
+        # into per-operator offsets
+        if offsets is not None:
+            return cast(
+                MaterializedQArray, self._materialize().assparsedia(offsets=offsets)
+            )
+        return replace(self, operators=tuple(op.assparsedia() for op in self.operators))
```
gautierronan, line 235: This one is a bit strange in terms of API. By giving `offsets` as input which is supposedly more information, we end up with a less powerful format (a `MaterializedQArray` instead of a `CompositeTerm`). I guess that we gain jit-compatibility though, but still that's a weird and not very flexible API.

In principle, `CompositeTerm.assparsedia` doesn't need to have the same API as `QArray.assparsedia`. I would thus propose that we accept offsets in a format that allows to keep a `CompositeTerm` as output and maintain jit-compatibility. In other words, the input `offsets` need to be specified for every operator of the term, e.g. as a `tuple[tuple[int, ...], ...]`.

This would also align the signature with that of `CompositeTerm.asdense`, i.e. something that always outputs a `CompositeTerm`.

gautierronan (changes requested):
Thanks for the PR @Nicolas-Lepage. I have one possibly important comment below, otherwise the PR looks good to me!

PR #1158, opened 2026-09-04
Feat/composite qarray batch axes

Nicolas-Lepage:
This is PR 2 out of 4 expected in the CompositeQArray breakdown, stacked directly on top of the construction, inspection and materialization PR. It adds the batch axis manipulation methods that PR1 explicitly left out.

## Scope

For both `CompositeTerm` and `CompositeQArray`, this PR covers:

* reshape, broadcast_to, swapaxes, moveaxis, expand_dims, restricted to the batch axes
* squeeze, which goes either lazy or materialized depending on the axis
* the batch axis case of `__getitem__`
* sum, which always falls back to materializing

Nothing here touches arithmetic (`__mul__`, `__matmul__`, `__and__`, `__add__`), `trace`, `conj`, `isherm`, `ptrace`, or the eig/eigh/expm/powm/norm group. Those still raise NotImplementedError and are for the next two PRs.

## Choices

* Aligning batch dims before touching an axis. Terms can start out with different batch shapes, so `_aligned()` broadcasts every leaf to the full shape first, otherwise something like swapaxes(0, 1) would not mean the same axis everywhere. Felt like the most natural fix, open to other opinions.
* squeeze with axis None materializes whenever a matrix dim is already 1, and stays lazy otherwise by computing the squeezable batch axes explicitly. Conservative on purpose, same MIXED convention as getitem.
* sum always materializes every term, even for a plain batch axis reduction. Simpler than special casing the separable cases, can revisit if it matters for performance.

## Tests

22 new tests across 15 functions in test_composite_qarray.py, plus 35 more cases in test_getitem.py from adding a composite variant to the existing parametrized dense/dia tests. 57 new tests in total.

* test_reshape_matches_oracle, test_broadcast_to_matches_oracle, test_squeeze_batch_axis_stays_lazy, test_getitem_batch_axis_stays_lazy: run on both dense and dia, checked against a numpy oracle built straight from the raw arrays.
* test_reshape_rejects_matrix_axis_change, test_broadcast_to_rejects_matrix_axis_change: cover the ValueError raised when the last two dims would change.
* test_swapaxes_matches_oracle, test_moveaxis_matches_oracle, test_expand_dims_matches_oracle: batch axis reshuffling against numpy, moveaxis and expand_dims each covering one axis and several axes at once.
* test_squeeze_matrix_axis_materializes, test_getitem_matrix_axis_materializes: confirm that touching the matrix dims falls back to materializing.
* test_squeeze_none_axis_materializes_only_when_matrix_dims_are_one: the axis None guard, one case with matrix dims of 1 that must materialize, one without that stays lazy.
* test_sum_batch_axis_matches_oracle, test_sum_all_axes_returns_scalar_array: sum with a partial axis and with axis None, both materialize.
* test_reshape_unchecked_materializes: `_reshape_unchecked` always goes through `_materialize()`.

All of them build on a shared `_batched_composite` fixture with two terms that deliberately start out with different batch shapes, so every method actually exercises `_aligned` instead of getting lucky with shapes that already match.


dynamiqs/qarrays/composite_qarray.py
```diff
@@ -41,6 +41,18 @@ def _coeff_batch_shape(coeff: ArrayLike) -> tuple[int, ...]:
     return coeff.shape[:-2] if coeff.ndim >= 2 else ()
 
 
+def _key_in_batch_dims(key: IndexType, ndim: int) -> bool:
+    # whether `key` only indexes batch axes, and thus leaves the tensor-product
+    # structure intact
+    return not key_touches_last_two_dims(key, ndim)
```
gautierronan, line 47: This helper function seems a bit trivial: I would propose removing it.

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -124,6 +136,11 @@ def _offsets(self) -> set[int]:
             offsets = {o * op.shape[-1] + p for o in offsets for p in op_offsets}
         return offsets
 
+    def _aligned(self) -> CompositeTerm:
```
gautierronan, line 139: Can we find a more explicit name for this function? I might be a bit slow, but it took me a bit of time to understand what it actually did

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -153,21 +170,75 @@ def conj(self) -> CompositeTerm:
         # conj(c·⊗A_k) = conj(c)·⊗conj(A_k) → each op's .conj() + jnp.conj(coeff).
         raise NotImplementedError
 
+    def reshape(self, *shape: int) -> CompositeTerm:
+        """Reshapes the term's batch axes."""
+        term = self._aligned()
+        operators = tuple(
+            op.reshape(*shape[:-2], *op.shape[-2:]) for op in term.operators
+        )
+        coeff = jnp.asarray(term.coeff).reshape(*shape[:-2], 1, 1)
```
gautierronan, line 179: I guess the `coeff` is of shape `(..., 1, 1)` for broadcasting, but I guess it would be better to keep it of shape `(...,)` and to add the `(1, 1)` shape when needed. This would make it cleaner if users want to access this coefficient.

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -153,21 +170,75 @@ def conj(self) -> CompositeTerm:
         # conj(c·⊗A_k) = conj(c)·⊗conj(A_k) → each op's .conj() + jnp.conj(coeff).
         raise NotImplementedError
 
+    def reshape(self, *shape: int) -> CompositeTerm:
```
gautierronan, line 173: This is an interesting one. So let's take the following example where `operators[0].shape == (3, 4, 2, 2)` and `operators[1].shape == (4, 5, 5)` such that `term.shape == (3, 4, 10, 10)`. Then `term.reshape(6, 2, 10, 10)` is indeed a valid call and the way you implement it is the only solution I believe.

Can you add the small example above to an inline comment to help readers understand why it's implemented as such?

dynamiqs/qarrays/composite_qarray.py
```diff
@@ -153,21 +170,75 @@ def conj(self) -> CompositeTerm:
...
 
-    def sum(self, axis: int | tuple[int, ...] | None = None) -> CompositeTerm:
-        # MATERIALIZE → _materialize().sum(axis).
-        raise NotImplementedError
-
-    def squeeze(self, axis: int | tuple[int, ...] | None = None) -> CompositeTerm:
-        # batch axes only → each op's .squeeze(axis) + jnp.squeeze(coeff, axis).
-        raise NotImplementedError
+    def sum(self, axis: int | tuple[int, ...] | None = None) -> QArray | Array:
+        return self._materialize().sum(axis=axis)
```
gautierronan, line 224: Do we require materializing if the `sum` is called over a batching axis? I would guess not. In this case, this should return a `CompositeTerm`, and this function should be written as a dichotomy

gautierronan (changes requested):
This PR was quite challenging to review just because the logic is quite non-trivial. I think AIs are now much better than we are at writing such logic. Nonetheless, below are a few comments.

For me to do a proper review, I would kindly request that you add inline comments with simple examples for all methods with complicated logic (e.g. like the one I made for `reshape`). If you have the time to add such examples, I'm happy to review again. Otherwise, I'll just trust the tests and the choices you made (which is completely fine!).

PR #1161, opened 2026-09-04
[DRAFT] CompositeQarray : algebra unary methods

Nicolas-Lepage:
This is PR 3 out of 4 expected in the CompositeQArray breakdown, stacked directly on top of the batch axis manipulation PR. It fills in the unary algebra, meaning everything that acts on a single composite qarray and returns a value or another qarray.

## Scope

For both `CompositeTerm` and `CompositeQArray`, this PR covers:

* conj and trace, both fully lazy
* `__mul__` by a scalar or a batched scalar, which only touches the coefficients
* powm and elpow, lazy for a single term, materialized for a sum
* addscalar, which stays lazy by appending one extra separable term
* ptrace, lazy, including the case where an operator spans several subsystems
* isherm, with a sufficient condition that avoids materializing in the common case
* the spectral group: `_eig`, `_eigvals`, `_eigvalsh`, `_eigh`, norm and expm

Nothing here touches the binary arithmetic (`__add__`, `__matmul__`, `__rmatmul__`, `__and__`, elmul). Those still raise NotImplementedError and are the subject of PR 4, which is where the actual point of the feature lives.

## Choices

* isherm is two stage. A real coefficient and Hermitian factors prove Hermiticity, so that check answers on its own and never builds the matrix. It is not necessary though, since a term like iY ⊗ iY is Hermitian with factors that are not, and a sum can be Hermitian purely by cross term cancellation, so a negative always gets confirmed on the full matrix. Short circuiting needs a concrete boolean, so under tracing the matrix is always built.
* `_eig` diagonalizes each factor with `_eig` rather than `_eigh`, exactly because of the iY ⊗ iY case above. `_eigh` on the other hand always materializes, even for a single term, because the tensor product of the per factor eigenvectors is not orthonormal when a factor has a degenerate eigenvalue, so it cannot satisfy the eigh contract. So `_eig` and `_eigh` deliberately disagree on laziness here.
* addscalar adds y times the all ones matrix as one extra term, which works because the all ones matrix is itself separable, J_a ⊗ J_b = J_ab. The factors are fully dense, so a dia composite falls back to dense with a warning, same behaviour as addscalar on main. A y that broadcasts against the matrix axes is not a batched scalar and cannot be represented by any separable term, so that case materializes.
* powm, elpow, `_eig`, `_eigvals` and `_eigvalsh` keep a single term fast path and materialize for a sum, since none of them distribute over a sum. norm is built from trace and `_eigvalsh` so that it inherits those fast paths, and matches `dq.norm`, tr for a PSD operator and the sum of absolute eigenvalues otherwise.
* Three term level methods are left raising NotImplementedError on purpose, with a comment saying why rather than a formula: expm, norm and `_eigh`. The corresponding `CompositeQArray` methods are either always materialized or built from other primitives, so nothing would ever call them.
* ptrace validates keep itself instead of delegating, and returns a materialized qarray when fewer than two subsystems are kept, since `dims` of a `CompositeQArray` must describe at least two subsystems.

## Tests

29 new cases across 20 functions in test_composite_qarray.py, all checked against a numpy oracle built straight from the raw arrays.

* Five new fixtures, chosen so the methods are actually exercised rather than trivially satisfied: a two term composite with complex entries, a batched coefficient and no symmetry, a single term with non normal factors that both have distinct eigenvalues, iY ⊗ iY which is Hermitian with factors that are not and is not PSD, that same term plus 2 I ⊗ I which is Hermitian and PSD, and a three subsystem composite where one operator spans two subsystems.
* test_conj_stays_lazy, test_trace_stays_lazy, test_mul_by_batched_scalar_stays_lazy, test_mul_by_non_scalar_raises: the fully lazy methods, plus the error message pointing at elmul.
* test_powm_matches_oracle, test_elpow_matches_oracle: one case for the single term fast path, one for the materialized sum.
* test_expm_materializes: against `jax.scipy.linalg.expm`.
* test_eig_single_term_stays_lazy, test_eig_multi_term_materializes: eigenvalues compared as a sorted spectrum, and the eigenvector columns checked to pair with the eigenvalues in the same order, which is what actually pins down the Kronecker ordering in `_kron_evals`.
* test_eigvalsh_single_term_handles_non_hermitian_factors: the iY ⊗ iY spectrum, which comes out wrong if the per factor route uses `_eigh`.
* test_eigh_matches_oracle: orthonormality and reconstruction, on a single term and on a sum, both fixtures having a degenerate eigenvalue.
* test_norm_matches_oracle: psd False on the operator that is not PSD, where the sum of absolute eigenvalues is 4 while the trace is 0, and psd True on the PSD one.
* test_isherm_matches_oracle, test_isherm_short_circuits_on_sufficient_condition, test_isherm_materializes_under_tracing: the three outcomes (sufficient, Hermitian but not sufficient, not Hermitian), the fact that the sufficient case never materializes, and the tracer guard under jit.
* test_ptrace_matches_oracle, test_ptrace_rejects_invalid_keep: whole operators kept, an operator partially kept, a single subsystem kept which must materialize, and the three validation errors.
* test_addscalar_batched_scalar_stays_lazy, test_addscalar_dia_warns_and_converts_to_dense, test_addscalar_non_scalar_materializes: the three branches, the lazy one also checking that exactly one term was appended.

Laziness is asserted structurally, by monkeypatching both `_materialize` methods to raise, for trace, `__mul__`, `_eig`, `_eigvalsh` and the isherm fast path. The suite was also run against a few deliberate mutations of the implementation (coefficient conjugation dropped in conj, coefficient dropped in trace, Kronecker ordering flipped in `_kron_evals`) to confirm the oracles actually bite.

## Known limitation

ptrace on a dia composite fails when a kept operator spans several subsystems. `MaterializedQArray.ptrace` returns a dense qarray, so the reduced term ends up with mixed layouts and is rejected by `__check_init__`. Not fixed here since it needs a decision on whether the reduced operator should be converted back to dia or the whole term densified, happy to take either in review or in a follow up.


gautierronan:
Let me know when you want me to review this one! :)

